PR #43, opened 2026-03-02
fix(agents): agents.list returns all DB agents in managed mode

tuntran:
## Problem

In managed mode, the WebSocket `agents.list` method only returns agents currently in the router's in-memory cache. The router uses lazy loading — agents are only cached when `Get(agentKey)` is first called (e.g. on first chat). Newly created agents are never visible in `agents.list` until they are actually used.

**Repro:** Create a new agent via HTTP API or WebSocket `agents.create`, then call `agents.list` — the new agent does not appear.

The HTTP `GET /v1/agents` endpoint works correctly because it queries the DB directly via `agentStore.ListAccessible()`.

## Root Cause

`handleList` called `m.agents.ListInfo()` which iterates the in-memory map only.

## Fix

- In managed mode, `handleList` now queries `agentStore.ListAccessible(ctx, client.UserID())` from DB.
- `isRunning` is enriched from the router cache via a new `GetCached()` helper (no lazy-load side effect).
- Standalone mode is unchanged (all agents pre-registered at startup).

## Files Changed

- `internal/agent/router.go` — add `GetCached()` method
- `internal/gateway/methods/agents.go` — fix `handleList` for managed mode

### DIFF
--- a/internal/agent/router.go
+++ b/internal/agent/router.go
@@ -90,6 +90,18 @@ func (r *Router) Get(agentID string) (Agent, error) {
 	return nil, fmt.Errorf("agent not found: %s", agentID)
 }
 
+// GetCached returns an agent only if it's already in the cache (no lazy-load).
+// Used by handleList to check isRunning without triggering DB resolution.
+func (r *Router) GetCached(agentID string) (Agent, error) {
+	r.mu.RLock()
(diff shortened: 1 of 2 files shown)


viettranx:
PR #48

PR #55, opened 2026-03-04
feat: deliver intermediate assistant text during tool iterations (block.reply)

### LINKED ISSUE #54: Agent loop drops intermediate assistant content during tool iterations
## Bug Description

When the agent loop runs multiple LLM iterations (text → tool call → text), only the **final iteration's content** is sent back to channels. Intermediate assistant messages (from iterations that include tool calls) are silently dropped.

### Example

User sends: "hello, nhớ mình ko bạn"

Agent runs 2 LLM iterations:
1. **Iteration 1**: LLM responds "Chào bạn! Mình vừa mới thức dậy..." + calls `memory_read` tool
2. **Iteration 2**: LLM responds "À, hóa ra workspace mới tinh luôn..."

**Expected**: User receives BOTH messages (as 2 separate messages)  
**Actual**: User only receives iteration 2. Iteration 1's content is lost.

### Root Cause

In `internal/agent/loop.go`, `finalContent` is only set when the loop breaks (no tool calls):

```go
// Line 777 — only captures the LAST iteration
finalContent = resp.Content
break
```

Intermediate iterations (with tool calls) save `resp.Content` to session history but never deliver it to the channel.

### OpenClaw Behavior (Reference)

OpenClaw handles this with the **block reply** mechanism:
- `assistantTexts[]` array accumulates text from EVERY assistant message across iterations
- `onBlockReply` callback delivers each assistant text to the channel immediately (as separate messages)
- `buildEmbeddedRunPayloads()` uses the full `assistantTexts[]` array for final payloads
- This works for ALL channels (Telegram, Discord, WhatsApp, etc.) — not just streaming ones

### Proposed Fix

Add a `block.reply` agent event type. When the agent loop receives assistant text alongside tool calls, emit an event that the channel system can deliver as an intermediate message.

**Files affected**: `loop.go`, `events.go`, `manager.go`, `gateway_consumer.go`  
**Does NOT change**: `RunResult` struct, channel `Send()` implementations, WebSocket protocol

### Affected Channels

All non-streaming channels lose intermediate content when agent uses tools mid-conversation.

Luvu182:
## Summary

- When the agent loop runs multiple LLM iterations (text → tool call → text), only the final iteration's content is sent to channels. Intermediate assistant messages are lost for all non-streaming channels (Zalo, Discord, WhatsApp, etc.)
- Adds `block.reply` event — each intermediate assistant text is delivered as a separate message, matching the TypeScript `onBlockReply` pattern
- Streaming channels (Telegram DM with `dm_stream`) are skipped to avoid double-delivery
- Dedup logic suppresses the final message when it matches the last block reply (race-free via `RunResult` fields)

## Changes

| File | Change |
|------|--------|
| `pkg/protocol/events.go` | +1 line: `AgentEventBlockReply` constant |
| `internal/agent/loop.go` | Emit `block.reply` event; track state in `RunResult` for dedup |
| `internal/channels/manager.go` | `RegisterRun` extended with metadata/streaming; `block.reply` handler |
| `cmd/gateway_consumer.go` | Move `outMeta` before `RegisterRun`; dedup final message |

## How it works

1. Agent loop emits `block.reply` after appending assistant message with tool calls (sanitized, non-empty, non-silent)
2. `HandleAgentEvent` receives it → for non-streaming channels, publishes outbound with filtered metadata (strips `reply_to_message_id` and `placeholder_key`)
3. On run completion, consumer checks `RunResult.BlockReplies` / `LastBlockReply` — if final content matches last block reply and no media, suppresses duplicate

## Test plan

- [ ] Send message that triggers tool calls on a non-streaming channel (e.g. Zalo: "hello, nhớ mình ko bạn" → triggers `memory_read`)
- [ ] Verify intermediate text delivered as separate message BEFORE tool execution
- [ ] Verify streaming channels (Telegram DM with `dm_stream`) don't get duplicate messages
- [ ] Verify dedup: when final content = last block reply, only 1 message sent

Closes #54

### DIFF
--- a/cmd/gateway_consumer.go
+++ b/cmd/gateway_consumer.go
@@ -179,8 +179,9 @@ func consumeInboundMessages(ctx context.Context, msgBus *bus.MessageBus, agents 
 		if lk := msg.Metadata["local_key"]; lk != "" {
 			chatIDForRun = lk
 		}
+		blockReply := channelMgr != nil && channelMgr.ResolveBlockReply(msg.Channel, cfg.Gateway.BlockReply)
 		if channelMgr != nil {
-			channelMgr.RegisterRun(runID, msg.Channel, chatIDForRun, messageID, outMeta, enableStream)
+			channelMgr.RegisterRun(runID, msg.Channel, chatIDForRun, messageID, outMeta, enableStream, blockReply)
 		}
 
 		// Group-aware system prompt: help the LLM adapt tone and behavior for group chats.
@@ -239,7 +240,7 @@ func consumeInboundMessages(ctx context.Context, msgBus *bus.MessageBus, agents 
 		})
 
 		// Handle result asynchronously to not block the flush callback.
-		go func(channel, chatID, session, rID string, meta map[string]string) {
+		go func(channel, chatID, session, rID string, meta map[string]string, blockReplyEnabled bool) {
 			outcome := <-outCh
 
 			// Clean up run tracking (in case HandleAgentEvent didn't fire for terminal events)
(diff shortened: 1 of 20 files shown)


viettranx:
Squash merged into main via faa47ab

PR #64, opened 2026-03-05
fix(summoner): sequential file generation with progress + HTTP timeout fix

Luvu182:
## Problem

Agent summoning fails with `context deadline exceeded` when using providers with deep reasoning (e.g. Kimi). The root cause is a dual-timeout conflict:

- `http.Client{Timeout: 120s}` kills each HTTP request after 120 seconds
- Some providers (Kimi with `kimi-for-coding`) need **160s+** for non-streaming requests due to extended reasoning (13K chars reasoning for a 2.5K chars output)
- The retry logic treats timeout as retryable, so it retries 3 times: `120s + 120s + 60s = 300s` → parent context expires → `context deadline exceeded`

Additionally, the single-prompt approach (generating both SOUL.md and IDENTITY.md in one call) gives no progress feedback — users see a spinner for minutes with no indication of activity.

### Backend (`internal/http/summoner.go`)
- **Sequential generation**: Split summoning into two LLM calls — SOUL.md first, then IDENTITY.md using SOUL.md as context for consistency
- **Per-file progress events**: Emit `file_generated` event after each file completes, so the UI can show incremental checkmarks (UI already handles this)
- **Resume on retry**: On retry, detect previously generated files (by comparing against default template) and skip them — only generate what's missing
- **Better logging**: Add `summoning: calling LLM` and `summoning: LLM responded` log lines for observability

### Provider timeout (`internal/providers/openai.go`, `anthropic.go`)
- Increase `http.Client` timeout from **120s → 300s** to accommodate providers with deep reasoning on non-streaming requests

### UI (`ui/web/src/pages/teams/`)
- Add info tooltip on team member labels explaining that only predefined agents can be added as team members (open agents lack shared context for collaboration)

## Test plan
- [ ] Create a new agent with a slow provider — verify SOUL.md checkmark appears before IDENTITY.md starts
- [ ] Kill the process after SOUL.md completes, restart, retry — verify SOUL.md is skipped and only IDENTITY.md is generated
- [ ] Verify regenerate (edit prompt) still works unchanged
- [ ] Check team create dialog and team members tab show the info tooltip

### DIFF
--- a/internal/providers/openai.go
+++ b/internal/providers/openai.go
@@ -103,6 +103,7 @@ func (p *OpenAIProvider) ChatStream(ctx context.Context, req ChatRequest, onChun
 	accumulators := make(map[int]*toolCallAccumulator)
 
 	scanner := bufio.NewScanner(respBody)
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024) // 1MB max line for large tool call / thinking chunks
 	for scanner.Scan() {
 		line := scanner.Text()
 		if !strings.HasPrefix(line, "data:") {


viettranx:
Superseded by #72, #73, #74 — split into focused PRs, kept USER_PREDEFINED.md, excluded browser headless switching.

PR #124, opened 2026-03-10
fix: route subagent announce through correct agent in delegation context

duhd-vnpay:
## Summary
- **Bug**: When a delegated agent spawns subagents, the announce results fail with `agent default not found` because the scheduler's `RunFunc` only parses `agent:{agentId}:{rest}` session keys
- **Root cause**: Delegate session keys use format `delegate:{uuid8}:{targetAgentKey}:{delegationId}` — the `delegate:` prefix was not handled, causing fallback to `"default"` (which doesn't exist in managed-mode DBs)
- **Fix**: Add `delegate:` case to the session key parser in `makeSchedulerRunFunc`, extracting the agent key from position 2

## Details

`gateway_consumer_process.go:makeSchedulerRunFunc` resolves which agent loop to use from the session key. It only handled:
```
agent:{agentId}:{rest}
```

But subagent announces within delegations use session keys like:
```
delegate:019cd724:ui-ux-design-agent:1a93a73f-795
```

This caused repeated errors:
```
ERROR "subagent announce: agent run failed" error="agent default not found: agent not found: default"
WARN  "unknown channel for outbound message" channel=delegate
```

The fix adds a `switch/case` on `parts[0]` to handle both `agent:` and `delegate:` prefixes.

## Test plan
- [x] Build succeeds (`docker compose build goclaw`)
- [x] Container starts healthy (`docker compose up -d goclaw`)
- [ ] Trigger a delegation that spawns subagents → verify announce routes to correct agent (no more `agent default not found` errors)
- [ ] Verify normal `agent:` session key routing still works (direct chat via Telegram/Zalo)

### DIFF
--- a/cmd/gateway_consumer_process.go
+++ b/cmd/gateway_consumer_process.go
@@ -14,10 +14,20 @@ import (
 // It extracts the agentID from the session key and routes to the correct agent loop.
 func makeSchedulerRunFunc(agents *agent.Router, cfg *config.Config) scheduler.RunFunc {
 	return func(ctx context.Context, req agent.RunRequest) (*agent.RunResult, error) {
-		// Extract agentID from session key (format: agent:{agentId}:{rest})
+		// Extract agentID from session key.
+		// Supported formats:
+		//   agent:{agentId}:{rest}
+		//   delegate:{sourceUUID8}:{targetAgentKey}:{delegationId}
 		agentID := cfg.ResolveDefaultAgentID()
-		if parts := strings.SplitN(req.SessionKey, ":", 3); len(parts) >= 2 && parts[0] == "agent" {
-			agentID = parts[1]
+		if parts := strings.SplitN(req.SessionKey, ":", 4); len(parts) >= 2 {
+			switch parts[0] {
+			case "agent":
+				agentID = parts[1]
+			case "delegate":
+				if len(parts) >= 3 {
+					agentID = parts[2]
+				}
+			}
 		}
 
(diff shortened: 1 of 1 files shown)


viettranx:
## PR #124 Decomposition Summary

This PR bundles multiple unrelated changes (38 files, +2765 LOC) and has conflicts with `main`. We've extracted and re-implemented each fix group as focused PRs on `main`:

### Merged
- **Cron context injection** (`c5b8860` on `main`) — `ExtraSystemPrompt` now tells the agent about the cron job's requester, channel, and delivery target. Also fixed `cron.create` from Web UI to capture `userID`.

### Open PRs
- **PR #127** `fix/delegate-session-key-routing` — Handles `delegate:` session key prefix in consumer routing (was causing "agent default not found" errors for delegated runs)
- **PR #128** `fix/orphan-history-cleanup` — `sanitizeHistory` now returns dropped count and auto-persists cleaned history, preventing repeated warnings on every request. Adds `SetHistory()` to session store.
- **PR #130** `fix/zalo-media-credentials` — Downloads Zalo CDN photos to temp files (auth-restricted/expiring URLs), adds `PhotoURL` field support, fixes `[]byte` credentials handling in `Update()`

### Deferred
- **Party Mode** (~20 files) — New feature, not a fix. Will be a separate PR when ready. Has open issues: missing authorization checks on session handlers, snake_case/camelCase event field mismatch.

Closing this PR in favor of the focused PRs above.

Issue #136, opened 2026-03-11
[BUG] Sandbox filesystem tools lack agent-specific path isolation (workspace escape)


Erudition:
## Description
Filesystem tools like `list_files`, `read_file`, `write_file`, and `edit` were incorrectly accessing the global `/workspace` mount in the sandbox instead of being confined to the agent's specific workspace subdirectory. This allowed agents to potentially "see" or access peer workspaces on the same host mount.

## Root Cause
The `executeInSandbox` implementations in these tools were not mapping the host paths to their corresponding container paths relative to the agent's subdirectory. While `ExecTool` had some mapping logic, it was inconsistent and prone to escape.

## Impact
- Information leak between agents sharing the same sandbox host volume.
- Functional errors when agents try to use paths they believe are absolute but aren't correctly rooted in the container.

## Proposed Fix
Implement a shared `MapHostPathToSandbox` utility to ensure all sandbox tools consistently resolve paths relative to the agent's specific container-side workspace directory.

thieung:
It's fixed in GoClaw 2.x

Issue #129, opened 2026-03-11
Hardcoded `"/workspace"` path in `executeInSandbox` ignores   `GOCLAW_SANDBOX_WORKSPACE_PATH`


Erudition:
# Bug: Hardcoded `"/workspace"` path in `executeInSandbox` ignores `GOCLAW_SANDBOX_WORKSPACE_PATH`

## Description
There is a logic error in the sandbox execution path where the container's working directory is hardcoded to `"/workspace"`. This causes OCI runtime errors (specifically `chdir` failures) when a user customizes their workspace path via `GOCLAW_WORKSPACE`, even if they correctly set `GOCLAW_SANDBOX_WORKSPACE_PATH` to match.

## Impacted Code
The issue is located in `internal/tools/shell.go`. The code calculates a relative path but then prepends a hardcoded string instead of using the configured sandbox directory.

https://github.com/nextlevelbuilder/goclaw/blob/main/internal/tools/shell.go#L332-L336

## Steps to Reproduce
1. Set `GOCLAW_WORKSPACE=/app/.goclaw` (or any non-default path).
2. Set `GOCLAW_SANDBOX_WORKSPACE_PATH=/.goclaw` to ensure the sandbox mount point matches the host-side configuration.
3. Trigger a tool call that uses the sandbox (e.g., `exec ls`).
4. The execution fails with:
   `OCI runtime exec failed: exec failed: unable to start container process: chdir to cwd ("/.goclaw/...") set in config.json failed: no such file or directory: unknown`

## Root Cause
The `executeInSandbox` function ignores the `SandboxContainerDir` property of the agent loop and instead uses a literal `"/workspace"` string when constructing the `containerCwd`. This results in a path that does not exist inside the sandbox container if the volume was mounted elsewhere.

## Suggested Fix
Update the `executeInSandbox` method in `internal/tools/shell.go` to use the dynamic `sandboxContainerDir` value provided in the agent's configuration.

thieung:
It's fixed in GoClaw 2.x

Issue #46, opened 2026-03-03
[BUG] LLM Workspace Path Mismatch Causes Access Errors in Multi-Tenant Setup


lehuubao2909:
The LLM detects the workspace path as:

```
~/.goclaw/workspace
```

However, the tools enforce a tenant-specific path:

```
~/.goclaw/workspace/6179755304/
```

Because of this mismatch, any file path generated by the LLM points to the root workspace directory instead of the enforced tenant subdirectory. As a result, all file operations fail with errors such as:

* `access denied`
* `no such file or directory`

### Steps to Reproduce

1. LLM generates a file path under `~/.goclaw/workspace`.
2. Tooling attempts to access files only within `~/.goclaw/workspace/<tenant_id>/`.
3. File operations fail due to path mismatch.

### Expected Behavior

The LLM should be aware of and operate within the enforced tenant-specific workspace directory, e.g.:

```
~/.goclaw/workspace/<tenant_id>/
```

### Suggested Fix

* Align the LLM workspace configuration with the enforced multi-tenant directory structure.
* Dynamically inject the tenant-specific workspace path into the LLM context.
* Ensure all file operations are constrained to the correct tenant directory.

### Additional Context

This issue impacts multi-tenant support and prevents correct file access when running in tenant-isolated environments. Adjusting the workspace resolution logic will improve multi-tenant reliability.

thieung:
It's fixed in GoClaw 2.x

PR #137, opened 2026-03-11
fix: sandbox path mounting and workspace isolation (#129, #136)

Erudition:
## Description
This PR addresses several critical issues related to sandbox path mounting and agent workspace isolation in Docker environments. Prior to these fixes, the sandbox was effectively unusable in many Docker-based deployments (such as Portainer) because hardcoded default paths (like `/.goclaw/...`) were not correctly mapped to persistent volumes, leading to `chdir` failures and sandbox startup errors.

### 1. Fix Docker Volume Mounting Issues (Issue #129)
- **Problem**: Sandbox containers were failing to start because they couldn't find the workspace directory. Inside Docker, `~` was expanding to ephemeral locations instead of the persistent `/app/workspace` mount.
- **Dynamic Path Resolution**: Implemented `resolveHostWorkspacePath` to dynamically discover the host volume or path for container directories like `/app/workspace`. This allows sibling containers in DooD (Docker-out-of-Docker) environments to mount the correct volumes.
- **Legacy Path Migration**: Added `MigrateLegacyPath` to automatically re-route old `~/.goclaw` paths to the persistent volume mount, ensuring backward compatibility.
- **Default Workspaces**: Updated the system to ensure new agents receive valid default workspace paths and existing agents with empty fields are proactively fixed.

### 2. Fix Sandbox Isolation Breach (Issue #136)
- **Problem**: Filesystem tools were defaulting to the global workspace root, allowing agents to potentially access files belonging to other agents.
- **Shared Path Mapping**: Introduced `MapHostPathToSandbox` utility to ensure all filesystem tools (`exec`, `read`, `write`, `list`, `edit`) consistently resolve paths relative to the agent's dedicated container-side subdirectory.
- **Escape Prevention**: Implemented strict checks to block any attempts to access paths outside the agent's specific subdirectory within the sandbox.

### 3. Nginx DNS Caching Fix
- Updated `ui/web/nginx.conf` to use Docker's internal DNS resolver (`127.0.0.11`) with dynamic upstream resolution. This prevents "Host unreachable" errors in the UI when the backend container is recreated with a new internal IP.

### 4. Build Fix
- Removed an unused `path/filepath` import in `internal/tools/shell.go`.

## Verification
- Verified successful compilation: `go build -o goclaw .`
- Verified path resolution logic in the sandbox manager.
- Manual verification of agent workspace isolation through tool calls.

### DIFF
--- a/internal/tools/context_keys.go
+++ b/internal/tools/context_keys.go
@@ -145,30 +145,7 @@ func ToolSandboxNetworkFromCtx(ctx context.Context) bool {
 	return v
 }
 
-// --- Vision / ImageGen config (per-agent overrides) ---
 
-const (
-	ctxVisionConfig   toolContextKey = "tool_vision_config"
-	ctxImageGenConfig toolContextKey = "tool_imagegen_config"
-)
-
-func WithVisionConfig(ctx context.Context, cfg *config.VisionConfig) context.Context {
-	return context.WithValue(ctx, ctxVisionConfig, cfg)
-}
-
-func VisionConfigFromCtx(ctx context.Context) *config.VisionConfig {
-	v, _ := ctx.Value(ctxVisionConfig).(*config.VisionConfig)
-	return v
-}
-
-func WithImageGenConfig(ctx context.Context, cfg *config.ImageGenConfig) context.Context {
-	return context.WithValue(ctx, ctxImageGenConfig, cfg)
-}
-
-func ImageGenConfigFromCtx(ctx context.Context) *config.ImageGenConfig {
-	v, _ := ctx.Value(ctxImageGenConfig).(*config.ImageGenConfig)
-	return v
-}
 
 // --- Builtin tool settings (global DB overrides) ---
 


Erudition:
Also touches on #46

viettranx:
Thanks for the thorough analysis in this PR — your identification of the sandbox isolation breach (#136) and DooD volume mounting issues (#129) was spot-on.

We've landed the fixes in a4a7a59 on main using a slightly different approach:

**Sandbox isolation**: Instead of `MapHostPathToSandbox` in every tool, we extracted `SandboxCwd()` / `ResolveSandboxPath()` helpers that map `ToolWorkspaceFromCtx(ctx)` to the correct container subdirectory. This keeps the global workspace mount (needed for team workspace + per-user workspace switching on cached containers) but scopes filesystem tool operations to the agent's subdirectory.

**DooD volume resolution**: Adopted your `resolveHostWorkspacePath` approach with enhanced container ID detection (parses `/proc/self/mountinfo` first, falls back to HOSTNAME/os.Hostname) and a 5-second timeout on `docker inspect`.

**Nginx DNS**: Adopted your resolver approach as-is — `resolver 127.0.0.11 valid=10s` with `$upstream_backend` variable.

**Sandbox hints**: Expanded the hint system from 1 pattern to 6 (permission denied, network disabled, read-only FS, missing file, resource limits) and added `MaybeFsBridgeHint()` for filesystem tools.

**MCP args**: Adopted space-separated parsing with improved regex that handles `--flag="value with spaces"` correctly.

Items we intentionally deferred:
- `MigrateLegacyPath` — current system handles `~` paths via `config.ExpandHome()`
- Per-request network override — network mode is agent-level config
- Dockerfile sudo — security concern for production

This PR can be closed. Thanks again for the detailed work! 🙏

PR #138, opened 2026-03-11
feat: add networking utilities and passwordless sudo to sandbox

Erudition:
...Because hey, it's a sandbox after all.

## Description
This PR upgrades the sandbox container image (`Dockerfile.sandbox`) to include essential networking utilities and dynamic tool installation capabilities.

### Changes
- **Networking Tools**: Added `iputils-ping` and `dnsutils`. This allows agents to use `ping`, `nslookup`, and `dig` for connection troubleshooting within the sandbox.
- **Sudo Access**: Added `sudo` and configured passwordless access for the `sandbox` user. This enables agents to dynamically install missing dependencies via `apt-get` (e.g., `sudo apt-get update && sudo apt-get install -y <package>`) without manual intervention.

## Verification
- Verified utilities are present in a test container build.
- Confirmed passwordless sudo access for the sandbox user.

### DIFF
--- a/Dockerfile.sandbox
+++ b/Dockerfile.sandbox
@@ -4,14 +4,19 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
-    bash \
-    ca-certificates \
-    curl \
-    git \
-    jq \
-    python3 \
-    python3-pip \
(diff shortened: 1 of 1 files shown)


Erudition:
Nevermind, installing packages still doesn't work since apt is on a read-only filesystem.

PR #144, opened 2026-03-11
fix: honor per-agent restrict_to_workspace from DB

Luvu182:
## Summary

- The per-agent `restrict_to_workspace` toggle (set via UI/API) was stored in the database but never propagated to the tool layer
- All file and exec tools used the global config default (`agents.defaults.restrict_to_workspace`) instead, so toggling the setting per-agent had no effect
- Fix: pass the DB value through context (same pattern as `WithToolWorkspace`) and have each tool check the context override before falling back to the constructor default

## Changes

- **`tools/context_keys.go`**: Add `ctxRestrictWs` context key with `WithRestrictToWorkspace()` / `RestrictFromCtx()` + `effectiveRestrict()` helper
- **`agent/loop_types.go`**: Add `RestrictToWs *bool` to `Loop` and `LoopConfig` (nil = use tool default)
- **`agent/loop.go`**: Inject per-agent value into context via `WithRestrictToWorkspace()` in `runLoop()`
- **`agent/resolver.go`**: Pass `ag.RestrictToWorkspace` from `AgentData` (DB) into `LoopConfig`
- **5 tools** (`read_file`, `write_file`, `list_files`, `edit`, `exec`): Use `effectiveRestrict(ctx, t.restrict)` instead of `t.restrict`

## How it works

```
DB (agents.restrict_to_workspace)
  → resolver.go: LoopConfig.RestrictToWs = &ag.RestrictToWorkspace
    → loop.go: ctx = WithRestrictToWorkspace(ctx, *l.restrictToWs)
      → tools: effectiveRestrict(ctx, t.restrict) → context override wins
```

Subagents inherit the parent's value via context. Delegated agents get their own DB value (resolved independently).

## Test plan

- [ ] Set `restrict_to_workspace = true` (default) on an agent → verify file tools reject paths outside workspace
- [ ] Toggle `restrict_to_workspace = false` on the same agent via UI → verify file tools allow paths like `/tmp`
- [ ] Verify other agents with `restrict_to_workspace = true` are still restricted
- [ ] Verify subagent inherits parent's restrict setting

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -73,6 +73,10 @@ func (l *Loop) runLoop(ctx context.Context, req RunRequest) (*RunResult, error) 
 	if req.ChannelType != "" {
 		ctx = tools.WithToolChannelType(ctx, req.ChannelType)
 	}
+	// Inject per-agent restrict_to_workspace override (from DB) so tools honor per-agent settings.
+	if l.restrictToWs != nil {
+		ctx = tools.WithRestrictToWorkspace(ctx, *l.restrictToWs)
+	}
 
 	// Per-user workspace isolation.
 	// Workspace path comes from user_agent_profiles (includes channel segment
--- a/internal/agent/loop_types.go
+++ b/internal/agent/loop_types.go
@@ -46,6 +46,7 @@ type Loop struct {
 	maxIterations int
 	maxToolCalls  int
 	workspace     string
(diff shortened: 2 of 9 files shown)


Luvu182:
Superseded by broader fix covering all per-agent config overrides (restrict + subagents + memory + sandbox).

PR #212, opened 2026-03-15
fix(subagent): detach async subagent context from parent run

duhd-vnpay:
## Summary

- **Bug**: Async subagents inherit parent run's Go context via `context.WithCancel(ctx)`. When the parent run completes (memory flush finishes), context cancellation cascades to the subagent, aborting its in-flight LLM call with `context canceled`.
- **Fix**: Use `context.WithoutCancel(ctx)` (Go 1.21+) to preserve all context values (tracing, store, tool keys) while detaching from the parent's cancellation chain. The subagent retains its own `cancelFunc` for explicit cancellation via steer/cancel commands.
- **Pattern**: Matches existing `DelegateAsync` which already uses `context.Background()` for detached context.

## Root Cause Analysis

Traced via GoClaw DB spans (trace `019cf090-ad79-71f2-9bd8-e41cf2652fd6`):

| Event | Timestamp |
|-------|-----------|
| Subagent spawned (LLM #1 starts) | `08:17:16.320` |
| Parent memory flush completes | `08:17:55.550` |
| Subagent LLM canceled | `08:17:55.573` (23ms after parent) |

```
Parent context (canceled on run completion)
  └── Subagent context (cascade canceled)  ← BUG
       └── LLM HTTP request (aborted → "context canceled")
```

## Change

```diff
-	taskCtx, taskCancel := context.WithCancel(ctx)
+	taskCtx, taskCancel := context.WithCancel(context.WithoutCancel(ctx))
```

`context.WithoutCancel` preserves all context values while removing the cancellation chain — cleaner than `context.Background()` + manual value propagation (which `DelegateAsync` does).

## Test plan

- [x] Spawn async subagent → verify it completes after parent run finishes
- [x] Cancel subagent via `steer` action → verify `cancelFunc` still works
- [x] Sync subagent (`RunSync`) unchanged — still uses parent context (correct, since parent blocks)
- [x] Verify tracing spans are properly linked (parent → subagent hierarchy)

### DIFF
--- a/internal/tools/subagent.go
+++ b/internal/tools/subagent.go
@@ -233,8 +233,10 @@ func (sm *SubagentManager) Spawn(
 		CreatedAt:         time.Now().UnixMilli(),
 		spawnConfig:       cfg,
 	}
-	// Create per-task context for real goroutine cancellation
-	taskCtx, taskCancel := context.WithCancel(ctx)
+	// Detach from parent context so async subagent survives parent completion.
+	// context.WithoutCancel preserves all context values (tracing, store, tools)
+	// while removing the parent's cancellation chain. Same pattern as DelegateAsync.
+	taskCtx, taskCancel := context.WithCancel(context.WithoutCancel(ctx))
 	subTask.cancelFunc = taskCancel
 
 	sm.tasks[id] = subTask


viettranx:
I've already fixed and removed some tools

PR #224, opened 2026-03-16
feat(agent): add tool call prefix stripping and assistant prefill removal for proxy provider compatibility  

xuandung38:
## Summary                                                                                                            
                                                                                                                     
  Two per-agent options to support proxy LLM providers (LiteLLM, OpenRouter, etc.):
                                               
  ### 1. Tool Call Prefix Stripping
  Proxy providers may prepend a prefix to tool call names returned by the model (e.g. `proxy_exec` instead of `exec`),
  causing tool policy validation failures and registry lookup errors.

  Per-agent `toolCallPrefix` config strips the prefix before registry lookup, permission checks, and
  `team_tasks`/`spawn` detection in both serial and parallel paths. Supports literal prefixes (`proxy_`) and template
  patterns (`proxy_{tool_name}`).

  ### 2. Strip Assistant Prefill
  Some models/providers reject requests ending with an assistant-role message (HTTP 400: "This model does not support
  assistant message prefill").

  Per-agent `strip_assistant_prefill` toggle removes trailing assistant messages before sending to the LLM. Configurable
   in LLM Configuration (General tab), stored in `other_config`.

  ### Additional Fix
  Config save handler now uses spread operator instead of hardcoded field list — new fields were silently dropped on
  save.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ k8s-*/*
 tests/**/creds.json
 ui/simple-saas
 *.tar
+# AI tool config directories (user-specific)
+.gemini/
+.opencode/


xuandung38:
Closed it because I felt this PR had too much influence, so I split it into two PRs 
#259 and #258 

Fix Issue : 240

PR #227, opened 2026-03-16
fix(routing): allow handoff routes to override channel instance agent

MarceloEmmerich:
## Problem

When a channel instance has an explicit `agent_id`, the Slack adapter always sets `msg.AgentID` on inbound messages. The handoff route check at `gateway_consumer_normal.go:47` has a guard `msg.AgentID == ""` that prevents handoff routes from ever activating for these channels.

This makes it impossible to route messages from different Slack channels to different agents through a single channel instance + handoff routes — which is the natural pattern for multi-agent setups where you want one Slack bot serving multiple agents in different channels.

## Fix

Remove the `msg.AgentID == ""` guard so handoff routes can override the channel's default agent. The handoff route lookup is a cheap DB call (already indexed by channel + chat_id), and returning `nil` when no route exists means the default agent is preserved.

## Before
```go
if teamStore != nil && msg.AgentID == "" {
```

## After
```go
if teamStore != nil {
```

## Use Case

Single Slack bot app with one Socket Mode connection. Mother agent handles DMs. Other agents (weather, invoices, etc.) handle their own `#cell-*` channels via handoff routes. No need for separate bot apps per agent.

### DIFF
--- a/cmd/gateway_consumer_normal.go
+++ b/cmd/gateway_consumer_normal.go
@@ -44,7 +44,7 @@ func processNormalMessage(
 	}
 
 	// Check handoff routing override
-	if teamStore != nil && msg.AgentID == "" {
+	if teamStore != nil {
 		if route, _ := teamStore.GetHandoffRoute(ctx, msg.Channel, msg.ChatID); route != nil {
 			agentID = route.ToAgentKey
 			slog.Info("inbound: handoff route active",


viettranx:
Hey — thanks for the PR and the detailed use case! The multi-agent Slack routing pattern you described (one Socket Mode connection, different agents per channel via handoff routes) is a valid scenario.

However, we've just removed the handoff tool and its infrastructure (handoff_routes table, HandoffRouteData, GetHandoffRoute, handleHandoffAnnounce) in commit 3f2b6e2 as part of a cleanup of deprecated delegation tools. The handoff_routes table and routing lookup no longer exist on main.

For the multi-agent routing use case, the recommended approach is to use **agent bindings** in your config — you can bind specific channels/chat IDs to specific agents without needing handoff routes. If agent bindings don't cover your use case, feel free to open a new issue describing the routing requirements and we can design a proper solution.

Closing this since the code it modifies has been removed. Sorry for the timing!

MarceloEmmerich:
No problem, and thanks for the heads up. Agent bindings are a much more elegant and clean solution!

PR #230, opened 2026-03-16
fix(providers): prevent Gemini thought_signature from leaking to other providers and breaking validation

badgerbees:
**Problem:**
When using Gemini models with extended thinking enabled, tool calls returned by the API include a specific `thought_signature`. These tool calls are saved to the persistent conversation database history. 

When the user switches the conversation to a non-Google provider (like DeepSeek, Mistral, Anthropic, or Groq) in the middle of a session, the previous conversation history is fed back into the new API to maintain context. However, non-Google APIs enforce strict JSON schema validation on historical tool calls and actively reject requests containing unknown fields like `thought_signature`. This causes immediate `422 Unprocessable Entity` or `400 Bad Request` exceptions.

**Why the old code didn't work:**
The old codebase only partially mitigated this issue. It checked if the provider was "gemini" at the *top* of [buildRequestBody](cci:1://file:///c:/Users/vyrgi/OneDrive/Documents/goclaw/internal/providers/anthropic_request.go:56:0-209:1) solely to trigger [collapseToolCallsWithoutSig()](cci:1://file:///c:/Users/vyrgi/OneDrive/Documents/goclaw/internal/providers/openai_gemini.go:4:0-69:1). But deeper down, during the actual JSON wire-format construction, it did an **unconditional injection**: 
```go
if sig := tc.Metadata["thought_signature"]; sig != "" {
    fn["thought_signature"] = sig // <-- Unconditional leak!
}
```
Because this injection wasn't gated, any non-Google provider iterating over a historical Gemini tool call would be forced to swallow a thought_signature key, crashing the API request.

**Fix/Solution:**
This PR introduces robust capability detection in [buildRequestBody](cci:1://file:///c:/Users/vyrgi/OneDrive/Documents/goclaw/internal/providers/openai.go:223:0-358:1) that evaluates precisely once per [Chat()](cci:1://file:///c:/Users/vyrgi/OneDrive/Documents/goclaw/internal/providers/anthropic.go:81:0-100:1) request to ensure extreme efficiency. 
We now calculate `supportsThoughtSignature` based on:
1. Provider `name` containing "gemini"
2. The `apiBase` URL containing "generativelanguage"
3. The actual `model` string (to catch Gemini proxy routes from LiteLLM/OpenRouter!).
Using this computation, we safely gate the serialization loop:
- **Gemini to Other Provider**: When building the request wire format, `fn["thought_signature"]` is entirely omitted for unsupported providers, creating clean standard tool calls that don't trigger 422 errors.
- **Other Provider to Gemini**: When going the reverse route, Gemini actually enforces that `thought_signature` MUST be present on older tool calls! Our logic successfully routes this edge-case into the pre-existing [collapseToolCallsWithoutSig()](cci:1://file:///c:/Users/vyrgi/OneDrive/Documents/goclaw/internal/providers/openai_gemini.go:4:0-69:1) function, stripping the tool blocks entirely and folding them into safe `user` blocks to keep Gemini happy.

### DIFF
--- a/internal/providers/openai.go
+++ b/internal/providers/openai.go
@@ -228,10 +228,13 @@ func (p *OpenAIProvider) buildRequestBody(model string, req ChatRequest, stream 
 	// Tool results are folded into plain user messages to preserve context.
 	inputMessages := req.Messages
 
-	// Compute provider capability once: does this endpoint support Google's extra content?
-	supportsExtraContent := strings.Contains(strings.ToLower(p.name), "gemini") || strings.Contains(strings.ToLower(p.apiBase), "generativelanguage")
+	// Compute provider capability once: does this endpoint support Google's thought_signature?
+	// We check name, apiBase, and the model string (which covers OpenRouter/LiteLLM routing to Gemini).
+	supportsThoughtSignature := strings.Contains(strings.ToLower(p.name), "gemini") ||
+		strings.Contains(strings.ToLower(p.apiBase), "generativelanguage") ||
+		strings.Contains(strings.ToLower(model), "gemini")
 
-	if supportsExtraContent {
+	if supportsThoughtSignature {
 		inputMessages = collapseToolCallsWithoutSig(inputMessages)
 	}
 
@@ -286,7 +289,7 @@ func (p *OpenAIProvider) buildRequestBody(model string, req ChatRequest, stream 
 				if sig := tc.Metadata["thought_signature"]; sig != "" {
 					// Only send thought_signature to providers that support it (Google/Gemini).
 					// Non-Google providers will reject the unknown field with 422 Unprocessable Entity.
-					if supportsExtraContent {
+					if supportsThoughtSignature {
 						fn["thought_signature"] = sig
 					}
 				}


nvtuan305:
This error still occurs in the latest code version @badgerbees 

```
LLM call failed (iteration 2): HTTP 400: goclaw-local: [{ "error": { "code": 400, "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:memory_search` , position 6. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.", "status": "INVALID_ARGUMENT" } } ]
```
Extended Thinking = OFF
image


image

badgerbees:
@nvtuan305  thank you for your comment, `Thinking Level = OFF` and the custom provider name `goclaw-local` were the two missing pieces that were breaking the logic.

Here’s what was happening chronologically:
1. The Naming Issue (A Blind Spot): In earlier versions, GoClaw's safety logic was blind unless the provider was named ‘gemini’. Because your provider is named `goclaw-local`, GoClaw thought it was talking to a standard OpenAI-compatible model (like DeepSeek or GROQ) and skipped the signature-repair logic entirely.

2. The "Thinking Level = OFF" Trap: Even after I added a check for the model name, I discovered a second hidden bug. When Extended Thinking is OFF, Gemini-3-Flash often emits "invalid" signatures that are just a single space, a newline, or a null value. The old code only looked for exactly "" (empty string), so it "missed" those whitespace-only signatures and decided not to collapse them.

Outcome: GoClaw mistakenly "kept" those broken tool calls in the history, but then rejected them at the wire as missing a valid thought_signature (hence the HTTP 400 error). I am currently writing a fix for this and putting out a PR asap

PR #273, opened 2026-03-19
fix: allow reusing agent_key after soft delete

### LINKED ISSUE #180: After deleting an agent on the web, the ID of that agent cannot be reused to create a new agent

### LINKED ISSUE #181: Can not recreate the same agent key after delete
Image

Luvu182:
## Summary
- Replace hard `UNIQUE` constraint on `agents.agent_key` with a partial unique index (`WHERE deleted_at IS NULL`)
- This allows creating a new agent with the same key after the original has been soft-deleted
- Added migration 000023 (up + down) and bumped `RequiredSchemaVersion` to 23

**Root cause:** The `agent_key` column had a plain `UNIQUE` constraint that included soft-deleted rows. Since `DELETE` only sets `deleted_at = NOW()`, the old row still occupied the unique slot, blocking reuse.

Fixes #180, fixes #181

## Test plan
- [ ] Run `./goclaw migrate up` — migration 23 applies cleanly
- [ ] Create agent with key `test-agent` → succeeds
- [ ] Delete agent → soft-deleted (`deleted_at` set)
- [ ] Create new agent with key `test-agent` → succeeds (previously failed with unique violation)
- [ ] Run `./goclaw migrate down 1` — rollback restores original constraint

### DIFF
--- a/internal/upgrade/version.go
+++ b/internal/upgrade/version.go
@@ -2,4 +2,4 @@ package upgrade
 
 // RequiredSchemaVersion is the schema migration version this binary requires.
 // Bump this whenever adding a new SQL migration file.
-const RequiredSchemaVersion uint = 22
+const RequiredSchemaVersion uint = 23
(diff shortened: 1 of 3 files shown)


viettranx:
Cái này để a xử lý cách khác nha

PR #290, opened 2026-03-19
fix: normalize delete confirmation matching

nguyennguyenit:
## Summary
- normalize delete confirmation input with NFC, trim, and case-insensitive matching
- trim the displayed confirmation target and placeholder for consistency

## Testing
- pnpm build


nguyennguyenit:
Screenshot 2026-03-20 at 02 15 42
Ví dụ tên Agent là Trà thì khi gõ Trà vô sẽ không delete được

PR #329, opened 2026-03-21
perf: wire embedding_cache table into IndexDocument hot path

### LINKED ISSUE #326: perf: embedding_cache table exists but is never used in IndexDocument hot path
## Summary

The `embedding_cache` table is defined in the schema (`migrations/000001`) with a composite primary key `(hash, provider, model)`, but `PGMemoryStore.IndexDocument()` in `internal/store/pg/memory_docs.go` never checks this cache before calling `s.provider.Embed(ctx, texts)`.

## Current Behavior

Every call to `IndexDocument()` regenerates embeddings for **all chunks**, even if the content hasn't changed:

```go
// memory_docs.go ~line 189
embeddings, embErr = s.provider.Embed(ctx, texts)
```

The content hash is already computed per chunk (`memory.ContentHash(tc.Text)` at ~line 198), but it's only used as a DB column value — never as a cache lookup key.

## Proposed Fix

Before calling the embedding API, batch-lookup existing embeddings by `(hash, provider, model)`:

```go
// 1. Compute hashes for all chunks
hashes := make([]string, len(chunks))
for i, c := range chunks {
    hashes[i] = memory.ContentHash(c.Text)
}

// 2. Batch lookup cached embeddings
cached, err := s.lookupEmbeddingCache(ctx, hashes, s.provider.Name(), s.provider.Model())

// 3. Only embed uncached texts
var uncachedTexts []string
var uncachedIndices []int
for i, h := range hashes {
    if _, ok := cached[h]; !ok {
        uncachedTexts = append(uncachedTexts, chunks[i].Text)
        uncachedIndices = append(uncachedIndices, i)
    }
}

// 4. Call API only for uncached
if len(uncachedTexts) > 0 {
    freshEmbeddings, err := s.provider.Embed(ctx, uncachedTexts)
    // merge into cached map + write back to embedding_cache
}
```

## Impact

- **Cost reduction**: Embedding API calls are paid per token. Documents that haven't changed (common during `IndexAll` or re-indexing) currently waste API calls.
- **Latency**: Skipping API calls for cached chunks speeds up indexing significantly.
- **The table already exists** — this is just wiring it into the hot path.

## Files to Modify

- `internal/store/pg/memory_docs.go` — `IndexDocument()`: add cache lookup + write-back
- Optionally add a `PruneEmbeddingCache()` method to clean stale entries

## References

- Schema: `migrations/000001_init_schema.up.sql` lines 200-209 (`embedding_cache` table)
- Hash function: `internal/memory/chunk.go` — `ContentHash()`
- Provider interface: `internal/memory/embeddings.go` — `Name()`, `Model()`, `Embed()`

imtylervo:
## Summary

- `embedding_cache` table exists in schema (migration 000001) but `IndexDocument()` never uses it
- Every re-index calls the embedding API for **all chunks**, even unchanged content
- This PR adds cache lookup before API call, writes fresh results back, and merges cached + fresh vectors

## Changes

- **`internal/store/pg/memory_embedding_cache.go`** (new): `lookupEmbeddingCache()`, `writeEmbeddingCache()`, `parseVector()` 
- **`internal/store/pg/memory_docs.go`**: Modified `IndexDocument()` to check cache first, only embed uncached chunks

## How it works

1. Compute content hashes for all chunks (`memory.ContentHash()`)
2. Batch SELECT from `embedding_cache` by `(hash, provider, model)` 
3. Only call `s.provider.Embed()` for uncached texts
4. UPSERT fresh embeddings back to cache
5. Merge cached + fresh into final slice

## Error handling

All cache operations are graceful — if lookup fails, falls through to full API call. If write fails, indexing continues normally.

## Impact

- Reduces embedding API cost on re-indexing unchanged documents
- Speeds up `IndexAll()` and document re-indexing

Closes #326

### DIFF
--- a/internal/store/pg/memory_docs.go
+++ b/internal/store/pg/memory_docs.go
@@ -178,18 +178,88 @@ func (s *PGMemoryStore) IndexDocument(ctx context.Context, agentID, userID, path
 		return nil
 	}
 
-	// Generate embeddings
+	// Generate embeddings with cache
 	var embeddings [][]float32
 	if s.provider != nil {
-		texts := make([]string, len(chunks))
+		providerName := s.provider.Name()
+		providerModel := s.provider.Model()
+
+		// Compute content hashes for all chunks
+		hashes := make([]string, len(chunks))
 		for i, c := range chunks {
-			texts[i] = c.Text
+			hashes[i] = memory.ContentHash(c.Text)
 		}
-		var embErr error
-		embeddings, embErr = s.provider.Embed(ctx, texts)
-		if embErr != nil {
-			slog.Warn("memory embedding failed, storing chunks without vectors",
-				"path", path, "chunks", len(chunks), "error", embErr)
+
+		// Batch lookup cached embeddings
+		cached, cacheErr := s.lookupEmbeddingCache(ctx, hashes, providerName, providerModel)
+		if cacheErr != nil {
+			slog.Warn("embedding cache lookup failed, falling back to full API call",
+				"path", path, "error", cacheErr)
+			cached = nil
+		}
+
+		// Determine which chunks need fresh embeddings
+		var uncachedIdxs []int
+		var uncachedTexts []string
+		for i, c := range chunks {
+			if cached != nil {
+				if _, ok := cached[hashes[i]]; ok {
+					continue
+				}
+			}
+			uncachedIdxs = append(uncachedIdxs, i)
+			uncachedTexts = append(uncachedTexts, c.Text)
+		}
+
+		if len(cached) > 0 {
+			slog.Info("embedding cache hit",
+				"path", path, "cached", len(cached), "uncached", len(uncachedTexts))
+		}
+
+		// Call embedding API only for uncached texts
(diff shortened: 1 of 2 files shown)


viettranx:
Thanks for this PR @imtylervo! Great idea wiring up the `embedding_cache` table — it was sitting unused since migration 000001.

We've incorporated your work and addressed a few additional improvements in a separate commit on `main` (69107eb):

1. **Removed `lib/pq` dependency** — replaced `pq.Array()` with dynamic positional params to stay consistent with the codebase's `pgx/v5` usage
2. **Batch INSERT** — single query with 100 rows/batch instead of N+1 round-trips
3. **Dimension mismatch handling** — gracefully skips cache write when vector dimensions don't match schema (`vector(1536)`), so indexing continues normally
4. **`parseVector` hardening** — handles both `[]` and `()` bracket formats
5. **Log level tuning** — cache hit log downgraded to `Debug`, added warning when embedding API returns mismatched count

Thanks again for the contribution!

PR #339, opened 2026-03-21
build(docker): add curl to full runtime image

tolkonepiu:
## What
- add `curl` to the full Docker runtime image

## Why
Some skills expect `curl` to be available out of the box.


mrgoonie:
@claude review

PR #346, opened 2026-03-22
fix(zalo): allow QR session restart instead of blocking retry

ducky-duke:
## Summary
- When a user closes the Zalo QR login dialog before scanning, the old session goroutine keeps running (2 min timeout), blocking any retry with "QR session already active for this instance"
- Fix: store `context.CancelFunc` in `activeSessions` instead of `struct{}`, cancel the previous session when a new request arrives for the same instance
- Moves the 2-minute timeout context creation from `runQRFlow` to `handleQRStart` so it can be cancelled externally

## Test plan
- [x] Start QR login flow for a Zalo Personal instance
- [x] Close the QR dialog without scanning
- [x] Immediately request QR login again — should succeed without error
- [x] Verify that scanning QR on retry completes login successfully

### DIFF
--- a/internal/channels/zalo/personal/zalomethods/qr.go
+++ b/internal/channels/zalo/personal/zalomethods/qr.go
@@ -22,7 +22,7 @@ import (
 type QRMethods struct {
 	instanceStore  store.ChannelInstanceStore
 	msgBus         *bus.MessageBus
-	activeSessions sync.Map // instanceID (string) -> struct{}
+	activeSessions sync.Map // instanceID (string) -> context.CancelFunc
 }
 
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

ducky-duke:
Fixed the **runtime panic bug** introduced by @viettranx in commit `6979bf0`.

`context.CancelFunc` is `func()` — a non-comparable type in Go. `sync.Map.CompareAndDelete` uses `==` internally, which panics when comparing function values. Wrapped the cancel func in a `*qrSession` struct pointer (comparable by address) to fix this.

See commit `df25488`.

ducky-duke:
Should I resolve conflicts by myself ?

Issue #349, opened 2026-03-22
Bug listing providers and not generate session key when click to "chat" in UI


anhle128:
- Go to providers in UI got error
- Click to "Chat" in navbar cannot send chat because not generate right URL with valid session key

PR #350, opened 2026-03-22
fix: listing providers error and chat session key generation

anhle128:
## Summary
- Fix providers list page error in UI by correcting the provider type field mapping
- Fix chat navigation to properly generate session URL with valid session key
- Update API constants to use correct endpoint path

Closes https://github.com/nextlevelbuilder/goclaw/issues/349

## Test plan
- [ ] Navigate to Providers page — verify list loads without error
- [ ] Click "Chat" in navbar — verify redirected to chat with valid session key in URL
- [ ] Verify chat messages send successfully after navigation

### DIFF
--- a/ui/web/src/pages/chat/hooks/use-chat-messages.ts
+++ b/ui/web/src/pages/chat/hooks/use-chat-messages.ts
@@ -59,6 +59,11 @@ export function useChatMessages(sessionKey: string, agentId: string) {
     toolStreamRef.current = [];
     activityRef.current = null;
     blockRepliesRef.current = [];
+    // Clear messages when navigating away from a session (empty key).
+    // When switching to another session, loadHistory() will replace them.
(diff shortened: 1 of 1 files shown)


viettranx:
Closing this PR — the `main` branch has diverged significantly since this was opened (15+ commits touching the same files, including the multi-tenant isolation implementation in #359).

The bugs identified here are real and still exist on `main`:
1. **Nested `<button>` in provider-list-row** — invalid HTML causing potential rendering/accessibility issues
2. **Stale `agentId` state in chat-page** — `useState` initializer only runs once, doesn't re-derive from URL on session switch
3. **Premature session creation in `handleAgentChange`** — generates ghost sessions before user sends a message

These will be fixed directly on `main` to avoid merge conflicts. Thanks for identifying the issues.

PR #352, opened 2026-03-22
fix(ui): fix chat nav route and crypto.randomUUID in non-secure contexts

maxfraieho:
## Summary

- **sidebar nav**: `ROUTES.CHAT` (`/chat/:sessionKey?`) was used as a navigation target, causing the browser to literally navigate to `/chat/:sessionKey?` instead of `/chat`. Fixed by using the hardcoded `"/chat"` string.
- **`crypto.randomUUID` crash**: `buildNewSessionKey()`, `handleAgentChange()`, and `skill-upload-dialog` called bare `crypto.randomUUID()` which is only available in secure contexts (HTTPS/localhost). On HTTP the call threw `TypeError: crypto.randomUUID is not a function`, silently aborting the send. Fixed by replacing all calls with `generateId()` from `@/lib/utils` which already had the correct fallback.
- **`use-chat-send` UX**: split the silent `!ws.isConnected || ...` guard into a separate branch that sets a visible error message so users see feedback when the WebSocket is not ready.

## Test Plan

- [x] Build passes: `pnpm build` ✓
- [x] Navigating to Chat from sidebar opens `/chat` (not `/chat/:sessionKey?`)
- [x] Typing a message and clicking Send on HTTP successfully calls `chat.send` (verified via server logs)
- [x] New chat session key generated correctly on HTTP deployment

### DIFF
--- a/ui/web/src/components/layout/sidebar.tsx
+++ b/ui/web/src/components/layout/sidebar.tsx
@@ -73,7 +73,7 @@ export function Sidebar({ collapsed, onNavItemClick }: SidebarProps) {
       <nav className="flex-1 space-y-4 overflow-y-auto px-2 py-4">
         <SidebarGroup label={t("groups.core")} collapsed={collapsed}>
           <SidebarItem to={ROUTES.OVERVIEW} icon={LayoutDashboard} label={t("nav.overview")} collapsed={collapsed} />
-          <SidebarItem to={ROUTES.CHAT} icon={MessageSquare} label={t("nav.chat")} collapsed={collapsed} />
+          <SidebarItem to="/chat" icon={MessageSquare} label={t("nav.chat")} collapsed={collapsed} />
           <SidebarItem to={ROUTES.AGENTS} icon={Bot} label={t("nav.agents")} collapsed={collapsed} />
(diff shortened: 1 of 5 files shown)


viettranx:
Thanks for identifying these issues! After reviewing, here's what we found:

**2 of 5 changes already exist on `main`:**
- Sidebar nav: `ROUTES.CHAT` → `"/chat"` — already fixed
- `chat-page.tsx` `handleAgentChange`: already refactored, no longer uses `crypto.randomUUID()`

**3 remaining fixes applied directly on `main`** (commit incoming):
1. `use-chat-sessions.ts`: `crypto.randomUUID()` → `uniqueId()`
2. `skill-upload-dialog.tsx`: `crypto.randomUUID()` → `uniqueId()`
3. `use-chat-send.ts`: added visible error message when WS is not connected (with i18n support for en/vi/zh)

**Important note on the fix approach:** This PR used `generateId()` which produces `req-N-timestamp` format — but session keys require UUID format (`agent:X:ws:direct:<UUID>`). We used `uniqueId()` from `@/lib/utils` instead, which wraps `crypto.randomUUID()` with a proper UUID v4 fallback for non-secure contexts.

Closing in favor of the direct fix on `main`. Thanks again for the bug report!

PR #358, opened 2026-03-23
fix(mcp): prevent stdio command field from being sent as a single exe…

thinhfromtop:
…cutable path

Issue:
When users entered a full command with arguments in the MCP form's Command field (e.g. "npx -y @henkey/postgres-mcp-server"), the entire string was sent as-is to `exec.Command()`, causing a `fork/exec` error:

  fork/exec npx -y @henkey/postgres-mcp-server: no such file or directory

The OS treats the whole string as a single binary path instead of splitting it into command + arguments. This was compounded by a misleading placeholder ("npx -y @modelcontextprotocol/server-everything") that encouraged users to paste the full command into one field.

Additionally, the Args field only supported space-separated values, but users naturally entered comma-separated values (e.g. "-y, @foo/bar, --connection-string, postgres://...") which were passed with literal commas.

Fix:
- Auto-split Command field: if user enters multiple tokens (e.g. "npx -y @foo/bar"), extract first token as command, prepend the rest to Args.
- Extract shared `splitShellTokens()` helper that handles both comma and space delimiters while respecting quoted strings.
- Update Command placeholder from full command example to just "npx" to avoid misleading users.

### DIFF
--- a/ui/web/src/pages/mcp/mcp-form-dialog.tsx
+++ b/ui/web/src/pages/mcp/mcp-form-dialog.tsx
@@ -33,6 +33,20 @@ interface MCPFormDialogProps {
   onTest: (data: { transport: string; command?: string; args?: string[]; url?: string; headers?: Record<string, string>; env?: Record<string, string> }) => Promise<{ success: boolean; tool_count?: number; error?: string }>;
 }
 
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

PR #370, opened 2026-03-23
fix: address 5 bugs from issue #294 — stale context, cron reliability…

### LINKED ISSUE #294: bug: user_context_files becomes stale after agent_context_files update — seeding trap breaks per-user prompts
## Summary

When updating agent system prompts via `agent_context_files`, existing users' `user_context_files` remain unchanged. GoClaw reads **only** `user_context_files` at runtime for predefined agents, so prompt updates are invisible to existing users until manually propagated via direct SQL.

### Root Cause — Seeding Trap

GoClaw's context file architecture has a **one-way seeding** design:

1. On first user message → GoClaw copies `agent_context_files` → `user_context_files`
2. After seeding, **per-user copies are independent** — no sync mechanism exists
3. Updating `agent_context_files` (via Dashboard, WebSocket `agents.files.set`, or SQL) does **NOT** propagate to existing `user_context_files`

### Observed Impact

After updating `agent_context_files` with new instructions (e.g., adding tool usage rules, new command handlers), existing users continue receiving the old system prompt. The operator must manually run SQL to propagate changes — there is no API or automatic mechanism.

---

### 1. 🔴 No API to propagate `agent_context_files` → `user_context_files`

**Current workaround:** Direct SQL `UPDATE` on the database. No REST or WebSocket API exists to propagate agent-level file changes to existing per-user copies.

**Suggested fix:** Add a propagation option, e.g.:
- `agents.files.set` with `propagate: true` param
- `PUT /v1/agents/{id}/files/{name}?propagate=true`
- Auto-propagate with opt-out flag

```sql
-- Current manual workaround (requires DB access):
UPDATE user_context_files SET content = (
  SELECT content FROM agent_context_files 
  WHERE agent_id = '<AGENT_ID>' AND file_name = '<FILE>'
), updated_at = NOW()
WHERE agent_id = '<AGENT_ID>' AND file_name = '<FILE>';
```

### 2. 🔴 Cron session reuse causes agent to fabricate stale errors

**Problem:** Cron runs reuse sessions. If a previous run encountered a tool error, the agent sees the error in chat history on the next run and **skips the tool call entirely**, fabricating the same error response without actually invoking the tool.

**Impact:** Cron jobs silently produce incorrect output even after the underlying issue is fixed. Requires manual `DELETE FROM sessions` to recover.

**Suggested fix:** Cron runs should use isolated/fresh sessions, or provide a configurable `fresh_session: true` option per cron job.

### 3. 🟡 In-memory session write-behind cache → data loss on crash

**Problem:** Sessions use write-behind cache — all reads/writes happen in-memory, data flushes to DB only when `Save()` is called at end of agent run. Container crash mid-run → **unsaved messages permanently lost**. Direct SQL queries also return stale data during active runs.

**Suggested fix:** Provide option for write-through caching or configurable periodic flush intervals.

### 4. 🟡 Cron delivery silently fails when payload fields are missing

**Problem:** When `cron_jobs.payload` is missing any of the three required delivery fields (`deliver`, `channel`, `to`), `gateway_cron.go` silently skips `msgBus.PublishOutbound()` — **no error logged**, no warning, no notification. The cron job appears successful in `cron_run_logs` but the user never receives the output.

**Suggested fix:** Validate required delivery fields on `cron.create`. Log a warning when delivery is configured but required fields are missing.

### 5. 🟡 `paired_devices` in-memory cache not invalidated by direct SQL

**Problem:** GoClaw maintains an in-memory cache for paired devices. Direct SQL `INSERT INTO paired_devices` does not update the cache, so `IsPaired()` returns `false` even though the DB record exists. Only WebSocket `device.pair.approve` correctly updates both DB and cache.

**Suggested fix:** Add `LISTEN/NOTIFY` mechanism on the table, or surface a clearer error message when pairing state is inconsistent.

---

## Reproduction Steps (Primary Issue)

1. Deploy GoClaw with a predefined agent
2. A user sends their first message → `user_context_files` seeded from `agent_context_files`
3. Update `agent_context_files` via Dashboard UI, `agents.files.set` WS call, or direct SQL
4. Same user sends another message → agent uses **old** `user_context_files` content
5. Verify:
   ```sql
   SELECT file_name, length(content) FROM agent_context_files WHERE agent_id='<ID>';
   SELECT file_name, length(content) FROM user_context_files WHERE agent_id='<ID>';
   -- Sizes will differ — user_context_files has stale content
   ```

## Expected Behavior

Updating `agent_context_files` should provide a way to propagate to existing users. Suggested approaches:

- **Option A:** Auto-propagate to all `user_context_files` on update (with opt-out flag for customized per-user files)
- **Option B:** API endpoint with `propagate=true` parameter
- **Option C:** Version field — if agent file version > user file version, re-seed on next message

## Environment

- GoClaw Gateway (Go 1.25)
- PostgreSQL 18 + pgvector
- Agent type: `predefined`
- Affected tables: `agent_context_files`, `user_context_files`

thinhfromtop:
…, data safety

Fixes #294

1. Stale user_context_files after agent file update (#294.1, #294.2)
   - Added PropagateContextFile() to AgentStore — copies agent-level file to all existing user instances in one UPDATE
   - Added `propagate` param to agents.files.set WS handler
   - Response includes `propagated` count

2. Cron session reuse — tool errors block future runs (#294.3)
   - Reset cron session before each run via sessionMgr.Reset()
   - Each run starts clean, no stale error context

3. Write-behind cache risk — crash loses messages (#294.4)
   - Added periodic checkpoint flush after each tool iteration
   - Messages persisted to session after every tool call round
   - Worst case on crash: only current tool call lost, not entire run

4. Silent cron delivery failures (#294.5)
   - Added slog.Warn when deliver=true but channel/chatID missing

5. paired_devices cache gap (#294.6) — no fix needed
   - IsPaired() queries DB directly, no stale cache possible

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -761,7 +761,7 @@ func runGateway() {
 	defer sched.Stop()
 
 	// Start cron service with job handler (routes through scheduler's cron lane)
-	pgStores.Cron.SetOnJob(makeCronJobHandler(sched, msgBus, cfg, channelMgr))
+	pgStores.Cron.SetOnJob(makeCronJobHandler(sched, msgBus, cfg, channelMgr, pgStores.Sessions))
 	pgStores.Cron.SetOnEvent(func(event store.CronEvent) {
 		server.BroadcastEvent(*protocol.NewEvent(protocol.EventCron, event))
 	})
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/nextlevelbuilder/goclaw/internal/agent"
@@ -21,7 +22,7 @@ import (
 // Safe because cron jobs only fire after Start(), well after this is set.
 var cronHeartbeatWakeFn func(agentID string)
 
-func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg *config.Config, channelMgr *channels.Manager) func(job *store.CronJob) (*store.CronJobResult, error) {
+func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg *config.Config, channelMgr *channels.Manager, sessionMgr store.SessionStore) func(job *store.CronJob) (*store.CronJobResult, error) {
 	return func(job *store.CronJob) (*store.CronJobResult, error) {
 		agentID := job.AgentID
 		if agentID == "" {
@@ -31,6 +32,11 @@ func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg 
 		}
 
 		sessionKey := sessions.BuildCronSessionKey(agentID, job.ID)
+
+		// Reset session before each cron run to prevent tool errors from previous
+		// runs from polluting the context and blocking future executions (#294).
+		sessionMgr.Reset(context.Background(), sessionKey)
+
 		channel := job.Payload.Channel
 		if channel == "" {
 			channel = "cron"
@@ -99,6 +105,9 @@ func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg 
 			}
 			appendMediaToOutbound(&outMsg, result.Media)
 			msgBus.PublishOutbound(outMsg)
+		} else if job.Payload.Deliver {
+			slog.Warn("cron: delivery configured but channel/chatID missing — output discarded",
+				"job_id", job.ID, "job_name", job.Name, "channel", job.Payload.Channel, "to", job.Payload.To)
 		}
 
 		cronResult := &store.CronJobResult{
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1329,6 +1329,16 @@ func (l *Loop) runLoop(ctx context.Context, req RunRequest) (*RunResult, error) 
 			pendingMsgs = append(pendingMsgs, forSession...)
 		}
 
+		// Periodic checkpoint: flush pending messages to session after each tool
+		// iteration to prevent data loss on container crash (#294).
+		if len(pendingMsgs) > 0 {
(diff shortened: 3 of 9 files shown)


viettranx:
## Review — Superseded by direct fix on main

Cảm ơn effort fix 5 bugs từ issue #294. Đã review kỹ và tìm thấy một số vấn đề thiết kế cần cải thiện:

### Issues tìm được trong PR

| # | Mức độ | Fix | Vấn đề |
|---|--------|-----|--------|
| 1 | MEDIUM | PropagateContextFile | Comment nói "Only updates users who already have that file" nhưng SQL thiếu `AND file_name` trong WHERE → update tất cả user instances thay vì chỉ users có file đó |
| 2 | LOW | PropagateContextFile | 2 queries riêng biệt (SELECT + UPDATE) có thể gộp thành single CTE |
| 3 | **HIGH** | Checkpoint flush | Phá vỡ atomic flush design — code hiện tại flush ALL messages cuối run, comment ghi rõ "ensures concurrent runs never see each other's in-progress messages". Checkpoint mỗi iteration phá guarantee này |
| 4 | **HIGH** | Checkpoint flush | Token calibration sai — `pendingMsgs[:0]` reset buffer, nhưng `len(pendingMsgs)` ở cuối dùng để calibrate tokens sẽ under-count |
| 5 | MEDIUM | Checkpoint flush | Performance — flush mỗi iteration tốn quá (20+ DB writes thay vì 1) |
| 6 | MEDIUM | Cron Reset | Dùng `context.Background()` thay vì `cronCtx` (đã có tenant scope) cho Reset call |

### Đã fix trực tiếp trên main với cải tiến:
- **PropagateContextFile:** Single CTE query, WHERE clause đúng (`file_name = $2`)
- **Checkpoint flush:** Interval-based (mỗi 5 iterations) thay vì mỗi iteration + track `checkpointFlushedMsgs` cho token calibration
- **Cron Reset:** Dùng `cronCtx` (tenant-scoped) thay vì `context.Background()`
- **Delivery warning:** Giữ nguyên approach (đúng rồi)

Closing PR — fixes đã commit trực tiếp trên main.

PR #386, opened 2026-03-23
fix(providers): register Ollama provider in-memory and fix double /v1

hoangthuc701:
## Summary
- Add `ProviderOllama` case to `registerInMemory()` so Ollama providers created via UI are immediately usable without gateway restart
- Fix double `/v1` in `registerProvidersFromDB()` — UI sends full URL (`http://localhost:11434/v1`) but backend appended another `/v1`
- Add `ProviderOllama` case to `handleListProviderModels()` before the API key guard so model listing works for keyless Ollama

## Test plan
- [ ] Create Ollama provider via UI with default API base → verify succeeds without restart
- [ ] List models for Ollama provider → returns installed models
- [ ] Existing providers (OpenAI, Anthropic, etc.) unaffected — no logic changes for other types

### DIFF
--- a/cmd/gateway_providers.go
+++ b/cmd/gateway_providers.go
@@ -275,12 +275,13 @@ func registerProvidersFromDB(registry *providers.Registry, provStore store.Provi
 			continue
 		}
 		// Local Ollama requires no API key — handle before the key guard (same pattern as ClaudeCLI).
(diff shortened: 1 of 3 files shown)


mrgoonie:
@claude review

Issue #306, opened 2026-03-20
security: hardening audit — 10 findings (1 critical, 3 high, 3 medium, 3 low)


t1d38898:
## Security Audit Summary

Community security review of GoClaw codebase. 10 findings across secrets management, gateway hardening, and defense-in-depth.

| # | Severity | Finding |
|---|----------|---------|
| C1 | 🔴 Critical | `.env` has no file-permission hardening |
| H1 | 🟠 High | Default Postgres password in Docker Compose |
| H2 | 🟠 High | CORS allow-all fallback in WebSocket upgrader |
| H3 | 🟠 High | MCP bridge unauthenticated when no gateway token |
| M1 | 🟡 Medium | Config struct stores channel tokens in plain memory |
| M2 | 🟡 Medium | No TLS enforcement for database connection |
| M3 | 🟡 Medium | Postgres DSN length logged (information disclosure) |
| L1 | 🔵 Low | Health endpoint exposes protocol version |
| L2 | 🔵 Low | Prompt injection guard default is "warn" not "block" |
| L3 | 🔵 Low | Client IP extraction trusts proxy headers unconditionally |

---

## 🔴 C1. `.env` File Has No File-Permission Hardening

**File:** `.env`, `prepare-env.sh`

`.env` stores gateway token, encryption key, and LLM API keys in plaintext. `.gitignore` blocks commits, but no `chmod 600` enforcement or secrets manager integration exists.

**Recommendation:**
```bash
# In prepare-env.sh, after generating .env:
chmod 600 .env
```

---

## 🟠 H1. Default Postgres Password in Docker Compose

**File:** `docker-compose.postgres.yml:17,33`, `docker-compose.upgrade.yml:27`

Default password `goclaw` via `${POSTGRES_PASSWORD:-goclaw}`. If `.env` is missing, DB runs with weak known password.

**Recommendation:**
```yaml
POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:?Set POSTGRES_PASSWORD in .env}
```

---

## 🟠 H2. CORS Allow-All Fallback in WebSocket Upgrader

**File:** `internal/gateway/server.go` — `checkOrigin()`

When `gateway.allowed_origins` is empty, all origins are accepted. Enables Cross-Site WebSocket Hijacking (CSWSH).

**Recommendation:** Fail-closed when no whitelist configured:
```go
if len(allowed) == 0 {
    slog.Warn("security.cors_no_whitelist", "origin", origin)
[...]

PR #389, opened 2026-03-23
fix(security): harden secrets, MCP bridge auth, and file permissions

hoangthuc701:
## Summary

Addresses actionable findings from security audit (#306):

- **H1: Default Postgres password** — `POSTGRES_PASSWORD` now uses `:?` (fail-fast) instead of `:-goclaw` in all compose files. `prepare-env.sh` auto-generates it so existing workflows aren't broken.
- **H3: MCP bridge unauthenticated** — Bridge returns 403 when no gateway token configured instead of serving tools without auth.
- **C1: `.env` file permissions** — `prepare-env.sh` now sets `chmod 600` on `.env` after generation (skipped on Windows).
- **H2: CORS allow-all** — Added startup warning when no `allowed_origins` configured (warn-only, not fail-closed, to avoid breaking existing setups).

## Findings intentionally skipped

| Finding | Reason |
|---------|--------|
| M1 (tokens in memory) | Security theater — all processes hold secrets in memory at runtime; already encrypted at-rest |
| M2 (no DB TLS) | Only in docker-compose (same network); remote users set sslmode themselves |
| M3 (DSN length logged) | A number reveals nothing actionable |
| L1 (protocol version) | Not a software version, useless for fingerprinting |
| L2 (guard default warn) | Design choice for single-user deployments; users can set "block" |
| L3 (proxy headers) | Edge case best handled at reverse proxy layer |

## Test plan

- [ ] Run `./prepare-env.sh` — verify `POSTGRES_PASSWORD` is generated and `.env` has 600 perms
- [ ] `docker compose -f docker-compose.yml -f docker-compose.postgres.yml up` without `.env` — verify fail-fast error for missing `POSTGRES_PASSWORD`
- [ ] Start gateway without `GOCLAW_GATEWAY_TOKEN` — verify `/mcp/bridge` returns 403
- [ ] Start gateway without `allowed_origins` — verify startup log shows CORS warning
- [ ] `go build ./internal/gateway/...` and `go vet ./internal/gateway/...` pass

### DIFF
--- a/docker-compose.postgres.yml
+++ b/docker-compose.postgres.yml
@@ -14,7 +14,7 @@ services:
       - "${POSTGRES_PORT:-5432}:5432"
     environment:
       POSTGRES_USER: ${POSTGRES_USER:-goclaw}
-      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-goclaw}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:?Set POSTGRES_PASSWORD in .env (run ./prepare-env.sh)}
       POSTGRES_DB: ${POSTGRES_DB:-goclaw}
     volumes:
       - postgres-data:/var/lib/postgresql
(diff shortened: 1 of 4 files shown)


viettranx:
Thanks for this PR — the security findings are well-prioritized and the skip rationale is solid.

However, we identified a **breaking change** with the `POSTGRES_PASSWORD` handling: switching from `${POSTGRES_PASSWORD:-goclaw}` to `${POSTGRES_PASSWORD:?...}` will break existing Docker Compose users who never set `POSTGRES_PASSWORD` in `.env` and rely on the default. Their Postgres volumes already use password `goclaw`, so even if `prepare-env.sh` generates a new random password, it won't match the existing database — causing connection failures.

We'll implement an alternative approach that:
- Keeps the `:-goclaw` fallback in compose files (no breaking change)
- Adds **startup warnings** in the gateway when the default password or open CORS is detected
- Applies the MCP bridge 403 fix (same as your approach — this part is great)
- Defers `prepare-env.sh` changes to a separate follow-up

Closing this in favor of a non-breaking implementation. Thanks again for the thorough audit work! 🙏

PR #392, opened 2026-03-23
fix(ui): align stat card heights when sparkline data is missing

huyluseo:
## Summary

Stat cards on the overview dashboard have inconsistent heights — cards with sparkline charts are taller than cards without data.

## Fix

Always reserve the sparkline container (`h-[40px]`) regardless of whether chart data exists. The chart only renders inside when `sparkData.length > 1`, but the space is consistent across all cards.

## Before / After

**Before:** Cards with sparkline data are ~40px taller than cards without, causing jagged row alignment.

**After:** All cards have the same height. Empty sparkline area is invisible but reserves space.

## Files Changed

| File | Change |
|------|--------|
| `ui/web/src/pages/overview/stat-card.tsx` | Move sparkline container div outside the conditional |

### DIFF
--- a/ui/web/src/pages/overview/stat-card.tsx
+++ b/ui/web/src/pages/overview/stat-card.tsx
@@ -47,8 +47,8 @@ export function StatCard({
           {sub && <p className="text-xs text-muted-foreground">{sub}</p>}
         </div>
       </div>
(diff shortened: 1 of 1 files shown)


huyluseo:
Closing — the fix doesn't fully resolve the alignment issue.

PR #393, opened 2026-03-23
fix(telegram): clear draft after materializing DM stream

badgerbees:
## Summary
In Telegram DMs using `draft_transport`, the preview draft sometimes stays in the input area even after the "real" message is delivered, which briefly shows a duplicate reply.

This PR ensures that as soon as the real message is delivered in `Send()`, we explicitly call `sendMessageDraft` with an empty string to clear the stale draft.
## Changes: 
- Capture the `draftID` when a stream ends and clear it in the background after the final message is sent.
- Robustness: The cleanup is asynchronous and "best-effort," so it won't block the actual message delivery if the Telegram API is slow.


badgerbees:
hmm CI/CD error not from me

mrgoonie:
@claude review

Issue #340, opened 2026-03-21
Upload webp format error


chungtran4078:
tool response: unsupported image format: (supported: jpg, png, gif, webp, bmp) although am upload webp
<img width="963" height="730" alt="Image" src="https://github.com/user-attachments/assets/7ad0c0ca-f310-436a-8efc-69f3073e8e4e" />

My images:
[Archive.zip](https://github.com/user-attachments/files/26159019/Archive.zip)

thieung:
<img width="954" height="431" alt="Image" src="https://github.com/user-attachments/assets/11631217-afff-4d77-aaba-754dff65bc78" />
It's fixed

PR #397, opened 2026-03-23
fix: error "unsupported image format: (supported: jpg, png, gif, webp, bmp)"

chungtran4078:
upload image error `unsupported image format` because `path` just an uuid string, does not contain `ext`
image

This fix for https://github.com/nextlevelbuilder/goclaw/issues/340 also


mrgoonie:
@claude review

PR #400, opened 2026-03-23
feat(agent): exclude MCP bridge tools from read-only streak detector

### LINKED ISSUE #399: feat: exclude MCP bridge tools from read-only streak detector
## Problem

The read-only streak detector in `toolloop.go` stops agent runs after 12 consecutive non-mutating tool calls. This guard was designed for filesystem tool loops (`read_file`, `list_files`, `find`) but catches MCP bridge tools in the default fallback of `recordMutation()`.

MCP tools are user-defined external integrations (email, calendar, databases) where read-heavy workflows are expected and legitimate. A workflow like "query inbox + read 10 emails + summarize" hits the critical threshold before the agent can respond.

## Current behavior

`recordMutation()` classifies tools into 3 buckets:
- **Mutating** (`write_file`, `edit`, `spawn`, etc.) → resets streak
- **Neutral** (`exec`, `bash`) → neither resets nor increments
- **Everything else** → increments streak

All `mcp_*` tools fall into "everything else" and increment the read-only streak, triggering the guard at 12 calls.

## Suggested fix

Treat `mcp_*` prefixed tools as neutral (same as `exec`/`bash`) in `recordMutation()` since GoClaw cannot determine whether an MCP tool is read or write.

```go
if toolName == "exec" || toolName == "bash" || strings.HasPrefix(toolName, "mcp_") {
    return // ambiguous — neither reset nor increment
}
```

## Alternatives considered

- Let MCP tool schemas declare mutating/read-only (more correct but higher complexity, could be a follow-up)
- Make thresholds configurable per-agent (addresses symptom not root cause)

## Affected file

`internal/agent/toolloop.go` line 127

kaitranntt:
## Summary

- Treat `mcp_*` prefixed tools as neutral in `recordMutation()` (same as `exec`/`bash`)
- MCP tools are user-defined external integrations where GoClaw cannot determine read vs write
- Prevents false positives on legitimate read-heavy MCP workflows (e.g., Gmail: query inbox + read N emails)

## Changes

- `internal/agent/toolloop.go`: Added `strings.HasPrefix(toolName, "mcp_")` to neutral check
- `internal/agent/toolloop_test.go`: Added `TestReadOnlyStreak_MCPNeutral` and `TestReadOnlyStreak_MCPOnlyNeverTriggers`

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` clean
- [x] `go test ./internal/agent/ -v` — all existing + new tests pass
- [x] Verified MCP tools neither reset nor increment read-only streak
- [x] Verified filesystem tools still trigger streak detection normally

Closes #399

### DIFF
--- a/internal/agent/toolloop.go
+++ b/internal/agent/toolloop.go
@@ -118,14 +118,17 @@ func (s *toolLoopState) detect(toolName string, argsHash string) (level, message
 }
 
 // recordMutation updates the read-only streak based on tool type.
-// Mutating tools reset the streak; exec is neutral (ambiguous); all others increment.
+// Mutating tools reset the streak; exec/bash/mcp are neutral (ambiguous); all others increment.
 func (s *toolLoopState) recordMutation(toolName string) {
 	if mutatingTools[toolName] {
 		s.readOnlyStreak = 0
 		return
 	}
-	if toolName == "exec" || toolName == "bash" {
-		return // ambiguous — neither reset nor increment
+	// exec/bash: ambiguous (could be ls or rm).
+	// mcp_*: user-defined external tools — GoClaw cannot determine read vs write.
+	// Neither reset nor increment the read-only streak.
+	if toolName == "exec" || toolName == "bash" || strings.HasPrefix(toolName, "mcp_") {
+		return
 	}
 	s.readOnlyStreak++
 }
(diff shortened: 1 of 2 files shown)


mrgoonie:
@claude review

PR #403, opened 2026-03-23
fix(zalo): improve inbound identity and resolve image-drop field mismatch

badgerbees:
## Summary
This PR hardens the Zalo OA integration by improving sender identity resolution and ensuring inbound images are not silently dropped due to API field name variations.

## Problem: 
- Inbound Zalo images were being ignored because the handler looked for the documented photo field, while live polling/webhook payloads often contain photo_url.
- Zalo senders were previously "anonymous" to the agent because only the raw technical ID was passed in the `HandleMessage`call.
## Solution:
- Updated `zaloFrom` and `zaloMessage` structs to include display_name, is_bot, photo_url, and message_type.
- Implemented a Compound Sender ID (ID|Name) that prioritizes display_name over legacy username or raw IDs.
- Optimized the image download logic to prefer photo_url for better environment compatibility.

### DIFF
--- a/internal/channels/zalo/zalo.go
+++ b/internal/channels/zalo/zalo.go
@@ -208,18 +208,24 @@ func (c *Channel) handleTextMessage(msg *zaloMessage) {
 		content = "[empty message]"
 	}
 
-	slog.Debug("zalo text message received",
-		"sender_id", senderID,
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

PR #406, opened 2026-03-23
perf(agent): move context maintenance (memory flush & summarization) to background

badgerbees:
## Problem
Maintenance tasks triggered by context thresholds were running synchronously, blocking the return of the agent's response to the user.
## Solution
Moved both `runMemoryFlush` and the summarization LLM call into a background goroutine.


mrgoonie:
@claude review

badgerbees:
@mrgoonie hey, I think your claude code review isnt working here since my PR came from a fork and the branch only exists on that fork

PR #410, opened 2026-03-23
feat(memory): add configurable embedding dimensions for self-hosted providers

### LINKED ISSUE #409: feat: add configurable embedding dimensions for self-hosted providers
## Description

Embedding dimensions are hardcoded to 1536 (OpenAI standard) in both the pgvector schema and provider wiring. Self-hosted providers like Ollama with models that output different native dimensions (e.g. `qwen3-embedding:8b` at 4096) cannot be used for embeddings even though:

1. The `WithDimensions()` plumbing already exists on `OpenAIEmbeddingProvider`
2. Ollama's OpenAI-compatible endpoint supports the `dimensions` parameter for Matryoshka truncation
3. The DB registry fallback path (`createEmbeddingProvider` → registry) creates providers without `WithDimensions()`

## Root Cause

Commit `4db5259` added `WithDimensions(1536)` only for the Gemini case. The registry-based provider path (commit `1b99406`) doesn't call `WithDimensions()` because all initial providers (OpenAI, OpenRouter) natively output 1536.

## Affected Files

- `internal/config/config.go` — `MemoryConfig` struct missing `EmbeddingDimensions` field
- `cmd/gateway_agents.go` — `createEmbeddingProvider()` registry path doesn't apply dimensions

kaitranntt:
## Summary

- Add `embedding_dimensions` config field to `MemoryConfig`, wiring existing `WithDimensions()` to all embedding provider paths (openai, openrouter, gemini, registry fallback)
- Enables self-hosted providers like Ollama whose models output non-1536 native dimensions (e.g. `qwen3-embedding:8b` at 4096) to truncate via Matryoshka to match pgvector `vector(1536)` schema
- Expose field in dashboard UI (global AI Defaults + per-agent memory config) with i18n (en, vi, zh)
- Gemini default behavior preserved (1536 if no override)

Closes #409

## Test plan

- [ ] Verify Gemini embeddings still work without `embedding_dimensions` set (should default to 1536)
- [ ] Configure Ollama provider with `embedding_dimensions: 1536` and verify embeddings are stored correctly
- [ ] Verify UI field appears in both global AI Defaults → Memory and per-agent config → Memory
- [ ] Verify `go build ./...` and `go vet ./...` pass
- [ ] Verify `tsc --noEmit` passes for UI

### DIFF
--- a/cmd/gateway_agents_test.go
+++ b/cmd/gateway_agents_test.go
@@ -1,0 +1,96 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/nextlevelbuilder/goclaw/internal/config"
+	"github.com/nextlevelbuilder/goclaw/internal/store"
+)
+
+// TestBuildEmbeddingProvider_Dimensions tests the dimension resolution chain
+// in buildEmbeddingProvider: EmbeddingSettings → MemoryConfig → provider default.
+func TestBuildEmbeddingProvider_Dimensions(t *testing.T) {
+	tests := []struct {
+		name         string
+		providerType string
+		esDims       int
+		memDims      int
+		wantDims     int // 0 = no WithDimensions call expected
+	}{
+		{
+			name:         "no dims configured, non-gemini → no truncation",
+			providerType: store.ProviderOpenAICompat,
+			wantDims:     0,
+		},
+		{
+			name:         "no dims configured, gemini → default 1536",
+			providerType: store.ProviderGeminiNative,
+			wantDims:     1536,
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

kaitranntt:
Superseded by #563. Upstream already landed the original provider-level embedding-dimensions support after this PR was opened, so the remaining useful work is now captured in the narrower follow-up that guards GoClaw's fixed vector(1536) memory schema and clarifies the provider UX.

PR #412, opened 2026-03-23
fix(feishu): ensure slash commands work in group chats by stripping bot mentions aggressively

badgerbees:
## Issue
In Feishu group chats, users almost always prefix their messages with a bot mention (e.g., @Bot /reset). Previously, our parser was replacing the mention token but leaving behind leading whitespace or a resolved @BotName string in the message body.

Because the system command parser (shared by all channels) looks for a strict leading slash (/), these leftover characters (like a single leading space) would cause commands such as /reset, /model, and /new to fail to trigger. Instead, the command would be forwarded to the LLM agent as regular text, creating a confusing user experience.

## Changes
I have hardened the mention normalizer in `bot_parse.go` to ensure slash commands are always recognized in both p2p and group contexts:
1. Comprehensive Id Stripping: The resolver now identifies and strips both the internal Feishu mention token (@_user_N) and the human-readable @BotName. This makes the "Slash Command Probe" robust against different message types ("Text" vs. "Post" messages).
2. Aggressive Leading-Space Trim: Added an explicit `strings.TrimSpace` to the normalization result. This guarantees that if a user sends @Bot /new, the resulting text starts exactly at the / character, allowing the command interceptor to catch it.
3. Consistent Cross-Context Logic: The logic now handles all conversation types correctly, ensuring that bot mentions never "leak" into the command probing path.

### DIFF
--- a/internal/channels/feishu/bot_parse.go
+++ b/internal/channels/feishu/bot_parse.go
@@ -234,19 +234,27 @@ func extractPostImageKeys(rawContent string) []string {
 }
 
 // resolveMentions replaces mention placeholders (@_user_1, @_user_2, etc.) in content.
-// Bot mention is stripped entirely; other user mentions become @Name.
+// Bot mention is stripped entirely (both token and @Name) to ensure slash commands work
+// in group chats. Others are resolved to @Name.
 func resolveMentions(text string, mentions []mentionInfo, botOpenID string) string {
 	for _, m := range mentions {
 		if m.Key == "" {
 			continue
 		}
-		if botOpenID != "" && m.OpenID == botOpenID {
-			// Strip bot mention
+		isBot := botOpenID != "" && m.OpenID == botOpenID
+
+		if isBot {
+			// Aggressively strip bot mention in all forms (token and resolved name)
+			// to ensure slash commands like "/reset" have no leading markers.
 			text = strings.ReplaceAll(text, m.Key, "")
+			if m.Name != "" {
+				text = strings.ReplaceAll(text, "@"+m.Name, "")
+			}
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

clark-cant (commented):
**Summary**: Hardens Feishu mention normalization so slash commands (`/reset`, `/model`, `/new`) work in group chats — the bot mention is now stripped in both token (`@_user_N`) and resolved-name (`@BotName`) form, with aggressive leading-space trimming so the command parser sees a clean leading `/`.

**Risk level**: Low — 1 file, +12/−4, internal channel parsing only, no API/DB/config changes.

**Mandatory gates**:
- Duplicate / prior implementation: clear — no overlapping open or merged PR found; the related feishu mention-detection work (#1189, merged) is complementary, not a duplicate.
- Project standards: no repo standards doc exists for this package; change matches existing file conventions and keeps the existing `TrimSpace` behavior.
- Strategic necessity: clear value — fixes a real user-facing bug (slash commands silently forwarded to the LLM as plain text in Feishu group chats).

**Findings**:
- Suggestion: `strings.ReplaceAll(text, "@"+m.Name, "")` strips the bot's `@Name` everywhere in the text, not only at mention boundaries. If the bot name is a prefix of another mention or appears quoted mid-message (e.g. bot named "Bot" and text contains "ping @Botty"), this can over-strip. Consider replacing only a leading mention occurrence (or stripping before `TrimSpace` only when it immediately follows the token position). Non-blocking.

**Verdict**: Comment — safe to merge as-is.

*Posted by /ck:review-pr at 2026-08-17T11:17:17Z*

PR #414, opened 2026-03-23
perf(agent): improve accuracy of context pruning by including Thinking and ToolCall overhead

badgerbees:
## The Issue
The agent's message pruner previously only counted the `Content` field. For models that provide extensive reasoning in a separate `Thinking` field (e.g., DeepSeek, Claude), or for agents that make frequent tool calls, this led to a significant Under-Estimation of the total context usage. In some cases, a 4000-token thought block was being treated as "0 characters," which delayed pruning until the conversation was already far beyond the provider's token limit, causing the session to crash.

## The Changes
This PR upgrades the `estimateMessageChars` function to be more comprehensive:
1. Thinking-Aware: It now includes the character count of the `Thinking` field (reasoning section) in the overall message size.
2. Tool-Aware: It adds the character count of any `ToolCall` names plus a fixed buffer of 100 characters per call to approximate the overhead of JSON arguments.
3. Improved Pruning Reliability: By accurately calculating the total "token weight" of each message, the `pruneContextMessages` logic will now correctly trigger "Soft Trimming" and "Hard Clearing" steps at the expected thresholds.els and tool-heavy agents

### DIFF
--- a/internal/agent/pruning.go
+++ b/internal/agent/pruning.go
@@ -244,9 +244,19 @@ func findAssistantCutoff(msgs []providers.Message, keepLast int) int {
 	return -1
 }
 
-// estimateMessageChars returns the character count of a message's content.
+// estimateMessageChars returns the character count of a message's content sections.
+// Includes Thinking content and ToolCalls to ensure pruning is accurate for
+// reasoning models and tool-heavy agents.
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

PR #415, opened 2026-03-23
feat(agent): support thinking tag promotion from assistant response text

badgerbees:
## The Issue
For models that output reasoning within inline tags (e.g., DeepSeek models or older Claude-3 versions), goclaw was previously identifying these tags and deleting them entirely during sanitization. This led to "invisible reasoning," making it impossible to see the agent's internal logic after the run finished. 

Furthermore, if the model returned malformed or unclosed reasoning tags via a proxy, the simple string replacement logic could lead to artifacts leaking into the final user message.

## The Changes
I have added a "Thinking Promotion" step to the `SanitizeAssistantContent` pipeline:
1. Tag Extraction: Added `PromoteThinkingTags` (in `internal/agent/sanitize.go`) to scan the assistant message body for <think>, <thinking>, <thought>, and <antThinking> tags.
2. Metadata Promotion: The content inside these tags is extracted, trimmed, and appended to the `Thinking` field of the `Message` struct. This ensures that the reasoning is saved to history (and DB) even if it was originally provided as raw text chunks.
3. Loop Integration: Updated `internal/agent/loop.go` to perform this promotion right before final sanitization. This ensures that the reasoning text is moved out of the main body before the body is cleaned for user delivery.
4. Stability Guard: By using `FindAllStringSubmatch`, we treat malformed or nil thinking attempts safely they are either captured as a group or ignored based on valid XML structure, satisfying the stability requirements for malformed assistant content.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -1348,7 +1348,8 @@ func (l *Loop) runLoop(ctx context.Context, req RunRequest) (*RunResult, error) 
 
 	}
 
-	// 4. Full sanitization pipeline (matching TS extractAssistantText + sanitizeUserFacingText)
+	// 4. Thinking promotion + Sanitization pipeline (matching TS extractAssistantText + sanitizeUserFacingText)
+	finalContent, finalThinking = PromoteThinkingTags(finalContent, finalThinking)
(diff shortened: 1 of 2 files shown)


mrgoonie:
@claude review

PR #417, opened 2026-03-23
perf(agent): optimize prefix cache efficiency and harden inbound security

badgerbees:
## Issue
The current system prompt structure causes frequent KV cache busts because dynamic context (such as subagent instructions and heartbeat checklists) is injected before the large, stable project context files. This results in significantly higher latency and token costs on providers that support prefix caching (like Anthropic and OpenAI). Additionally, there is a security risk where user-supplied messages can spoof system markers in the prompt.

## Changes & Solution
1. Prefix Cache Alignment: Reordered the system prompt in `internal/agent/systemprompt.go`to place the stable # Project Context sections before the dynamic extraSystemPrompt. This ensures the high-context prefix stays cached across turns.
2. Timeline-Based Events: Moved gateway system events (model switches, node status) from the system prompt into the conversation timeline. They are now prepended to the user's message with a System: [HH:MM:SS] marker in `internal/agent/loop_history.go`.
3. Inbound Sanitization: Implemented a markdown-aware `SanitizeInboundSystemTags` utility in `internal/agent/sanitize.go`. It rewrites user-supplied System: prefixes at the start of any line to System (untrusted):, but intelligently skips any text inside triple-backtick (```) code blocks to avoid false positives.

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -268,7 +268,7 @@ func (l *Loop) runLoop(ctx context.Context, req RunRequest) (*RunResult, error) 
 
 	// buildMessages resolves context files once and also detects BOOTSTRAP.md presence
 	// (hadBootstrap) — no extra DB roundtrip needed for bootstrap detection.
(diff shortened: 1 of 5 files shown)


mrgoonie:
@claude review

PR #420, opened 2026-03-23
fix: add panic recovery to tool, cron, and summarization goroutines

vanducng:
## Summary

- A panicking tool, cron job, or auto-summarization goroutine crashes the entire server — disconnecting all clients and aborting all running agents
- Added `defer/recover` at 4 critical goroutine sites with stack trace logging
- Tool panics return error results to the LLM so agents can recover gracefully
- Only 2 `recover()` calls existed in the entire codebase (both in WS read/write pumps)

## Changes

| File | What |
|------|------|
| `internal/tools/registry.go` | `safeExecute()` wrapper around `tool.Execute()` — catches all tool panics |
| `internal/agent/loop.go` | Recovery in parallel tool execution goroutine (defense-in-depth) |
| `internal/agent/loop_history.go` | Recovery in background auto-summarization goroutine |
| `internal/cron/service_execution.go` | Recovery in cron job execution goroutine |
| `internal/tools/registry_test.go` | 3 new tests: basic panic, index-OOB panic, concurrent panic recovery |

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test -race ./internal/tools/ ./internal/agent/ ./internal/cron/` — all pass
- [x] New tests verify: panicking tool returns error result, not crash
- [x] Concurrent panic recovery test (20 goroutines) passes with race detector

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -7,7 +7,6 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
-	"runtime"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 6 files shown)


mrgoonie:
@claude review

PR #425, opened 2026-03-24
feat(http): enhance secure CLI environment variable handling

ba0f3:
Improves secure CLI credential management so admins can define, update, and remove injected environment variables from the web UI — without exposing secret values to the client.

**Backend:**
Expose only env_keys (sorted names) on list/get; never return env values.
Support partial updates: merge non-empty `env`; `env_remove` deletes keys (applied before merge so a key can be removed and set in one request).

**Frontend:**
Manual/custom credentials: key/value editor when not using a preset.
Edit: pre-fill stored keys with empty values; distinct placeholders for existing vs new vars; row delete → `env_remove`.

### DIFF
--- a/internal/http/secure_cli.go
+++ b/internal/http/secure_cli.go
@@ -5,6 +5,7 @@ import (
 	"log/slog"
 	"net/http"
 	"regexp"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


mrgoonie:
@claude review

PR #436, opened 2026-03-24
fix(slack): enable thread-based session isolation for DMs (AI Panel support)

### LINKED ISSUE #435: [Bug] Slack DM threads are ignored, breaking AI Panel session isolation
## Bug Description

GoClaw's Slack integration currently ignores Slack threads when the conversation is a Direct Message (DM). This causes all interactions in Slack's native **AI Panel** (which uses threads behind the scenes) to be merged into a single, continuous GoClaw session. 

When a user clicks \"New Chat\" in the Slack AI Panel, it starts a fresh thread, but GoClaw continues using the same session ID because it only looks at the permanent Channel ID for DMs.

## Current Behavior
The Slack handler in `internal/channels/slack/handlers.go` has an explicit check that disables thread-based `localKey` generation for DMs:

```go
if !isDM && threadTS != \"\" {
    localKey = fmt.Sprintf(\"%s:thread:%s\", channelID, threadTS)
}
```

## Expected Behavior
Slack DMs should support thread-based session isolation. Each new thread in a DM (like those created by the AI Panel or manual threads) should map to a unique GoClaw session.

## Impact
- Slack AI Panel \"New Chat\" does not work (context is never cleared).
- Manual threads in DMs share context with the main channel, causing confusion and \"leaking\" memory between topics.

## Suggested Fix
1. Remove the `!isDM` restriction in the Slack handler.
2. Update the session key builder to support string-based thread IDs (Slack uses timestamps).
3. Update the global message consumer to respect thread overrides in `local_key`.

I have a fix ready and will submit a PR shortly.

Erudition:
Fixes #435

This PR enables thread-based session isolation for Slack Direct Messages. 

### Changes:
- **`internal/channels/slack/handlers.go`**: Removed the `!isDM` restriction, allowing `thread_ts` to be propagated for DMs. This is essential for supporting Slack's native **AI Panel**, which uses threads to isolate conversations.
- **`internal/sessions/key.go`**: Added `BuildScopedThreadSessionKey` to support string-based thread IDs (Slack uses timestamps like `1710000000.000000`).
- **`cmd/gateway_consumer_normal.go`**: Updated the message consumer to detect thread-scoped `local_key` values and automatically override the session key to ensure isolation.

Verified with Slack AI Panel: clicking \"New Chat\" now correctly starts a fresh GoClaw session with clean history.

### DIFF
--- a/cmd/gateway_consumer_normal.go
+++ b/cmd/gateway_consumer_normal.go
@@ -57,6 +57,14 @@ func processNormalMessage(
 	}
 	sessionKey := sessions.BuildScopedSessionKey(agentID, msg.Channel, sessions.PeerKind(peerKind), msg.ChatID, cfg.Sessions.Scope, cfg.Sessions.DmScope, cfg.Sessions.MainKey)
 
+	// Thread-based isolation override (e.g. Slack threads or AI Panel)
+	if lk := msg.Metadata["local_key"]; lk != "" && strings.Contains(lk, ":thread:") {
+		parts := strings.Split(lk, ":thread:")
+		if len(parts) == 2 {
+			sessionKey = sessions.BuildScopedThreadSessionKey(agentID, msg.Channel, sessions.PeerKind(peerKind), msg.ChatID, parts[1])
+		}
+	}
+
 	// Forum topic: override session key to isolate per-topic history.
 	// TS ref: buildTelegramGroupPeerId() in src/telegram/bot/helpers.ts
 	if msg.Metadata["is_forum"] == "true" && peerKind == string(sessions.PeerGroup) {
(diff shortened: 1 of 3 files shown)


mrgoonie:
@claude review

viettranx:
Thanks for the contribution @Erudition! Great root cause analysis and clean fix.

I've incorporated your changes into #449 with two minor review adjustments:
- `strings.Split` → `strings.SplitN` for more robust thread key parsing (prevents silent failure if threadID ever contains `:thread:`)
- Removed `IsThreadSession` (no callers yet — can be re-added when needed)

You're credited as co-author on the commit. Closing this in favor of #449.

PR #440, opened 2026-03-24
fix(ui): show toast error when package install/uninstall fails

thieung:
## Summary
- `HttpClient` throws `ApiError` on HTTP 500 but `installPackage`/`uninstallPackage` hooks didn't catch it — error bypassed toast logic, so users got no feedback on failure
- Wrapped `http.post()` in try/catch to always show toast with server error message
- Updated `PackageSection` handlers to check `res.ok` instead of try/catch flow control

## Test plan
- [x] Install a non-existent system package (e.g. `sh`) — verify toast error appears with server message
- [x] Install a valid package — verify success toast still works
- [x] Uninstall a non-existent package — verify error toast appears
- [x] Verify UI status indicator shows "error" state (red) on failure, not "success"

## Screenshot

Install valid package: Success toast
image

Install invalid package: Error toast
image

### DIFF
--- a/cmd/pkg-helper/main.go
+++ b/cmd/pkg-helper/main.go
@@ -157,10 +157,6 @@ func doInstall(pkg string) response {
 func doUninstall(pkg string) response {
 	slog.Info("pkg-helper: uninstalling", "package", pkg)
 
-	// Read other persisted packages before removing, so we can re-add any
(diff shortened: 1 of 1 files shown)


mrgoonie:
@claude review

PR #446, opened 2026-03-24
fix(tools): session ownership check uses UUID instead of agent_key

therichardngai-code:
## Summary

`resolveAgentIDString()` returns the agent **UUID** from context, but session keys use the agent **key** (string): `agent:<key>:ws:direct:<id>`. The `HasPrefix` ownership check in `session_status`, `session_history`, and `session_send` always fails because UUID != key, returning:

> access denied: session belongs to a different agent

### Why it's masked in single-tenant

In single-user/gateway-token setups, the agent UUID in context is often `Nil` (no agent resolution step), so `resolveAgentIDString` returns `""` and the check is skipped (`agentID != "" &&`). The bug surfaces when agents have both a UUID and a key — which is all agents created via the API or WebUI in multi-tenant deployments.

### Fix

- Add `AgentKeyKey` to context (`store.WithAgentKey` / `AgentKeyFromContext`)
- Inject agent key (`l.id`) into context in `runLoop` alongside the UUID
- `resolveAgentIDString` prefers agent key, falls back to UUID for backward compat

### Files changed

- `internal/store/context.go` — new `AgentKeyKey` context key + accessors
- `internal/agent/loop.go` — inject `l.id` via `WithAgentKey`
- `internal/tools/sessions_send.go` — `resolveAgentIDString` prefers key over UUID

## Test plan

- [ ] Agent calls `session_status` on its own session — should succeed (was "access denied")
- [ ] Agent calls `session_history` — should succeed
- [ ] Agent calls `session_send` to its own session — should succeed
- [ ] Cross-agent access still denied (different key in session prefix)

### DIFF
--- a/internal/agent/loop.go
+++ b/internal/agent/loop.go
@@ -39,9 +39,12 @@ func (l *Loop) runLoop(ctx context.Context, req RunRequest) (*RunResult, error) 
 		l.emit(event)
 	}
 
-	// Inject agent UUID into context for tool routing
+	// Inject agent UUID + key into context for tool routing
 	if l.agentUUID != uuid.Nil {
 		ctx = store.WithAgentID(ctx, l.agentUUID)
+	}
+	if l.id != "" {
+		ctx = store.WithAgentKey(ctx, l.id)
 	}
 	// Inject tenant into context for tool-level tenant scoping (spawn, MCP, etc.)
 	if l.tenantID != uuid.Nil {
--- a/internal/store/context.go
+++ b/internal/store/context.go
@@ -12,7 +12,8 @@ const (
 	// UserIDKey is the context key for the external user ID (TEXT, free-form).
 	UserIDKey contextKey = "goclaw_user_id"
 	// AgentIDKey is the context key for the agent UUID.
-	AgentIDKey contextKey = "goclaw_agent_id"
+	AgentIDKey  contextKey = "goclaw_agent_id"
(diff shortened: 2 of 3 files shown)


viettranx:
Thanks for identifying this bug — the problem analysis is spot-on. Session keys use `agent:<key>:<rest>` but `resolveAgentIDString()` returns the UUID, so the `HasPrefix` check always fails.

However, the fix has an issue: **`resolveAgentIDString` is shared by the cron tool**, which needs the UUID (not the key) for DB operations:

1. **`ListJobs`** (`cron_crud.go:112`) does `uuid.Parse(agentID)` — passing the agent key (e.g. `"default"`) will fail the parse, silently skipping the `agent_id` filter → **all cron jobs visible to all agents** (security regression)
2. **`checkJobOwnership`** (`cron.go:317`) compares `job.AgentID` (UUID from DB) with the agent key → never matches → **all cron mutations permanently denied**

I've pushed an alternative fix on `fix/panic-recovery-goroutines` (commit 62be52e0) that splits the resolver into two functions:
- `resolveAgentKey()` — returns agent string key for session prefix matching
- `resolveAgentUUIDString()` — returns agent UUID string for DB operations (cron)

This also updates all 5 call sites (your PR only covered `sessions_send.go` but `sessions.go` and `sessions_history.go` have the same bug).

viettranx:
Closing — the fix has a cron tool regression (see comment above). The corrected version is on `fix/panic-recovery-goroutines` (62be52e0).

PR #474, opened 2026-03-25
fix(storage): hide tenant roots from master storage

gbut277:
## Summary

This prevents the master Storage page from exposing the cross-tenant isolation root.

Before:
- The master storage listing surfaced the top-level `tenants/` directory and its contents.
- Master storage reads could target `tenants/...` paths directly.
- Master storage size calculations included files stored under tenant-specific directories.

After:
- The master storage API hides the `tenants/` root from listings.
- Direct reads into `tenants/...` return `404` from the storage API.
- Master storage size calculations exclude tenant-specific files.

## Changes

- added a storage-level hidden-path rule for the master tenant scope
- filtered hidden tenant paths out of storage list and size traversal
- rejected master storage reads into hidden tenant paths
- added regression tests for listing, reading, and size calculation against tenant roots

## Test plan

- `go test ./internal/http`
- Verified master storage listing does not return `tenants/`
- Verified master storage reads for `tenants/...` return `404`

### DIFF
--- a/internal/http/storage.go
+++ b/internal/http/storage.go
@@ -71,7 +71,7 @@ func (h *StorageHandler) tenantBaseDir(r *http.Request) string {
 // tenants (tenant isolation root — each tenant's data is scoped internally).
 var protectedDirs = []string{"skills", "skills-store", "media", "tenants"}
 
-func isProtectedPath(rel string) bool {
+func topLevelPath(rel string) string {
 	top := rel
 	if before, _, ok := strings.Cut(rel, "/"); ok {
 		top = before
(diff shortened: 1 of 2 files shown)


viettranx:
Closing in favor of a direct fix on `dev`. The fix has been applied with the following improvements over this PR:

**Bug fix:** `handleSize` walk callback now checks `isHiddenPath` *before* the `d.IsDir()` early return, allowing `filepath.SkipDir` to skip the entire `tenants/` subtree. The original PR placed the check after `d.IsDir()`, which meant the walk would descend into all tenant directories and skip files individually — a performance issue for deployments with many tenants.

**Dead code cleanup:** Removed the redundant `strings.IndexByte(top, '/')` check in `topLevelPath()` — after `strings.Cut(rel, "/")`, the result cannot contain another `/`.

**Additional test coverage:** Added `TestIsHiddenPathOnlyAffectsMaster` (non-master tenants unaffected) and `TestStorageListSubpathTenantReturnsNotFound` (explicit `?path=tenants` query).

Thanks for identifying the security gap — the core approach (isHiddenPath scoped to master tenant + case-insensitive match) was solid.

PR #526, opened 2026-03-28
fix(security): decrypt downgrade warnings, pairing expiry race, sender-ID validation

kabuto-png:
## Summary

Lower-priority security hardening — defense-in-depth improvements.

### Crypto: Decrypt downgrade detection
- Log `slog.Warn` when `Decrypt()` encounters values without `aes-gcm:` prefix (potential downgrade attack or legacy plaintext)
- Also warns on invalid base64 and too-short encrypted values
- Backward compatibility preserved — still returns plaintext, but now observable in logs

### Pairing: Expiry race condition fix
- Add `AND expires_at > NOW()` to `ApprovePairing` SELECT query
- Closes race where a pairing code could be approved after TTL expiry (between prune DELETE and lookup SELECT)

### Pairing: Sender-ID format validation
- Validate sender-ID format in `handleRequest` and `handleBrowserPairingStatus`
- Regex: `^[a-zA-Z0-9][a-zA-Z0-9._:@-]*$`, max 128 chars
- Prevents log injection and bus event poisoning via crafted sender IDs

## Test plan

- [ ] `go build ./...` — passes
- [ ] `go vet ./...` — passes
- [ ] Decrypt a value without `aes-gcm:` prefix — should log warning and return plaintext
- [ ] Approve an expired pairing code — should fail with "not found or expired"
- [ ] Request pairing with `senderId: "admin'; DROP TABLE--"` — should return 400
- [ ] Request pairing with `senderId: "group:12345"` — should succeed

### DIFF
--- a/internal/crypto/aes.go
+++ b/internal/crypto/aes.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/hex"
 	"errors"
+	"log/slog"
 	"strings"
 )
 
@@ -55,6 +56,9 @@ func Decrypt(ciphertext, key string) (string, error) {
 	}
 
 	if !IsEncrypted(ciphertext) {
+		slog.Warn("crypto.unencrypted_value_read",
+			"hint", "value lacks aes-gcm: prefix — may be legacy plaintext or tampered",
+		)
 		return ciphertext, nil
 	}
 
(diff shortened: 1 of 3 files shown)


viettranx:
Closing — all changes from this PR (decrypt downgrade warnings, pairing expiry race fix, sender-ID validation) have been incorporated directly into the dev branch with additional improvements:

- SQLite store parity: expiry race fix also applied to `sqlitestore/pairing.go`
- Sender-ID validation extended to `handleRevoke` (was missing in this PR)
- Slack compound senderID cleanup: removed legacy `senderID|displayName` format, now uses plain senderID like all other channels

These will reach main via the next dev → main merge.

PR #535, opened 2026-03-28
fix(telegram): duplicate contact from inconsistent senderID

Luvu182:
## Summary
- Non-mention group message path passed raw `userID` as sender_id to `EnsureContact`
- Processed message path passed `senderID` (which includes `|username` suffix)
- `UNIQUE(channel_type, sender_id)` constraint created two rows for the same person
- Fix: use `senderID` consistently in both paths

## Test plan
- [ ] Send message in Telegram group without @mentioning bot
- [ ] Then @mention bot in same group
- [ ] Verify only one contact row exists for the user


Luvu182:
Merged into #537 (channel_groups PR already includes this fix)

PR #558, opened 2026-03-29
fix: include tenant ID header in skill upload requests

### LINKED ISSUE #557: bug: Custom skill upload missing X-GoClaw-Tenant-Id header
**Context**

 When a user uploads a custom skill from the web UI while scoped to a non-master tenant, the skill is created on the Master tenant instead. The user then can't see it under their tenant's Custom tab.

 **Root Cause**

 The upload() API call in ui/web/src/api/http-client.ts (line 71) is missing the X-GoClaw-Tenant-Id header. It builds headers manually (only Authorization + X-GoClaw-User-Id) instead of using
 this.authHeaders() which includes X-GoClaw-Tenant-Id and X-GoClaw-Sender-Id.

 Without X-GoClaw-Tenant-Id in the request, the backend auth (internal/http/auth.go:192-194) defaults to MasterTenantID for owner users, so the skill always lands on Master regardless of which tenant
 the user is currently scoped to.

 **Fix**

 File: ui/web/src/api/http-client.ts (line 71-82)

 Replace the manual header construction in upload() with this.authHeaders():

 async upload<T>(path: string, formData: FormData): Promise<T> {
     const res = await fetch(this.buildUrl(path), {
       method: "POST",
       headers: this.authHeaders(),  // reuse shared auth headers (includes tenant scope)
       body: formData,
     });
     // ... rest unchanged

 Note: authHeaders() intentionally omits Content-Type (unlike headers()), which is correct for FormData — the browser sets multipart/form-data boundary automatically.

anhle128:
## Summary
- Fix custom skill uploads always landing on Master tenant instead of the user's current tenant
- Replace manual header construction in `upload()` with `authHeaders()` which includes `X-GoClaw-Tenant-Id`
- Removes 6 lines of duplicated header logic

Closes #557

## Test plan
- [ ] Switch to a non-master tenant in the web UI
- [ ] Upload a custom skill via the Custom tab
- [ ] Verify the skill appears under the current tenant, not Master

### DIFF
--- a/ui/web/src/api/http-client.ts
+++ b/ui/web/src/api/http-client.ts
@@ -69,15 +69,9 @@ export class HttpClient {
   }
 
   async upload<T>(path: string, formData: FormData): Promise<T> {
-    const headers: Record<string, string> = {};
-    const token = this.getToken();
-    if (token) headers["Authorization"] = `Bearer ${token}`;
-    const userId = this.getUserId();
-    if (userId) headers["X-GoClaw-User-Id"] = userId;
-
     const res = await fetch(this.buildUrl(path), {
       method: "POST",
-      headers,
+      headers: this.authHeaders(),
       body: formData,
     });
 


thieung:
LGTM

PR #570, opened 2026-03-30
fix(summoner): preserve user-set agent name after summon completes

teexiii:
## Problem

  When creating a predefined agent with a custom name, the name was
  overwritten in two places once summoning completed:

  1. `display_name` in the DB was replaced by the LLM-extracted name from
     IDENTITY.md (e.g. "Tiểu La"), ignoring the user's input.
  2. IDENTITY.md itself still contained the LLM-generated name, so the
     agent introduced itself with that name regardless of what was stored
     in `display_name`.

  ## Root cause

  `finishSummon` unconditionally applied `extractIdentityName(identityContent)`
  to the `display_name` update map with no check for an existing value.
  Even if the user had set a name at creation time, both the DB field and
  the IDENTITY.md persona file ended up with whatever name the LLM chose.

  ## Fix

  - Before writing the LLM-derived name, load the current agent record.
    Only update `display_name` if it is empty (user did not provide one).
  - If the user *did* set a name that differs from the LLM-generated one,
    rewrite the `Name:` line in IDENTITY.md via `replaceIdentityName()` so
    the agent's persona file matches the user's intended name.
  - Regeneration (`ResummonAgent`) is left unchanged — when explicitly
    rebuilding a persona, updating the name is expected.

  ## Result

  An agent created as "La Alilala" will have both `display_name = "La Alilala"`
  and `Name: La Alilala` in IDENTITY.md after summoning completes, and will
  introduce itself with the correct name in chat.


viettranx:
UX concern đã được fix trong commit ecf6463d:

**Trước:** Summon/regenerate/resummon ghi đè `display_name` vô điều kiện từ LLM → user mất tên custom đã đặt.

**Sau:** Nếu user đã đặt tên custom → giữ nguyên `display_name` + sync IDENTITY.md `Name:` cho khớp. Agent tự giới thiệu đúng tên user chọn, UI cũng hiện đúng. Nếu chưa có tên (tạo qua API) → dùng tên LLM như cũ.

PR #573, opened 2026-03-30
fix: use agent_key instead of UUID for session tool authorization

jooray:
## Summary

Session tools (`sessions_list`, `session_status`, `sessions_history`, `sessions_send`) were using `resolveAgentIDString(ctx)` which returns the agent **UUID**, but session keys are constructed using the agent **key** (a human-readable identifier like `"default"` or `"my-agent"`). This caused all session tool operations to fail silently or return "access denied" for every channel (SimpleX, WhatsApp, Discord, Zalo, HTTP).

## Root cause

- `resolveAgentIDString(ctx)` reads `store.AgentIDFromContext(ctx)` → returns UUID (e.g., `"550e8400-..."`)
- Session keys are built by `BuildSessionKey()` / `BuildGroupSessionKey()` using agent_key → `"agent:my-agent:ws:direct:abc123"`
- `sessions.List(ctx, agentUUID)` does `LIKE 'agent:<UUID>:%'` → matches nothing
- Cross-agent check `strings.HasPrefix(key, "agent:"+agentUUID+":")` → always fails for legitimate sessions

## Fix

Replace `resolveAgentIDString(ctx)` with `ToolAgentKeyFromCtx(ctx)` in all four session tools. Add fail-closed guard for empty agent key.

`resolveAgentIDString` is **not removed** — it is still correctly used by `cron.go` which genuinely needs the UUID for DB queries.

## Affected tools

| Tool | File | Impact |
|------|------|--------|
| `sessions_list` | `sessions.go` | Always returned 0 sessions |
| `session_status` | `sessions.go` | Always returned "access denied" |
| `sessions_history` | `sessions_history.go` | Always returned "access denied" |
| `sessions_send` | `sessions_send.go` | Always returned "access denied" + label resolution broken |

## Affects all channels

No channel sets tool context directly. All channels go through `BaseChannel.HandleMessage()` → bus → consumer → agent loop, where `loop_context.go:211` uniformly calls `WithToolAgentKey(ctx, l.id)`. The bug was in the tools reading the wrong context value.

## Test changes

- **`sessions_test.go`**: Updated `agentCtx()` helper to also call `WithToolAgentKey()`, mirroring what the agent loop does in production. Without this, all happy-path tests hit the new fail-closed `"agent context required"` guard.
- **`context_file_interceptor_test.go`**: Added missing `MigrateUserDataOnMerge` stub to `stubAgentStore`. This is a pre-existing issue (the mock fell behind the `AgentStore` interface) that prevented the entire `tools` test package from compiling.

All 18 session tests pass. Full `./internal/tools/` test suite passes.

### DIFF
--- a/internal/tools/sessions_test.go
+++ b/internal/tools/sessions_test.go
@@ -170,7 +170,10 @@ func agentCtx(agentID string) context.Context {
 	if err != nil {
 		return ctx
 	}
-	return store.WithAgentID(ctx, uid)
+	ctx = store.WithAgentID(ctx, uid)
+	// Session keys use agent_key, not UUID.  In production WithToolAgentKey
+	// is set by the agent loop; tests must mirror this.
+	return WithToolAgentKey(ctx, agentID)
 }
 
 func agentCtxWithSandbox(agentID, sandboxKey string) context.Context {


viettranx:
Already merged via commit 33f975ed

PR #585, opened 2026-03-30
fix(ui): restore provider-owned Codex pool inherit state

### LINKED ISSUE #584: bug(ui): fix provider-owned Codex pool state drift
## Description
Provider-owned ChatGPT OAuth pool membership and agent routing state drift apart in the dashboard.

Users can get stuck in a custom primary-only agent override that does not inherit later provider pool members, while the UI makes that state look like the default. Provider pages also hide saved members when those aliases are not ready, which makes ownership conflicts feel arbitrary.

## Root Cause
- Agent pages derive `custom` mode when no provider default exists, so users cannot cleanly return to an inherit/default state later.
- Saving a no-op custom primary-only override persists an explicit opt-out from provider defaults.
- Provider pool membership UI only surfaces ready aliases, so saved blocked/signed-out members disappear from the editor even though backend ownership rules still enforce them.

## Affected Files
- ui/web/src/pages/agents/agent-detail/agent-codex-pool-page.tsx
- ui/web/src/pages/agents/agent-detail/agent-advanced-dialog.tsx
- ui/web/src/pages/agents/agent-detail/codex-pool-routing-draft-utils.ts
- ui/web/src/pages/agents/agent-detail/agent-display-utils.ts
- ui/web/src/pages/agents/agent-detail/config-sections/chatgpt-oauth-routing-section.tsx
- ui/web/src/pages/agents/agent-detail/config-sections/chatgpt-oauth-pool-sections.tsx
- ui/web/src/pages/providers/provider-detail/provider-overview.tsx
- internal/http/chatgpt_oauth_pool_validation.go

kaitranntt:
## Summary
This PR keeps the backend provider-owned Codex pool ownership model unchanged and fixes the UI/state drift that made the feature look inconsistent from both the provider page and the agent page.

## Visual Review
Rendered HTML report: [open before/after report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaitranntt/goclaw/3662bdccda20c295e119ec067bf83bba4dfce37d/.github/pr-assets/585/index.html)

### Provider overview: saved disabled member

| Before | After |
| --- | --- |
| Before provider overview | After provider overview |

Before: the saved extra member could disappear entirely when it was disabled, signed out, or otherwise not ready.

After: the saved member stays visible and its blocked/non-ready state is shown explicitly.

### Agent Codex pool: no explicit override, no provider-owned pool yet

| Before | After |
| --- | --- |
| Before agent Codex pool | After agent Codex pool |

Before: an untouched agent could render as `Custom For This Agent`, which looked like a saved primary-only override.

After: the same untouched agent correctly stays on `Use Provider Defaults`, with copy that explains it will remain on the primary account until provider-owned members are added later.

## Root cause
The problem was not router failover logic. It was frontend state/render drift in two places:

- The agent draft builder materialized implicit state as `custom + primary_first + []` when the provider-owned pool had no saved members yet.
- The provider overview rendered pool membership too heavily from ready/quota-fetched accounts, so saved non-ready members vanished from the page.

## What changed
- Preserve real `inherit` draft state for untouched agents instead of manufacturing a synthetic custom override.
- Allow `Use Provider Defaults` even when the provider has not saved any extra pool members yet.
- Keep selected saved members visible even when they are disabled, signed out, or otherwise not ready.
- Render provider pool membership from saved provider settings, not only from ready quota lookups.
- Add regression coverage for inherit-without-provider-pool behavior.

## Why this approach
I intentionally did not change the backend ownership model.

- Provider still owns membership.
- Agent still chooses inherit vs custom routing behavior.
- `Primary First` remains a routing strategy, not "primary only forever".

This PR fixes the UI so it reflects that contract correctly instead of weakening the ownership boundary.

### Passed
- `go test ./internal/store ./internal/http ./internal/providers ./internal/providerresolve`
- `pnpm --dir ui/web build`
- `git diff --check`

### Scenario validation
Validated locally with seeded before/after UI runs for the two reported cases above.

## Known unrelated issues
- `go test ./...` still fails in `pkg/browser` at `TestResolveRemoteCDP_DefaultPort` with `/json/version returned HTTP 404`; this appears unrelated to this patch.
- `pnpm --dir ui/web lint` fails because `eslint` is not installed in `ui/web` in the current environment.

Closes #584

### DIFF
--- a/internal/store/agent_store_test.go
+++ b/internal/store/agent_store_test.go
@@ -124,6 +124,27 @@ func TestResolveEffectiveChatGPTOAuthRoutingAllowsInheritWithoutSavedProviderPoo
 	}
 }
 
+func TestResolveEffectiveChatGPTOAuthRoutingInheritForwardsProviderDefaults(t *testing.T) {
+	defaults := &ChatGPTOAuthRoutingConfig{
+		Strategy:           ChatGPTOAuthStrategyRoundRobin,
+		ExtraProviderNames: []string{"codex-work"},
+	}
+	override := &ChatGPTOAuthRoutingConfig{
+		OverrideMode: ChatGPTOAuthOverrideInherit,
+	}
+
+	got := ResolveEffectiveChatGPTOAuthRouting(defaults, override)
+	if got == nil {
+		t.Fatal("ResolveEffectiveChatGPTOAuthRouting() = nil, want config forwarding provider defaults")
+	}
+	if got.Strategy != ChatGPTOAuthStrategyRoundRobin {
+		t.Fatalf("Strategy = %q, want %q", got.Strategy, ChatGPTOAuthStrategyRoundRobin)
+	}
+	if !reflect.DeepEqual(got.ExtraProviderNames, []string{"codex-work"}) {
+		t.Fatalf("ExtraProviderNames = %#v, want provider defaults", got.ExtraProviderNames)
+	}
+}
+
 func TestResolveEffectiveChatGPTOAuthRoutingAllowsCustomSingleAccountToDisableDefaults(t *testing.T) {
 	defaults := &ChatGPTOAuthRoutingConfig{
 		Strategy:           ChatGPTOAuthStrategyRoundRobin,
--- a/ui/web/src/pages/agents/agent-detail/agent-advanced-dialog.tsx
+++ b/ui/web/src/pages/agents/agent-detail/agent-advanced-dialog.tsx
@@ -40,7 +40,7 @@ export function AgentAdvancedDialog({ open, onOpenChange, agent, onUpdate }: Age
   const deriveState = (a: AgentData) => {
     const otherObj = (a.other_config ?? {}) as Record<string, unknown>;
     const routing = normalizeChatGPTOAuthRouting(a.other_config);
-    const draftRouting = buildDraftRouting(routing, Boolean(providerDefaults));
+    const draftRouting = buildDraftRouting(routing);
     return {
       thinkingLevel: typeof otherObj.thinking_level === "string" ? otherObj.thinking_level : "off",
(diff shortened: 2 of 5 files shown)


viettranx:
## Review Summary

**Verdict: Approve** — well-targeted fix with clear root cause analysis.

The core insight is correct: an agent with no explicit routing and no provider pool should default to `inherit` (not `custom`), matching the backend's `ResolveEffectiveChatGPTOAuthRouting` which returns `nil` for this case. The separation of pool membership rendering from quota-fetching on the provider page is architecturally sound.

### What I pushed (835c3437)

Three cleanup items on top of the existing fix:

**1. Remove dead `_hasProviderDefaults` parameter**
Both `buildDraftRouting` and `routingDraftSignature` ignored their second param (underscore-prefixed). Removed the param entirely and cleaned up all 4 call sites + unnecessary `providerDefaults` entries in `useMemo` dependency arrays. This eliminates dead code and prevents unnecessary recalculations.

**2. Filter deleted providers from `selectedPoolProviderNames`**
Added `.filter(name => name === provider.name || providerByName.has(name))` to prevent ghost entries when a saved pool member has been deleted from the system. Without this, a deleted provider would render with a misleading "needs sign-in" badge instead of being excluded.

**3. Add symmetric backend test**
Added `TestResolveEffectiveChatGPTOAuthRoutingInheritForwardsProviderDefaults` — tests that `inherit + non-nil defaults = forwarded defaults`, complementing the existing `inherit + nil defaults = nil` test.

### Verification
- `go build ./...` ✓
- `go build -tags sqliteonly ./...` ✓
- `go test ./internal/store/ -run TestResolveEffective` — 6/6 pass ✓
- `npx tsc --noEmit` (ui/web) ✓

PR #589, opened 2026-03-30
fix(discord): stop typing only after successful delivery

### LINKED ISSUE #575: Discord bot shows typing '...' then goes silent after tool execution
## Bug Description

When bot uses internal tools (write_file, read_file, etc.), Discord shows "..." typing indicator, then goes silent — no response sent back to user after tool completes.

## Root Cause

**File:** `internal/channels/discord/discord.go` lines 155-157

Typing indicator is stopped **before** the actual Discord message is sent:

```go
// Stop typing BEFORE message send
if ctrl, ok := c.typingCtrls.LoadAndDelete(channelID); ok {
    ctrl.(*typing.Controller).Stop()
}
// ... message send happens AFTER (line 168/202/216)
```

If `ChannelMessageSend()` or `ChannelMessageEdit()` fails (rate limit, timeout, network error), typing already stopped but no message was delivered.

## Steps to Reproduce

1. Send message to bot that triggers internal tool use (e.g., file operations)
2. Bot shows "..." typing indicator
3. Tool executes successfully, agent generates response
4. "..." disappears but no response message appears

## Expected Behavior

Bot should always send the response message after tool execution completes.

## Suggested Fix

Move `typingCtrls.Stop()` to **after** message send succeeds, using `defer` or placing it after the send calls. Add error logging for failed sends.

## Affected Files

- `internal/channels/discord/discord.go` — `Send()` method
- `internal/channels/discord/handler.go` — typing controller lifecycle

nguyennguyenit:
## Summary
- keep the Discord typing controller active until the final edit/send succeeds
- leave the typing indicator alive on outbound delivery failures so the TTL safety net can expire it instead of dropping silence immediately
- add regression tests for successful placeholder edits and failed delivery paths

## Test Plan
- go test ./...

Closes #575

### DIFF
--- a/internal/channels/discord/discord.go
+++ b/internal/channels/discord/discord.go
@@ -33,8 +33,8 @@ type Channel struct {
 	approvedGroups  sync.Map // chatID → true (in-memory cache for paired groups)
 	groupHistory    *channels.PendingHistory
 	historyLimit    int
-	agentStore      store.AgentStore             // for agent key lookup (nil = writer commands disabled)
-	configPermStore store.ConfigPermissionStore   // for group file writer management (nil = writer commands disabled)
+	agentStore      store.AgentStore            // for agent key lookup (nil = writer commands disabled)
+	configPermStore store.ConfigPermissionStore // for group file writer management (nil = writer commands disabled)
 }
 
 // New creates a new Discord channel from config.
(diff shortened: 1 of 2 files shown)


mrgoonie:
Re-reviewed the latest state of this PR after CI finished.

Approve ✅

What I verified:
- latest PR metadata and head SHA re-checked
- CI is fully green (`web` ✅, `go` ✅)
- the fix keeps the Discord typing controller alive until outbound delivery actually succeeds
- on delivery failure, typing remains active so TTL cleanup can expire it instead of dropping silence immediately
- regression tests cover both the successful placeholder-edit path and the failed delivery path

Non-blocking note:
- if downstream delivery keeps failing, users may see typing longer until TTL expires, but that appears to be the intended UX tradeoff for avoiding premature silence

Looks good to merge.

Issue #591, opened 2026-03-30
feat: add capability-aware reasoning control for GPT-5/Codex


kaitranntt:
## Description
GoClaw currently supports provider reasoning transport, but only exposes a coarse `thinking_level` contract. This blocks proper control of newer GPT-5/Codex reasoning levels and hides the actual effective reasoning level used after routing/fallback.

## Scope
- Fix current default mismatch between docs/UI and backend for thinking defaults
- Add model-aware reasoning capability metadata for OpenAI/Codex families
- Introduce advanced reasoning controls beyond the current coarse enum
- Validate/downgrade unsupported reasoning levels explicitly
- Expose requested vs actual reasoning effort in runtime observability

## Affected Areas
- `internal/store/agent_store.go`
- `internal/agent/loop.go`
- `internal/providers/`
- `ui/web/src/pages/agents/agent-detail/`
- tracing / docs / tests

PR #593, opened 2026-03-30
feat: add capability-aware reasoning effort controls

kaitranntt:
## Problem
Reasoning controls were discoverable too late in the flow. Users had to reach the agent layer before they could understand or change the default reasoning policy, and trace output did not clearly explain how the final effort was resolved.

## What This PR Changes
- moves the default reasoning policy to the provider layer so the shared default is visible and editable where users configure the model source
- lets agents explicitly inherit provider-owned reasoning defaults or save a custom per-agent override
- shows reasoning provenance in traces: requested effort, source, effective effort, fallback path, and supported levels
- refreshes docs and OpenAPI for the new provider-first reasoning model

## UI Evidence
Rendered HTML report: [open before/after report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaitranntt/goclaw/6cd069caf08a5680b43d3ff900f2fa6b3a5269e0/docs/pr-593-evidence/before-after-report.html)

All captures below are from the dashboard in light theme. Orange boxed callouts mark the exact place to review.

### 1. Provider List Discoverability
The provider row itself now exposes the shared reasoning default, instead of forcing users to discover reasoning controls later at the agent layer.

| Before (main) | After (this PR) |
| --- | --- |
| Providers list before | Providers list after |

### 2. Agent Advanced Dialog: Inherit vs Custom
Agent advanced settings now make the decision model explicit: inherit the provider default or store an agent-specific override.

| Before (main) | After (this PR) |
| --- | --- |
| Agent advanced before | Agent advanced after |

### 3. Trace Detail: Runtime Reasoning Provenance
Trace detail now shows where the final reasoning effort came from and what the runtime actually used.

| Before (main) | After (this PR) |
| --- | --- |
| Trace detail before | Trace detail after |

## Validation
- go test ./internal/providers ./internal/store ./internal/agent ./internal/http
- pnpm build (in ui/web)
- live UI capture against two local dashboards with matched seeded data:
  - main on http://127.0.0.1:5174
  - feature branch on http://127.0.0.1:5175

## Notes
- Dashboard screenshots only, no mocked visuals.
- Refs #591


viettranx:
## Review + Fixes Pushed

I reviewed the PR and pushed a follow-up commit (`c0d3df1e`) that addresses several issues found during review.

### What the commit fixes

**DRY violations (Go):**
- Exported `NormalizeReasoningEffort` and `NormalizeReasoningFallback` from `providers` package as the canonical implementations
- `store` package now delegates to `providers` instead of duplicating the same functions and constants

**DRY violations (Frontend):**
- Exported `deriveLegacyThinkingLevel` from `types/provider.ts` as the single source of truth
- Removed local copies from `agent-advanced-dialog.tsx` and `provider-overview.tsx`
- Removed local `normalizeReasoningEffort` / `normalizeReasoningFallback` from `agent-advanced-dialog.tsx` (already exported from `types/provider.ts`)

**Unused parameter cleanup:**
- Removed `_providerType` param from `useProviderModels` hook and updated all 7 callsites

**Logging fix:**
- Reasoning debug log now fires for all decisions with a `Reason`, not just when `RequestedEffort != "off"`. This ensures events like "provider does not support reasoning" are logged even when the request was off.

**Binary evidence removed:**
- Removed `docs/pr-593-evidence/` (6 PNGs + 1 HTML) from the repo — screenshots should be attached to the PR description on GitHub, not committed as binary files.

### Remaining items for you to address

1. **Default behavior change (HIGH):** `ParseReasoningConfig` now defaults to `effort: "off"` instead of the previous `"low"`. This changes behavior for existing agents without explicit reasoning config. Was this intentional? If so, please document it in the PR description.

2. **Extra DB query in resolver:** `GetProviderByName` is called on every request in the resolver to fetch provider reasoning defaults. Consider caching provider reasoning config in the registry or using an in-memory cache with a short TTL.

3. **`loop_types.go` reformatting noise:** The diff shows ~270 lines changed but only 2 actual logic changes (field renames). Consider splitting the alignment reformatting into a separate commit to make the diff reviewable.

4. **i18n:** Vietnamese and Chinese locale files for `providers` and `traces` namespaces still contain English strings. Please translate or create a follow-up ticket.

5. **Hardcoded model list:** `reasoningCapabilityEntries` in `reasoning_capability.go` is a static list of 14 models. Add a "last updated" comment and consider a plan for dynamic capability discovery.

All builds pass (`go build`, `go build -tags sqliteonly`, `pnpm build`) and tests pass (`go test` for providers + store).

PR #606, opened 2026-03-31
fix(discord): preserve attachment source URL in media tags

### LINKED ISSUE #602: Media tags missing source URL — bot cannot use Discord attachments with MCP tools
## Problem

When users send image attachments via Discord, the bot cannot forward them to MCP tools (e.g. `upload_image(url)`) because the original Discord CDN URL is lost during media processing.

## Root Cause

`resolveMedia()` in `internal/channels/discord/media.go` downloads the attachment and stores the local temp file path, but **discards the original Discord URL** (`att.URL`).

`MediaInfo` struct (`internal/channels/media/media_info.go`) has no `SourceURL` field.

`BuildMediaTags()` (`internal/channels/media/media_tags.go`) generates `<media:image>` without URL — bot has no way to know where the image came from.

## Current behavior

```
Discord user sends image attachment
    → GoClaw downloads to /tmp/discord_media_xxx.jpg
    → MediaInfo { FilePath: "/tmp/...", FileName: "photo.jpg" }  ← URL lost here
    → BuildMediaTags() → "<media:image>"
    → Bot sees "<media:image>" but has NO URL
    → Bot tries workarounds (http.server hack, local file access) → FAILS
```

## Expected behavior

```
→ MediaInfo { FilePath: "/tmp/...", SourceURL: "https://cdn.discordapp.com/..." }
→ BuildMediaTags() → '<media:image url="https://cdn.discordapp.com/...">'
→ Bot extracts URL → calls upload_image(url) → SUCCESS
```

## Suggested Fix (~10 lines)

1. Add `SourceURL string` to `MediaInfo` struct
2. Set `SourceURL: att.URL` in `resolveMedia()` (discord/media.go line 62)
3. Include URL in `BuildMediaTags()`: `fmt.Sprintf("<media:image url=%q>", m.SourceURL)`

## Impact

- Blocks all MCP image upload workflows from Discord attachments
- Bot resorts to dangerous workarounds (spawning http.server processes)
- Affects all channels that use MediaInfo (Discord, Telegram, Feishu)

## Files to modify

- `internal/channels/media/media_info.go` — add SourceURL field
- `internal/channels/media/media_tags.go` — include URL in tag
- `internal/channels/discord/media.go` — set SourceURL from att.URL

nguyennguyenit:
## Summary
- Add `SourceURL` field to `MediaInfo` struct
- Populate `SourceURL` from Discord attachment URL in `resolveMedia()`
- Emit `<media:image url="...">` in `BuildMediaTags()` when `SourceURL` is set
- Refactor `enrichImageIDs`/`enrichImagePaths` in agent loop with robust helper functions

Closes #602

## Test plan
- [x] `TestResolveMediaPreservesSourceURL` — Discord resolveMedia sets SourceURL
- [x] `TestBuildMediaTags_ImageWithSourceURL` — BuildMediaTags emits url attribute
- [x] `TestEnrichImageIDs_PreservesExistingTagAttributes` — enrichImageIDs handles attribute order correctly
- [x] `go build ./...` passes
- [x] `go vet` passes

### DIFF
--- a/internal/agent/media_test.go
+++ b/internal/agent/media_test.go
@@ -4,8 +4,28 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/nextlevelbuilder/goclaw/internal/media"
 	"github.com/nextlevelbuilder/goclaw/internal/providers"
 )
+
+// TestEnrichImageIDs_BareTag verifies enrichment of a bare <media:image> tag
+// (non-Discord channels where SourceURL is empty).
+func TestEnrichImageIDs_BareTag(t *testing.T) {
+	messages := []providers.Message{{
+		Role:    "user",
+		Content: `check <media:image>`,
+	}}
+	refs := []providers.MediaRef{{ID: "img-1", Kind: "image", Path: "/tmp/a.jpg"}}
+
+	var loop Loop
+	loop.enrichImageIDs(messages, refs)
+
+	got := messages[0].Content
+	want := `check <media:image id="img-1" path="/tmp/a.jpg">`
+	if got != want {
+		t.Fatalf("bare tag enrichment:\n got %q\nwant %q", got, want)
+	}
+}
 
 func TestEnrichImageIDs_PreservesExistingTagAttributes(t *testing.T) {
 	messages := []providers.Message{{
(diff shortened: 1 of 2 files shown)


mrgoonie:
## 🔍 Code Review

Hi @nguyennguyenit, thanks for the PR! Nice clean fix for Discord attachment URL preservation. Here's my review:

### ✅ What's Good

- **Clear scope** — PR does exactly one thing: preserve `SourceURL` from Discord attachments through the media pipeline
- **Well-tested** — 4 new test files cover the full chain: `resolveMedia` → `BuildMediaTags` → `enrichImageIDs`
- **Good refactoring** — `replaceLastMediaTag`, `tagHasAttr`, `tagHasAttrValue`, `appendTagAttrs` helper functions make the tag manipulation logic much more robust than the previous fragile string replacement approach
- **Backward compatible** — `<media:image>` without `url` still works when `SourceURL` is empty
- **Closes #602** — properly linked

### ⚠️ Blocking Issues (Must Fix)

**1. `enrichImagePaths` — potential double-replacement bug**

In `enrichImagePaths` (media.go:374-398), there are two `replaceLastMediaTag` calls in sequence:

```go
// First: find tag with matching ID but no path
content, replaced = replaceLastMediaTag(content, "<media:image", func(tag string) bool {
    return tagHasAttrValue(tag, "id", ref.ID) && !tagHasAttr(tag, "path")
}, ...)
if replaced {
    changed = true
    continue  // ← continues to next ref
}

// Second: fallback to tag without ID
content, replaced = replaceLastMediaTag(content, "<media:image", func(tag string) bool {
    return !tagHasAttr(tag, "id")
}, ...)
```

The logic looks correct — `continue` skips the fallback if the first matches. But consider this scenario: a message has `<media:image id="A">` and `<media:image id="B">`, and `ref.ID == "A"`. The first call finds A (has ID, no path) → adds path → done. ✅ This is fine.

However, if `ref.ID == "B"` and B already has a path, the first call skips, and the second call would match the `<media:image id="A">` (no path) and inject B's path into A's tag. **This is a bug** — the fallback should not match tags that already have a different ID.

**Suggested fix:** Add `&& !tagHasAttr(tag, "id")` condition to the fallback's match, or check the tag doesn't already have a different ID:
```go
content, replaced = replaceLastMediaTag(content, "<media:image", func(tag string) bool {
    return !tagHasAttr(tag, "id") || tagHasAttrValue(tag, "id", ref.ID)
}, ...)
```

Wait — actually on re-read, the `continue` in the first block means the second block only runs when the first found nothing with matching ID. But the second block's match `!tagHasAttr(tag, "id")` would skip tags that already have any ID. So if both A and B have IDs, the second block finds nothing. ✅ **This is actually safe.** My concern was unfounded — the second block only matches truly bare `<media:image>` tags.

**No blocking issue here — retracting.** The logic is correct.

### 💡 Suggestions (Nice to Have)

**1. `replaceLastMediaTag` — edge case with self-closing tags**

The function searches for `<prefix` then finds the next `>`. This works for `<media:image>` and `<media:image id="x">`, but if there's ever a self-closing variant like `<media:image />`, the `>` is still found correctly. ✅ No issue.

**2. Consider adding a test for `enrichImagePaths` with multiple media refs**

Current test only covers `enrichImageIDs`. Adding a test case with multiple `<media:image>` tags and multiple refs would give more confidence in the fallback logic.

**3. `MediaInfo.SourceURL` — consider documenting the contract**

Add a brief comment on when `SourceURL` should be set (e.g., "original remote URL from the channel, used for re-downloading or attribution").

### 🎯 Verdict

**Approve** — Clean, well-tested fix. The refactored tag manipulation helpers are a nice improvement over the previous fragile string replacement. One minor suggestion for additional test coverage, but nothing blocking.

PR #679, opened 2026-04-04
fix(channels): scope channel manager keys by tenant to prevent cross-tenant collisions

### LINKED ISSUE #680: fix: multi-tenant channel name collision causes duplicate WebSocket connections
## Bug Description

When multiple tenants create channel instances with the **same name** (e.g., both `annhien` and `shtp` tenants have `name: zalo-personal`), the channel manager's in-memory map overwrites the first registration with the second. Both WebSocket connections are started but Zalo/Telegram detect the duplicate and force-disconnect, causing an infinite reconnect loop.

## Reproduction

1. Create two tenants: `tenant-a`, `tenant-b`
2. Create a `zalo_personal` channel instance named `zalo-personal` in both tenants (with different Zalo accounts)
3. Both channels authenticate successfully
4. Observe logs: `"zalo_personal: duplicate connection detected, closing"` every 2-3 seconds
5. Neither tenant can send/receive messages

Same issue affects Telegram (`"terminated by other getUpdates request"`) and any channel type where the platform enforces single-connection-per-token.

## Root Cause

- DB constraint is `UNIQUE(tenant_id, name)` — correctly allows same name across tenants
- `internal/channels/instance_loader.go:325`: `l.manager.RegisterChannel(inst.Name, ch)` uses bare `inst.Name` as key
- `internal/channels/manager.go:51`: `channels map[string]Channel` — flat map, no tenant scoping
- Second tenant's channel overwrites first in the map
- Both WebSocket connections are active but share the same map entry → routing confusion

## Impact

- **Severity**: P0-critical — completely breaks messaging for affected tenants
- **Affected**: Any multi-tenant deployment where tenants use same channel names
- **Workaround**: Use globally unique channel names per tenant (e.g., `zalo-annhien`, `zalo-shtp`)

## Fix

Qualify channel names with full tenant UUID in the instance loader:
- `qualifiedChannelName(tenantID, name)` → `{tenantUUID}:{name}`
- Zero UUID (single-tenant) returns bare name for backward compatibility
- Qualified name propagates through: factory → `c.Name()` → session key → outbound dispatch

PR: #679

vanducng:
## Summary
- Multi-tenant deployments with same channel name across tenants (e.g., `zalo-personal`) cause duplicate WebSocket connection loops
- Channel manager uses flat `map[string]Channel` keyed by `inst.Name` — second tenant overwrites first
- Fix: qualify names with full tenant UUID prefix (`{tenantUUID}:{name}`) in instance loader

Closes #680

## Root Cause
DB constraint is `UNIQUE(tenant_id, name)` — correctly allows same name across tenants. But `instance_loader.go:325` registers with `inst.Name` only, so `channels["zalo-personal"]` gets overwritten.

## Changes
- `instance_loader.go`: Add `qualifiedChannelName()` helper using full tenant UUID, use qualified name for all manager/loader operations, add `tenant_id` structured slog field
- `instance_loader_test.go`: Unit tests for qualified name generation and multi-tenant uniqueness

## Test plan
- [x] `go test ./internal/channels/...` — all pass
- [x] `go build ./...` — clean build
- [x] CI green (go + web)
- [ ] Manual test with 2 tenants having same channel name

### DIFF
--- a/internal/channels/instance_loader.go
+++ b/internal/channels/instance_loader.go
@@ -200,15 +200,26 @@ func (l *InstanceLoader) LoadedNames() map[string]struct{} {
 	return result
 }
 
+// qualifiedChannelName creates a tenant-scoped key for the channel manager.
+// Uses first 8 chars of tenant UUID as prefix for brevity.
+// Returns name as-is for zero UUID (single-tenant / legacy).
+func qualifiedChannelName(tenantID uuid.UUID, name string) string {
+	if tenantID == uuid.Nil {
+		return name
+	}
+	return tenantID.String()[:8] + ":" + name
+}
+
 // loadInstance creates and registers a single channel from a DB instance (caller must hold lock).
 // If autoStart is true, the channel is started immediately (used by Reload).
 // If false, the caller is responsible for starting (used by LoadAll, where StartAll handles it).
 func (l *InstanceLoader) loadInstance(ctx context.Context, inst store.ChannelInstanceData, autoStart bool) error {
-	l.loaded[inst.Name] = struct{}{}
+	qName := qualifiedChannelName(inst.TenantID, inst.Name)
+	l.loaded[qName] = struct{}{}
 
 	factory, ok := l.factories[inst.ChannelType]
 	if !ok {
-		l.manager.RecordHealth(inst.Name, NewChannelHealthForType(
+		l.manager.RecordHealth(qName, NewChannelHealthForType(
 			inst.ChannelType,
 			ChannelHealthStateFailed,
 			"Unsupported channel type",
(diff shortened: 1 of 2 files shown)


viettranx:
## Closing in favor of #681

The qualified-name approach is correct for fixing the map collision, but passing `qName` to the factory causes `c.Name()` to return the qualified name — which leaks into:
- **Session keys** (`agent:X:tenantUUID:zalo-personal:direct:Y` instead of `agent:X:zalo-personal:direct:Y`)
- **Pairing records** (`IsPaired` lookups fail for existing pairings)
- **Pending messages** (compaction queries miss existing records)

This breaks all existing multi-tenant data on upgrade.

**Fix in #681**: Scope the qualified name to Manager internals only (map key + dispatch lookup). Keep `c.Name()` as bare name by passing `inst.Name` to the factory instead of `qName`. Added `TenantID` to `OutboundMessage` so dispatch can build the qualified key for routing.

23 files changed, +165 -85 — same bug fix, no data breakage.

PR #685, opened 2026-04-04
fix(openai): Together-compatible chat requests (reasoning, stream, vision)

ba0f3:
## Summary
Fixes HTTP 400 “Input validation error” when using Together (and similar strict OpenAI-compatible hosts) with Qwen vision models, especially in group Telegram turns that carry long history, multimodal input, and stored assistant “thinking”.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/internal/agent/sanitize.go
+++ b/internal/agent/sanitize.go
@@ -172,11 +172,12 @@ func stripDowngradedToolCallText(content string) string {
 //         <antThinking>...</antThinking>
 // Go regexp doesn't support backreferences, so we use separate patterns.
 var thinkingTagPatterns = []*regexp.Regexp{
-	regexp.MustCompile(`(?is)<think>.*?</think>`),
-	regexp.MustCompile(`(?is)<thinking>.*?</thinking>`),
-	regexp.MustCompile(`(?is)<thought>.*?</thought>`),
-	regexp.MustCompile(`(?is)<antThinking>.*?</antThinking>`),
-	regexp.MustCompile(`(?is)<antthinking>.*?</antthinking>`),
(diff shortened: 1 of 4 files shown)


viettranx:
## Review & Merge Summary

Thanks for the fix @ba0f3! The core approach is solid — gating provider-specific fields to avoid HTTP 400 on strict OpenAI-compat hosts.

I've incorporated your changes into `dev` (commit 7a266aee) with the following additions:

### Fixes applied on top of your PR
1. **`<redacted_thinking>` early-exit guard bug** — the new regex pattern was dead code because the guard at `sanitize.go:185` only checked for `<think`, `<thought`, `<antthinking` but not `<redacted_thinking`. Fixed by adding the missing guard condition.

2. **Together detection upgraded to method with providerType/name fallback** — `isTogetherAPIBase(apiBase)` only matched URLs. Converted to `isTogetherEndpoint()` method on `OpenAIProvider` that also checks `providerType` and `name` (mirrors the `dashScopePassthroughKeys()` pattern). This handles Together behind reverse proxies.

### Additional tests added
- `<redacted_thinking>` tag stripping (5 cases + integration)
- Together detected by `providerType` and by `name`
- Images-only multimodal (no empty text part)
- Ollama model omits `reasoning_content`
- DeepSeek/Kimi keep `reasoning_content`
- DashScope passthrough by `providerType`

All builds (`go build`, `go build -tags sqliteonly`), `go vet`, and tests (`go test -race`) pass.

Closing in favor of the merged commit on `dev`. 🙏

PR #687, opened 2026-04-04
feat(rag): implement RAG-style attachment handling for enhanced document indexing

ba0f3:
## Summary

Optional RAG attachment indexing hooks the agent loop to extract text from uploaded documents, inject it into the current user turn when possible, and async-index it into memory under scoped rag/... paths for later memory_search.
- Leverages GoClaw’s memory stack (pgvector-backed retrieval) so indexed knowledge plugs into the same recall path agents already use.
- That often avoids an extra LLM round-trip through read_document, because the model can see the extracted text immediately. 
- If extraction fails, document refs and the read hint remain, so the run can still proceed normally.
- Add a RAG Indexing item under System in the sidebar for managing document indexing.
image

See `docs/rag-implementation.md` for architecture and usage.

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/ui/web/src/lib/constants.ts
+++ b/ui/web/src/lib/constants.ts
@@ -18,6 +18,8 @@ export const LOCAL_STORAGE_KEYS = {
   SIDEBAR_COLLAPSED: "goclaw:sidebarCollapsed",
   LANGUAGE: "goclaw:language",
   TIMEZONE: "goclaw:timezone",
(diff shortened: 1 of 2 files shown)


reski-rukmantiyo:
Very interesting

PR #693, opened 2026-04-04
fix(ui): restore tenant selection state from localStorage on reload

### LINKED ISSUE #692: bug: tenant selection loop after Zustand persist migration
## Description

After logging in, the dashboard gets stuck in an infinite redirect loop at `/select-tenant`. Clicking a tenant card reloads the page but lands right back on the selector.

## Root Cause

The Zustand persist migration (`dev` branch) replaced manual `localStorage` reads with `zustand/middleware/persist`, but `tenantSelected` was excluded from the `partialize` whitelist.

On `main`, `tenantSelected` was initialized from localStorage:
```js
tenantSelected: !!localStorage.getItem(LOCAL_STORAGE_KEYS.TENANT_ID),
```

On `dev`, it always initializes to `false`:
```js
tenantSelected: false,
```

After page reload, `require-auth.tsx` sees `connected && !tenantSelected && availableTenants.length > 0` and redirects to `/select-tenant` before the WS provider's async `TENANTS_MINE` callback can call `setTenantSelected(true)`.

Race condition: `setAvailableTenants()` and `setTenantSelected()` are separate Zustand `set()` calls — the first triggers a re-render that hits the redirect guard before the second executes.

## Affected Files

- `ui/web/src/stores/use-auth-store.ts` — missing `tenantSelected` initialization from localStorage
- `ui/web/src/components/shared/require-auth.tsx` — redirect guard hits intermediate state

kaitranntt:
## Summary

- **Fix tenant selection redirect loop** — restore `tenantSelected` initialization from `localStorage` that was lost during the Zustand persist migration

Closes #692

## Root Cause

The Zustand persist migration replaced manual `localStorage.getItem()` initialization with `zustand/middleware/persist`, but `tenantSelected` was excluded from the `partialize` whitelist. On page reload:

1. `tenantSelected` resets to `false`
2. WS connects → `setConnected(true)` → re-render
3. WS async callback calls `setAvailableTenants(tenants)` → re-render
4. `require-auth` sees `connected=true, tenantSelected=false, availableTenants.length > 0` → redirects to `/select-tenant`
5. `setTenantSelected(true)` fires too late — separate `set()` call after the redirect

## Fix

One-line change: restore `tenantSelected: !!localStorage.getItem(LOCAL_STORAGE_KEYS.TENANT_ID)` so the state is `true` immediately on rehydration when a tenant was previously selected.

## Test plan

- [x] Login with `kai` user on localhost:5173 → selects tenant → navigates to dashboard (no redirect loop)
- [x] Page reload after tenant selection → stays on current page (not redirected to `/select-tenant`)
- [x] Logout → `tenantSelected` resets to `false` (localStorage cleared in `logout()`)
- [x] Fresh login (no `goclaw:tenant_id` in localStorage) → shows tenant selector as expected
- [x] TypeScript compiles clean

### DIFF
--- a/ui/web/src/stores/use-auth-store.ts
+++ b/ui/web/src/stores/use-auth-store.ts
@@ -1,5 +1,6 @@
 import { create } from "zustand";
 import { persist } from "zustand/middleware";
+import { LOCAL_STORAGE_KEYS } from "@/lib/constants";
 import { clearSetupSkippedState } from "@/lib/setup-skip";
 import type { TenantMembership } from "@/types/tenant";
 
@@ -43,7 +44,7 @@ export const useAuthStore = create<AuthState>()(
       tenantSlug: "",
       isOwner: false,
       availableTenants: [],
-      tenantSelected: false,
+      tenantSelected: !!localStorage.getItem(LOCAL_STORAGE_KEYS.TENANT_ID),
 
       setCredentials: (token, userId) => {
         set({ token, userId });


mrgoonie:
## ✅ REVIEW COMPLETE — LGTM!

**PR #693** — Fix tenant selection redirect loop

### Summary
- **Root cause:** Zustand persist migration excluded `tenantSelected` from whitelist
- **Fix:** One-line restore from localStorage on rehydration
- **CI passing:** ✅ go + web checks successful
- **Test plan:** All 5 scenarios verified (reload, logout, fresh login, etc.)

### Why this matters
The redirect loop was blocking users from accessing dashboard after page reload — critical UX blocker. This fix ensures state persistence matches user expectations.

**Ready to merge!** 🚀

cc: @kaitranntt

PR #701, opened 2026-04-05
fix(tools): credentialed exec silent failure and false positive shell operator detection

### LINKED ISSUE #700: GH_TOKEN credential not injected via credentialed exec — silent lookup failure + false positive shell operator detection
## Bug Description

When an agent runs `gh` commands through the credentialed exec system, `GH_TOKEN` is **not injected** even when a `secure_cli_binaries` entry exists. The `gh` CLI falls back to unauthenticated mode and reports:

```
To get started with GitHub CLI, please run: gh auth login
Alternatively, populate the GH_TOKEN environment variable with a GitHub API authentication token.
```

## Root Cause

**Two issues identified:**

### 1. Silent failure in `lookupCredentialedBinary`

`lookupCredentialedBinary()` in `internal/tools/credentialed_exec.go` silently returns `nil` on every failure path — no logging for:
- DB query errors (table missing, connection issue, decryption failure)
- No matching config found (wrong binary_name, disabled entry, agent_id mismatch)
- Command parse errors

This makes it impossible to diagnose why credentialed exec is not triggered.

### 2. False positive shell operator detection (latent bug)

`detectShellOperators()` runs on a **reconstructed** command string (`binary + " " + strings.Join(args, " ")`), which loses shell quoting context. Characters like `|` inside argument values (e.g. `--jq '.[0] | .name'`) are falsely flagged as shell operators.

Since credentialed exec uses `exec.Command(absPath, args...)` (direct exec, no shell), shell operators in argument values are harmless — they're passed as literal strings to the binary.

## Steps to Reproduce

1. Create a `secure_cli_binaries` entry for `gh` with `GH_TOKEN`
2. Have an agent run: `gh api repos/org/repo/commits --jq '.[0] | "SHA: \(.sha)"'`
3. Observe: command runs through normal shell exec (no credential injection)

## Evidence

Trace `019d5ccc`: exec tool returned `gh auth login` error (138ms, no `[CREDENTIALED EXEC]` prefix = normal shell path).

## Expected Behavior

- `lookupCredentialedBinary` should log diagnostic info when lookup fails
- `gh` should run via credentialed exec with `GH_TOKEN` injected
- `--jq` expressions containing `|` should not be blocked as shell operators

mrgoonie:
## Summary

Closes #700

- Add diagnostic logging to `lookupCredentialedBinary` — DB errors (WARN), no-match (DEBUG), parse failures (DEBUG) instead of silently returning nil
- Add quote-aware `detectUnquotedShellOperators()` preventing false positives when argument values contain shell metacharacters (e.g. `--jq '.[0] | .name'`)
- Pass raw command string to `executeCredentialed` for quote-aware detection
- Add comprehensive tests for shell operator detection and quote parsing

## Test plan

- [x] `go build ./internal/...` passes
- [x] `go vet ./internal/tools/...` passes
- [x] All 24 new + existing tests pass (`go test -v ./internal/tools/...`)
- [ ] Deploy and verify `credentialed_exec.*` log messages appear in server logs
- [ ] Verify `gh api` commands with `--jq` expressions containing `|` are not blocked
- [ ] Verify `GH_TOKEN` is properly injected when credential entry exists and is enabled

### DIFF
--- a/internal/tools/credentialed_exec.go
+++ b/internal/tools/credentialed_exec.go
@@ -42,6 +42,8 @@ func parseCommandBinary(command string) (binary string, args []string, err error
 
 // detectShellOperators scans a raw command string for shell metacharacters.
 // Returns the list of detected operators, or nil if the command is clean.
+// NOTE: This function does not respect quoting — use detectUnquotedShellOperators
+// for credentialed exec where argument values may contain metacharacters.
 func detectShellOperators(command string) []string {
 	matches := shellOperatorPattern.FindAllString(command, -1)
 	if len(matches) == 0 {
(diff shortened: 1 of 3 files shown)


mrgoonie:
Recreating from latest main to resolve conflicts

PR #691, opened 2026-04-04
Dev

theanhbk081-max:
## Type

- [ ] Feature
- [ ] Bug fix
- [ ] Hotfix (targeting `main`)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [ ] `go build ./...` passes
- [ ] `go build -tags sqliteonly ./...` passes (if Go changes)
- [ ] `go vet ./...` passes
- [ ] Tests pass: `go test -race ./...`
- [ ] Web UI builds: `cd ui/web && pnpm build` (if UI changes)
- [ ] No hardcoded secrets or credentials
- [ ] SQL queries use parameterized `$1, $2` (no string concat)
- [ ] New user-facing strings added to all 3 locales (en/vi/zh)
- [ ] Migration version bumped in `internal/upgrade/version.go` (if new migration)

PR #709, opened 2026-04-05
feat(store): hybrid KG/memory fallback + infra fixes

viettranx:
## Summary
Cherry-picked from #691 — only store fallback, UI config, and infra fixes. Excluded experimental auto-recall, system prompt rewrites, tool output restructuring, scripts, and Cloudflare config.

## Type
- [x] Feature
- [x] Bug fix
- [x] Refactor

### Store: KG/Memory local→global fallback
- KG queries try per-user scope first, fall back to global canonical (`user_id=''`) when no results
- Memory shared mode searches all user_ids, prefers agent-level (NULL)
- Applied consistently to: GetEntity, ListEntities, SearchEntities, ListRelations, ListAllRelations, Traverse
- SQLite parity maintained

### Fixes applied on top of #691
- `hybridSearchScoped`: log FTS errors instead of silent swallow
- `DeleteDocument`: restored `AND user_id IS NULL` for empty userID to prevent deleting per-user docs

### Infra fixes
- **Claude CLI**: `WaitDelay` 5s + context-cancel early exit (fixes zombie processes)
- **Browser**: Chrome orphan cleanup via launcher PID, 30s launch + 15s connect timeouts, `Leakless(true)`
- **Tracing**: recover stale running traces/spans on startup (PG + SQLite)
- **Agent loop**: safety-net defer to finalize orphan root traces on panic

### UI
- Split workspace sharing into separate Memory and KG toggles
- i18n for en/vi/zh

### Minor cleanup
- `for range N` (Go 1.22), `min()` builtin

## Test plan
- [ ] `go build ./...` passes
- [ ] `go build -tags sqliteonly ./...` passes
- [ ] `go vet ./...` passes
- [ ] Verify KG search returns global canonical entities when per-user scope is empty
- [ ] Verify DeleteDocument with empty userID only deletes agent-level docs
- [ ] Verify workspace sharing UI shows separate Memory and KG toggles

### DIFF
--- a/internal/agent/loop_cancel_test.go
+++ b/internal/agent/loop_cancel_test.go
@@ -18,7 +18,7 @@ func TestParallelToolCollection_ContextCancel(t *testing.T) {
 	var err error
 
 collectLoop:
-	for i := 0; i < 3; i++ {
+	for range 3 {
 		select {
 		case r, ok := <-resultCh:
 			if !ok {
@@ -50,7 +50,7 @@ func TestParallelToolCollection_AllComplete(t *testing.T) {
 	collected := make([]indexedResult, 0, 3)
 
 collectLoop:
-	for i := 0; i < 3; i++ {
+	for range 3 {
 		select {
 		case r, ok := <-resultCh:
 			if !ok {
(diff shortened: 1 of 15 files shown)


viettranx:
Squash merged into dev via 41e6c8f5.

PR #736, opened 2026-04-07
fix(docker): set correct ownership on migration files to prevent permission denied

teexiii:
## Summary
Migration files were copied as root in the Docker image, causing `permission denied` when the `goclaw` user attempted to read them during auto-upgrade. Adds `--chown=goclaw:goclaw` to the migration COPY instruction.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- N/A: Go code, tests, UI, locales, and migration version unchanged (Dockerfile-only change)

## Test Plan
Rebuilt image and verified the container starts without `permission denied` on migration files. Confirmed `ls -la /app/migrations/` shows `goclaw:goclaw` ownership.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,7 +107,7 @@ WORKDIR /app
 # Copy binary, migrations, and bundled skills
 COPY --from=builder /out/goclaw /app/goclaw
 COPY --from=builder /out/pkg-helper /app/pkg-helper
-COPY --from=builder /src/migrations/ /app/migrations/
+COPY --from=builder --chown=goclaw:goclaw /src/migrations/ /app/migrations/
 COPY --from=builder /src/skills/ /app/bundled-skills/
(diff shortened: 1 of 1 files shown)


viettranx:
Closing — this bug doesn't exist. Migration files are COPY'd from builder as root with default 644 permission (rw-r--r--), so the goclaw user can read them fine. The `goclaw upgrade` command only needs read access to /app/migrations/*.sql.

Verified:
- Docker COPY default perms = 644 (world-readable)
- goclaw user only needs read access to migration SQL files
- No reports of "permission denied" on migrations in production

The existing chown block at Dockerfile:140 intentionally doesn't include /app/migrations/ because read-only root ownership is the correct security posture for migration files (immutable, should not be modified at runtime).

PR #747, opened 2026-04-07
test(http): add regression coverage for inline auth tenant context

### LINKED ISSUE #734: fix(http): inline-auth handlers missing tenant ID context — agent lookup always fails
## Bug

Three HTTP handlers (`/v1/chat/completions`, `/v1/responses`, `/v1/agents/{id}/wake`) perform authentication inline instead of using the `requireAuth` middleware. They authenticate correctly but **skip the tenant ID injection step**, causing all downstream agent lookups to fail with `"agent not found"`.

## Root Cause

`requireAuth` middleware in `auth.go:290-315` injects `store.WithTenantID()` into the request context after authentication. The three affected handlers bypass this middleware and do auth inline via `resolveAuth(r)`, but never call `store.WithTenantID()`.

When the agent resolver calls `store.GetByKey()`, it reads `TenantIDFromContext(ctx)` which returns `uuid.Nil` → the SQL query filters by `tenant_id = uuid.Nil` → no rows match → `"agent not found"`.

## Affected Files

- `internal/http/chat_completions.go` — `ServeHTTP()` line ~137
- `internal/http/responses.go` — `ServeHTTP()` line ~81
- `internal/http/wake.go` — `handleWake()` line ~95

## Reproduction

```bash
curl -X POST http://localhost:18790/v1/chat/completions \
  -H "Authorization: Bearer <gateway_token>" \
  -H "X-GoClaw-User-Id: <user>" \
  -H "Content-Type: application/json" \
  -d '{"model": "agent:<agent_key>", "messages": [{"role": "user", "content": "Hello"}]}'
# Returns: {"error":{"message":"agent not found: <agent_key>"}}
```

## Fix

Inject tenant ID into context right after auth check, before agent lookup, in all three handlers:

```go
if auth.TenantID != uuid.Nil {
    r = r.WithContext(store.WithTenantID(r.Context(), auth.TenantID))
} else {
    r = r.WithContext(store.WithTenantID(r.Context(), store.MasterTenantID))
}
```

Fix already implemented and tested locally on branch `fix/memory-tenant-id-ambiguous`.

kaitranntt:
## Problem
Issue #734 described an HTTP inline-auth tenant-context bug, but current `dev` already contains the runtime fix from commit `781ed2a2f`.

What was still missing was regression coverage for that fix surface.

## What changed
- added regression coverage for tenant-context propagation in:
  - `/v1/chat/completions`
  - `/v1/responses`
  - `/v1/agents/{id}/wake`
  - `/v1/tools/invoke`
- added a regression test for wake owner-bound `user_id` override protection
- added a warm-cache tenant-scoped wake-path test so coverage is not limited to cold resolver misses

## Validation
- [x] `go test ./internal/http -run 'TestInlineAuthHandlersInjectTenantContext|TestToolsInvokeHandlerInjectsTenantContext|TestWakeHandlerBlocksBodyUserOverrideForBoundOwner|TestWakeHandlerUsesTenantScopedCacheEntry' -count=1`
- [x] `go test ./internal/http -count=1`
- [x] `go test -race ./internal/http`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./tests/...`
- [ ] `go test -race ./tests/integration/...` (`tests/integration/` does not exist in this repo layout)

## Notes
- This PR is test-only; it does not change runtime behavior.
- Invalid/unresolved tenant-hint fallback to master tenant and HTTP auth docs/runtime drift remain broader concerns and should be handled separately.

Closes #734

### DIFF
--- a/internal/http/inline-auth-tenant-context_test.go
+++ b/internal/http/inline-auth-tenant-context_test.go
@@ -1,0 +1,296 @@
+package http
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/google/uuid"
+
+	"github.com/nextlevelbuilder/goclaw/internal/agent"
+	"github.com/nextlevelbuilder/goclaw/internal/crypto"
+	"github.com/nextlevelbuilder/goclaw/internal/providers"
+	"github.com/nextlevelbuilder/goclaw/internal/store"
+	"github.com/nextlevelbuilder/goclaw/internal/tools"
+)
+
+type agentLookupCapture struct {
+	called   bool
+	agentKey string
+	tenantID uuid.UUID
+}
+
+func newCaptureRouter(capture *agentLookupCapture) *agent.Router {
+	router := agent.NewRouter()
+	router.SetResolver(func(ctx context.Context, agentKey string) (agent.Agent, error) {
+		capture.called = true
+		capture.agentKey = agentKey
+		capture.tenantID = store.TenantIDFromContext(ctx)
+		return nil, fmt.Errorf("agent not found: %s", agentKey)
+	})
+	return router
+}
+
+type toolContextCapture struct {
+	called   bool
+	tenantID uuid.UUID
+}
+
+type captureTool struct {
+	capture *toolContextCapture
+}
+
+func (t *captureTool) Name() string {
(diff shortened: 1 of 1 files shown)


kaitranntt:
Closing this PR for cleanup. The branch will remain on the fork for reference, but the PR is no longer needed.

PR #802, opened 2026-04-09
fix(media): add claude-cli as fallback provider for read_image and read_document

### LINKED ISSUE #801: read_image and read_document fail when only claude-cli provider is configured
## Problem

`read_image` and `read_document` tools fail with **"No vision provider configured"** when the only available provider is `claude-cli` (no separate Anthropic API key, no Gemini/OpenRouter configured).

## Root Cause

The media tool fallback chains (`visionProviderPriority`, `documentProviderPriority`) do not include `claude-cli`. They only list `openrouter`, `gemini`, `anthropic`, and `dashscope`.

`claude-cli` is registered under that name in the provider registry, not as `anthropic`, so `registry.Get(ctx, "anthropic")` returns an error and the chain comes up empty.

## Expected Behavior

Since Claude CLI natively supports vision (via stream-json input with images), it should be included in the fallback chain so media tools work out of the box on claude-cli-only installations.

## Environment

- GoClaw with only `claude-cli` provider configured
- No `providers.anthropic.api_key` set
- `read_image` called with a valid image path

henkedk:
## Problem

`read_image` and `read_document` fail with "No vision provider configured" when only `claude-cli` is available, because the media provider fallback chains don't include it.

Closes #801

## Root Cause

`claude-cli` is registered under that name in the provider registry, but `visionProviderPriority` and `documentProviderPriority` only list `anthropic` (which requires a separate API key). `registry.Get(ctx, "anthropic")` fails, and the chain comes up empty.

## Fix

- **`internal/tools/read_image.go`** — Add `claude-cli` to `visionProviderPriority` (after `anthropic`) and `visionModelDefaults`. Add `disable_tools: true` to Chat() options to skip MCP loading for one-shot vision calls.
- **`internal/tools/read_document.go`** — Add `claude-cli` to `documentProviderPriority` and `documentModelDefaults`.
- **`internal/tools/read_document_resolve.go`** — Add `disable_tools: true` to Chat() options.

## Testing

Verified on a claude-cli-only installation. `read_image` successfully analyzes images via Claude CLI's native stream-json vision support.

### DIFF
--- a/internal/tools/read_document.go
+++ b/internal/tools/read_document.go
@@ -51,12 +51,13 @@ const documentMaxBytes = 20 * 1024 * 1024
 // documentProviderPriority is the order in which providers are tried for document analysis.
 // Gemini has best native PDF support (50MB, 258 tokens/page).
 // "alibaba" is included as an alias for dashscope (common DB registration name).
-var documentProviderPriority = []string{"gemini", "anthropic", "openrouter", "dashscope"}
+var documentProviderPriority = []string{"gemini", "anthropic", "claude-cli", "openrouter", "dashscope"}
 
 // documentModelDefaults maps provider names to preferred document-capable models.
 var documentModelDefaults = map[string]string{
 	"gemini":     "gemini-2.5-flash",
 	"openrouter": "google/gemini-2.5-flash",
+	"claude-cli": "",
 	"dashscope":  "qwen-vl-max",
 }
 
--- a/internal/tools/read_document_resolve.go
+++ b/internal/tools/read_document_resolve.go
@@ -95,8 +95,9 @@ func (t *ReadDocumentTool) callProvider(ctx context.Context, cp credentialProvid
 		},
 		Model: model,
 		Options: map[string]any{
-			"max_tokens":  16384,
-			"temperature": 0.2,
+			"max_tokens":    16384,
+			"temperature":   0.2,
+			"disable_tools": true,
 		},
 	})
 	if err != nil {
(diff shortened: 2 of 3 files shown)


viettranx:
Thanks for the fix @henkedk! Closing in favor of 934ab7e4 on `dev`, which supersets this PR and addresses a deeper root cause surfaced while reviewing it.

**What the follow-up adds on top of this PR:**

1. **Root cause in `buildStreamJSONInput`** — the CLI bridge hardcoded `type: "image"` for every content block. Adding `claude-cli` to `documentProviderPriority` alone would still fail on real PDFs because the Anthropic API rejects image blocks with `media_type: application/pdf`. Fixed by routing `application/pdf` to a `type: "document"` block (Claude 3.5+ native PDF support). Without this, `read_document` on a claude-cli-only install would still error out.

2. **`disable_tools` scoping** — your PR put `disable_tools: true` into the shared `Options` map passed to every provider in the chain. Other providers silently ignore unknown keys today, but it couples a CLI-bridge-specific flag to a universal call site. The follow-up scopes it with `if providerName == "claude-cli"` so only the CLI bridge sees it.

3. **Tests** — added a table test for `buildStreamJSONInput` covering png / pdf / mixed / unknown MIME routing plus the empty-text edge case, so this doesn't regress.

Your `read_image` fix was spot-on and verified in practice — that path was correct. The `read_document` path needed the deeper CLI-bridge change to actually deliver on its promise for PDFs. Appreciate you flagging #801 and getting the ball rolling on this!

PR #805, opened 2026-04-10
feat: add delete-all chats action to web sidebar

TheSethRose:
## Summary
- add a `sessions.delete-bulk` RPC for deleting the currently visible web chat sessions while preserving per-session delete side effects
- add a delete-all action to the web chat sidebar with confirmation, route reset behavior, and localized success/error copy
- fix the existing timer typing issues that were blocking the embedded web UI build

## Verification
- `pnpm --dir ui/web build`
- `make build-full`
- `go build ./...`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,5 @@ compose.d/*
 /mise.*.toml
 /mise
 /.mise
+
+.agents/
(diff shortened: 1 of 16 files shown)


TheSethRose:
image

PR #821, opened 2026-04-10
feat(web-search): add ranked native Exa and Tavily providers

kaitranntt:
## Problem

GoClaw `web_search` only exposed Brave Search and DuckDuckGo, while the native runtime had no way to rank optional higher-quality backends like Exa and Tavily. The config page also could not express that provider stack, and once new provider secrets were stored via the dashboard they did not rebuild the live tool.

This draft PR adds the ranked native provider stack and the matching config UI, but it stays a draft because I still want to keep the contribution in the normal maintainer-style visual-review flow rather than marking it ready immediately.

## Visual Review

[Full comparison report](https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaitranntt/goclaw/d949ac4eddf6be40e96e8a9a16b5f57919cc2c80/.github/pr-assets/821/index.html)

### 1. Web Search card becomes an explicit provider-order control

| Before | After |
| --- | --- |
| Before | After |

The old UI only exposes DuckDuckGo and Brave as a flat pair of controls. The new UI turns that area into a ranked provider-order surface with a visible chain and DuckDuckGo pinned as the last fallback.

### 2. Detailed provider rows expose the expanded inline workflow

| Before | After |
| --- | --- |
| Before | After |

Before, the detailed controls only cover DuckDuckGo and Brave. After, the UI supports the full provider chain with inline max-result and API-key fields. The highlighted after state was captured with local demo values entered and switches toggled ON to show the enabled-provider layout; those demo values were not part of the committed config.

All captures use the same live backend on `100.120.60.40:18790`, light theme, and the same authenticated dashboard session. Red callout borders mark the reviewed UI surface.

## Root Cause

The original gap had three parts:

1. `web_search` only built a fixed Brave → DuckDuckGo chain, so optional native providers could not participate.
2. config secrets for search providers were persisted, but `web_search` did not rebuild from DB-backed secrets on startup or after config changes.
3. the config page had no UI for ranked search-provider configuration, and blank/masked secret states needed explicit handling to distinguish preserve-vs-delete behavior.

## What changed

| File | Change |
| --- | --- |
| `internal/tools/web_search{,_exa,_tavily,_brave,_ddg}.go` | Add Exa/Tavily providers, provider-order normalization, live config updates, and provider-specific result caps |
| `cmd/gateway.go` | Re-sync `web_search` after DB-backed config secrets are applied |
| `cmd/gateway_setup.go` | Build the native web-search config once and expose runtime re-registration helpers |
| `cmd/gateway_lifecycle.go` | Reload the live `web_search` tool on `config:changed` events |
| `internal/config/config_channels.go` | Add `tools.web.provider_order`, Exa config, and Tavily config |
| `internal/config/config_secrets.go` | Treat Exa/Tavily keys like Brave for masking, extraction, and DB overlay |
| `internal/gateway/methods/config.go` | Preserve masked secrets, delete cleared ones, and keep config-save semantics correct |
| `ui/web/src/pages/config/sections/tools-web-section.tsx` | Add provider-order UI plus Exa/Tavily toggles, max-results inputs, and key fields |
| `ui/web/src/i18n/locales/*` + `internal/i18n/catalog_*.go` | Update user-facing copy to match the expanded provider stack |
| `docs/03-tools-system.md`, `docs/17-changelog.md` | Document the native ranked provider stack and its fallback behavior |
| `*_test.go`, `tests/integration/*` | Add regression coverage for runtime reload, secret deletion/preservation, provider ordering, and fix two integration-test drifts surfaced by the review checklist |

## Validation

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `go test ./internal/tools ./internal/gateway/methods ./cmd/...`
- [x] `go test -race -tags integration ./tests/integration`
- [x] `pnpm --dir ui/web build`
- [x] Maintainer-style Before/After UI evidence captured and committed under `.github/pr-assets/821/`

## Notes

- This PR targets `dev`.
- It remains a **draft** intentionally and is **not ready for review** yet.


kaitranntt:
Superseded by #825. I rebuilt the branch onto a clean commit stack and moved the same feature/report there so the polluted self-reference timeline on #821 is no longer part of the review surface.

PR #833, opened 2026-04-11
refactor: update default timezone to UTC and apply code style formatt…

reski-rukmantiyo:
## Summary
feat: add default configuration, and update timezone settings

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
- dev

## Checklist
- [x] `go build ./...` passes
- [ ] `go vet ./...` passes

### DIFF
--- a/internal/agent/loop_history.go
+++ b/internal/agent/loop_history.go
@@ -96,7 +96,7 @@ func (l *Loop) buildMessages(ctx context.Context, history []providers.Message, s
 	// DM only — group chats have permission checks and multiple senders.
 	if hadBootstrap && peerKind == "direct" {
 		if senderName := store.SenderNameFromContext(ctx); senderName != "" {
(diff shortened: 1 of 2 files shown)


viettranx:
Closing in favor of a better approach.

**Problem with UTC default:** Setting UTC would cause confusion for all users - Vietnamese users would have times interpreted wrong, and international users still need to specify their timezone anyway.

**Better solution (implemented in dev):** Remove hardcoded default timezone entirely. Instead, hint the model to ask for timezone on-demand when user mentions times/schedules:

```go
hint := fmt.Sprintf("Known user info (from %s): Name=%q\nTimezone: not yet known. When the user mentions times, schedules, or reminders, ask for their timezone and update USER.md.", channelType, senderName)
```

**Benefits:**
- No wrong assumptions for any user
- Only asks when actually needed (scheduling, reminders)
- Persistent after first ask (stored in USER.md)
- Consistent with datetime/cron tools that already require explicit timezone

PR #834, opened 2026-04-11
fix(store): coerce null emoji/agent_description on agent update

tuannt23065:
## Summary
- Agent update fails with `null value in column "emoji" violates not-null constraint` when UI sends `emoji: null` or `agent_description: null`
- Adds null → `""` coercion for NOT NULL text columns in both PG and SQLite agent stores, matching existing pattern for `skill_nudge_interval` (null → 0)

## Test plan
- [ ] Update agent via Web UI without setting emoji → should succeed (previously 500)
- [ ] Update agent with explicit emoji → still works as before
- [ ] PG and SQLite builds compile cleanly

### DIFF
--- a/internal/store/pg/agents.go
+++ b/internal/store/pg/agents.go
@@ -193,6 +193,12 @@ func (s *PGAgentStore) Update(ctx context.Context, id uuid.UUID, updates map[str
 	if v, ok := updates["skill_nudge_interval"]; ok && v == nil {
 		updates["skill_nudge_interval"] = 0
 	}
+	// NOT NULL text columns: null → empty string.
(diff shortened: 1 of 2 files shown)


viettranx:
Đã fix đầy đủ hơn ở commit fbfae1e6 trên dev. Fix đó bao gồm:
- TEXT columns: emoji, agent_description, thinking_level → ""
- INT columns: skill_nudge_interval + max_tokens → 0
- Frontend: gửi "" thay vì null

PR này là subset của fix đó nên không cần merge nữa. Cảm ơn bạn đã báo bug!

PR #839, opened 2026-04-11
fix(pipeline): block_reply dedup falsely suppresses final message

### LINKED ISSUE #838: v3 pipeline: block_reply dedup falsely suppresses final message when no tools are called
GoClaw v3.3.0

  ### Description

  When the v3 pipeline handles a simple LLM response (no tool calls), the bot receives the message and completes the trace successfully, but **never sends the reply to Discord**. The "Thinking..." placeholder appears then disappears with no response.

  This affects all non-streaming channels (Discord groups, Zalo, WhatsApp) with `block_reply: true` (the gateway default).

  ### Root Cause

  A mismatch between `ObserveStage` and `ThinkStage` causes the gateway dedup logic to falsely suppress the final message.

  **`ObserveStage`** increments `BlockReplies` for *every* LLM response with content, including the final answer:

  ```go
  // observe_stage.go:34
  if resp.Content != "" {
      state.Observe.BlockReplies++
      state.Observe.LastBlockReply = resp.Content
  }

  ThinkStage only calls EmitBlockReply during tool iterations — the final answer returns early before the emit:

  // think_stage.go:102-104
  if len(resp.ToolCalls) == 0 {
      s.result = BreakLoop
      return nil  // returns before EmitBlockReply at line 120
  }
  // ... (line 120)
  s.deps.EmitBlockReply(resp.Content)  // only reached when tool calls exist

  Gateway dedup trusts BlockReplies > 0 to mean content was already delivered via block.reply events:

  // gateway_consumer_normal.go:442
  if blockReplyEnabled && outcome.Result.BlockReplies > 0 &&
     outcome.Result.Content == outcome.Result.LastBlockReply &&
     len(outcome.Result.Media) == 0 {
      // sends empty content → Discord deletes placeholder → nothing replaces it
  }

  Reproduction

  1. Enable v3 pipeline + block_reply: true (gateway default)
  2. Send a simple message to the bot on a non-streaming channel (Discord group, Zalo, WhatsApp)
  3. LLM responds with text only (no tool calls)
  4. "Thinking..." appears then disappears — no reply

  Works correctly when:
  - LLM calls tools (1+ iterations) — EmitBlockReply fires, dedup is valid
  - block_reply: false — dedup path is skipped entirely
  - Sanitization modifies content — Content != LastBlockReply, dedup doesn't match

  Fix

  ObserveStage should only count BlockReplies when tool calls are present (i.e., when EmitBlockReply actually fires):

  - if resp.Content != "" {
  + if resp.Content != "" && len(resp.ToolCalls) > 0 {
        state.Observe.BlockReplies++
        state.Observe.LastBlockReply = resp.Content
    }

  Affected Files

  ┌──────────────────────────────────────────┬────────────────────────────────────────────┐
  │                   File                   │                    Role                    │
  ├──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ internal/pipeline/observe_stage.go:34    │ Bug — counts BlockReplies for final answer │
  ├──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ internal/pipeline/think_stage.go:102-121 │ EmitBlockReply only fires with tool calls  │
  ├──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ cmd/gateway_consumer_normal.go:442       │ Dedup relies on BlockReplies count         │
  └──────────────────────────────────────────┴────────────────────────────────────────────┘

  Severity

  High — bot is completely silent on non-streaming channels for any simple response without tool calls. Workaround: set "block_reply": false in gateway config.
  ```

namnhfreelancer:
Fixes #838. See issue for full details.

### DIFF
--- a/internal/pipeline/observe_stage.go
+++ b/internal/pipeline/observe_stage.go
@@ -30,8 +30,11 @@ func (s *ObserveStage) Execute(_ context.Context, state *RunState) error {
 		return nil
 	}
 
-	// 2. Track block replies (every response with content counts as a block)
-	if resp.Content != "" {
+	// 2. Track block replies — only count tool-iteration responses where
+	// EmitBlockReply actually fires (think_stage emits block.reply only when
+	// tool calls are present). The final answer (no tool calls) must NOT be
+	// counted, otherwise the gateway dedup check falsely suppresses delivery.
+	if resp.Content != "" && len(resp.ToolCalls) > 0 {
 		state.Observe.BlockReplies++
 		state.Observe.LastBlockReply = resp.Content
 	}


viettranx:
Thanks @namnhfreelancer for the detailed issue analysis and fix! 🙏

I've incorporated your fix directly into `dev` with the following additions:
- Updated the existing test to reflect the new behavior (tool calls required for BlockReplies)
- Added a regression test specifically for the #838 scenario

Commit: 034187ac

Closing this PR as the fix is now on `dev`. Great catch on this subtle interaction between ObserveStage and ThinkStage!

PR #885, opened 2026-04-13
fix: capture claude-cli stream errors and surface real error to user

### LINKED ISSUE #884: claude-cli stream errors discarded — users see generic 'something went wrong' instead of real error
## Bug

When claude-cli exits with a non-zero status due to an API error (rate limit, usage limit, overloaded, etc.), users see a generic "something went wrong" message instead of the actual error.

## Root Cause

Claude-cli reports API errors via its stdout stream-json protocol, not stderr. It emits an error result event like:

```json
{"type":"result","subtype":"error","result":"","is_error":true,"error":"Maximum usage limit reached"}
```

However:

1. **`cliStreamEvent` in `internal/providers/claude_cli_types.go`** doesn't map `is_error` or `error` fields — they're silently dropped by `json.Unmarshal`
2. **`ChatStream()` in `internal/providers/claude_cli_chat.go`** discards the error signal when `finalResp.Content` is empty, falling through to the stderr path which is genuinely empty
3. Error surfaces as `claude-cli: exit status 1 (stderr: )` — useless
4. Downstream in `cmd/gateway_errors.go`, this becomes an "unclassified agent error" → generic user message

## Impact

- Users get no actionable feedback on API failures
- Operators can't diagnose failures from logs (empty stderr)
- Rate limit / usage limit errors are indistinguishable from crashes

## Expected Behavior

Real error message should surface both in logs and to the user, e.g. "⚠️ API rate limit reached. Please try again later."

## Versions Affected

Confirmed on v3.6.0. Likely present since claude-cli provider was introduced.

henkedk:
Fixes #884

Claude-cli reports API errors via stdout stream-json, not stderr. The error events were silently discarded — users saw generic 'something went wrong' instead of the real error.

Fix: parse is_error/error fields from stream events, use captured stream error when stderr is empty, classify 'usage limit' as rate-limit for proper user messaging.


viettranx:

PR #887, opened 2026-04-13
fix: whitelist effort/thinking env vars in filterCLIEnv + wire per-agent --effort

### LINKED ISSUE #886: filterCLIEnv strips CLAUDE_CODE_EFFORT_LEVEL and CLAUDE_CODE_DISABLE_ADAPTIVE_THINKING — no per-agent effort control
## Bug

`filterCLIEnv` in `internal/providers/claude_cli_session.go` strips all `CLAUDE_*` env vars (except `CLAUDE_CODE_OAUTH_TOKEN`) before launching claude-cli subprocesses. This was a deliberate defensive measure to prevent session identity vars from leaking into child processes.

However, it also strips behavioral tuning vars:
- `CLAUDE_CODE_EFFORT_LEVEL` — controls adaptive reasoning depth (low/medium/high/max)
- `CLAUDE_CODE_DISABLE_ADAPTIVE_THINKING` — disables dynamic thinking allocation

These are not session identity vars — they're performance/cost tuning knobs. Stripping them means:
1. No way to control thinking depth per agent
2. All agents run at the same global default regardless of task complexity
3. Significant token waste on simple tasks (e.g. Scout doing lookups at full thinking depth)

Additionally, `ClaudeCLIProvider` doesn't wire the existing reasoning config (`OptThinkingLevel`) into `--effort` CLI flags, so the DB-level per-agent effort config is dead code for claude-cli agents.

## Expected Behavior

- Behavioral tuning env vars should pass through to claude-cli
- Per-agent effort level from DB config should map to `--effort` CLI flag
- When explicit `--effort` is set, `CLAUDE_CODE_EFFORT_LEVEL` env should be stripped to avoid precedence ambiguity

## Versions Affected

Present since claude-cli provider was introduced (the original `filterCLIEnv` stripped everything).

henkedk:
Fixes #886

Two commits:

**1. Whitelist behavioral tuning env vars + wire --effort (4bee2a07)**
- `filterCLIEnv` now uses explicit allowlist: `CLAUDE_CODE_OAUTH_TOKEN`, `CLAUDE_CODE_EFFORT_LEVEL`, `CLAUDE_CODE_DISABLE_ADAPTIVE_THINKING`. All other `CLAUDE*` vars still stripped.
- `buildArgs` accepts effort param, appends `--effort <level>` when set.
- `Chat` and `ChatStream` extract `OptThinkingLevel` from request options and pass through.

**2. Sanitize effort + resolve env/flag precedence (69506609)**
- Effort value normalized: lowercase, trimmed, alpha-only validation (no hardcoded allowlist — future values like `max` pass through).
- Invalid values logged as warning and skipped.
- When explicit `--effort` is set, `CLAUDE_CODE_EFFORT_LEVEL` is stripped from subprocess env to avoid ambiguity.

Reviewed by Lars (Codex). Approved with the sanitization follow-up which is included in this PR.


viettranx:

PR #889, opened 2026-04-13
fix: return error on Codex response.failed instead of treating as success

### LINKED ISSUE #888: Codex provider silently swallows response.failed events — empty results, no retry
## Bug

When the Codex backend fails during generation (server-side failure within an HTTP 200 SSE stream), the agent returns an empty/garbage response ("...") instead of surfacing the error or retrying.

## Root Cause

`internal/providers/codex.go` line ~250: `response.failed` is handled in the same switch case as `response.completed`. `processSSEEvent` had no return value, so there was no way to signal a stream-level failure back to the caller.

Result:
1. Codex sends `response.failed` event (with error details)
2. GoClaw treats it as a successful completion with empty content
3. No error returned → retry logic never triggers
4. `finalize_stage.go` delivers "..." as the user-facing response
5. 61k+ input tokens billed with no useful output

On manual retry the backend typically succeeds — hence the intermittent "flaky" pattern.

## Impact

- Agents on Codex (e.g. code reviewers) intermittently return empty results
- Wasted API tokens on silent failures
- No error surfaced to user or logs — indistinguishable from a legitimate empty response

## Fix

Separate `response.failed` from the success case in `processSSEEvent`, return an error with the failure message/code, and let the existing error handling pipeline surface it.

## Versions Affected

Present since the Codex provider was introduced.

henkedk:
Fixes #888

`processSSEEvent` had no return value, so `response.failed` was handled in the same switch case as `response.completed`. Server-side Codex failures during generation were silently swallowed — no error, no retry, user got "...".

Changes:
- Added `codexErrorDetail` struct to `codex_types.go` (Code, Message fields)
- `processSSEEvent` now returns `error`
- `response.failed` separated from success case, returns error with message/code from the failed response
- SSE loop checks returned error and propagates through `ChatStream` → think stage → `formatAgentError`

Reviewed by Lars (Codex). Approved — correct error propagation, no false-positive risk on legitimate empty responses.


viettranx:

PR #911, opened 2026-04-15
fix(channels): suppress error messages for external channels

khanhtranhd:
## Summary
- Suppress internal error messages (e.g. "something went wrong") for external channels (Facebook, Telegram, Discord, Zalo, WhatsApp, Feishu, LINE)
- Errors are logged server-side but end users no longer see technical error details
- Applies to both normal message flow and subagent announce flow

## Test plan
- [ ] Trigger agent error on Facebook Messenger → verify no error message sent to user
- [ ] Trigger agent error on WS/internal channel → verify error message still shown
- [ ] Build passes: `go build ./...` and `go vet ./cmd/...`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ compose.d/*
 /mise.*.toml
 /mise
 /.mise
+goclaw-patched-linux-amd64
--- a/cmd/gateway_consumer_normal.go
+++ b/cmd/gateway_consumer_normal.go
@@ -220,6 +220,7 @@ func processNormalMessage(
 			outMeta["reply_to_message_id"] = mid
 		}
 	}
+	// Channel routing keys — keep in sync with routingMetaKeys in channels/events.go.
 	for _, k := range []string{
 		tools.MetaMessageThreadID, "local_key", "placeholder_key", "group_id",
 		"feishu_reply_target_id",
--- a/cmd/gateway_consumer_normal.go
+++ b/cmd/gateway_consumer_normal.go
@@ -415,10 +415,19 @@ func processNormalMessage(
 				return
 			}
 			slog.Error("inbound: agent run failed", "error", outcome.Err, "channel", channel)
+			// Suppress technical error messages for external channels (Facebook, Telegram, etc.)
+			// — end users should not see internal error details.
+			errContent := formatAgentError(outcome.Err)
+			if deps.ChannelMgr != nil {
+				if ct := deps.ChannelMgr.ChannelTypeForName(channel); isExternalChannel(ct) {
+					slog.Info("inbound: suppressed error for external channel", "channel", channel, "type", ct)
+					errContent = ""
+				}
+			}
 			deps.MsgBus.PublishOutbound(bus.OutboundMessage{
 				Channel:  channel,
 				ChatID:   chatID,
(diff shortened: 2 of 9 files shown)


viettranx:
Closing in favor of a reworked fix landed directly on `dev` as commit `5040fe4c` — thanks @khanhtranhd for the initial work and bug report.

### Why this PR wasn't merged as-is

The PR was targeted at `main` but the project convention is `dev → main`, so bypassing `dev` would have skipped ~30 unmerged commits. More importantly, review against `dev` surfaced two correctness issues with the whitelist approach:

**1. `isExternalChannel` whitelist used string literals that don't match real channel type constants.**

Actual constants in `internal/channels/channel.go`:

| PR literal | Real constant(s) | Effect |
|---|---|---|
| `"zalo"` | `TypeZaloOA` = `"zalo_oa"`, `TypeZaloPersonal` = `"zalo_personal"` | Zalo OA + Personal still leaked errors |
| `"line"` | — (only a Pancake sub-platform, never a top-level type) | Dead branch |
| *(missing)* | `TypePancake` = `"pancake"` | Pancake still leaked errors |
| *(missing)* | `TypeSlack` = `"slack"` | Slack still leaked errors |

So the PR would have achieved its stated goal for 4 of 9 public-facing channels, not all of them.

**2. Empty-content regression on Facebook (and Pancake).**

Suppression publishes an outbound with `Content = ""`. Downstream `dispatch.go` only skips empty messages when the media-filter branch runs, so for normal agent-error flow (no media) `channel.Send` gets called with empty content.

- Telegram, Discord, WhatsApp, Slack, Zalo Personal, Feishu already short-circuit on empty content.
- **Facebook** did not — `splitMessage("", 640)` returns `[""]` (slice with one empty string), loop then calls `graphClient.SendMessage(chatID, "")` → Graph API returns 400 → `handleAPIError` marks the channel degraded. Comment mode hits the same issue via `ReplyToComment(commentID, "")`.
- Pancake had the same empty-POST path.

### What landed on `dev`

Commit `5040fe4c` — `fix(channels): suppress agent-error text on public-facing channels`:

- `cmd/gateway_errors.go`: `isExternalChannel()` now uses `channels.Type*` constants (compile-time safe), covers all 9 real channel types (FB, Telegram, Discord, Feishu, WhatsApp, Zalo OA, Zalo Personal, Pancake, Slack).
- `cmd/gateway_consumer_normal.go` + `cmd/gateway_subagent_announce_queue.go`: same suppression logic as this PR.
- `internal/channels/facebook/facebook.go` + `internal/channels/pancake/pancake.go`: added `if msg.Content == "" && len(msg.Media) == 0 { return nil }` guard at the top of `Send()`, matching the pattern already used by Telegram/Discord/Slack.
- `cmd/gateway_errors_test.go` (new): table-driven test for `isExternalChannel` covering every real type + empty + unknown + the legacy `"zalo"` short form (as a regression guard).

Build + vet + tests all pass (PG and `-tags sqliteonly`).

The admin-reply / bot-echo tracking work and routing meta key additions from your earlier commits are already on `dev` — nothing lost. The `.gitignore` change for `goclaw-patched-linux-amd64` was skipped as it's a per-developer artifact.

Closing. 🙏

PR #924, opened 2026-04-16
fix(zalo): correct json field mappings

### LINKED ISSUE #922: fix(zalo): OA bot silently drops all messages — wire type and deserialization mismatch
## Summary

The Zalo OA bot channel starts successfully but **silently drops every incoming message**. Root cause: the code was ported from OpenClaw TS assuming Telegram-like array responses, but the Zalo Bot API (`bot-api.zaloplatforms.com`) returns a single update object per `getUpdates` call.

### P0 — `getUpdates` deserializes as array, API returns single object

`internal/channels/zalo/zalo.go:486-500`

```go
// Current (broken) — expects array
var updates []zaloUpdate
json.Unmarshal(result, &updates) // fails: cannot unmarshal object into []zaloUpdate
```

The Zalo Bot API returns:
```json
{"ok": true, "result": {"event_name": "message.text.received", "message": {...}}}
```

GoClaw expects:
```json
{"ok": true, "result": [{"event_name": "message.text.received", "message": {...}}]}
```

Every `getUpdates` call with a pending message triggers an unmarshal error → 5s backoff → retry → same failure forever. Messages are permanently lost.

**Verified against:** [`go-zalo-bot-api`](https://github.com/nduyhai/go-zalo-bot-api) reference library — `BaseResp[UpdateResult]` is a single object, not `BaseResp[[]UpdateResult]`.

### P1 — Wire type JSON field names mismatch

| GoClaw struct field | GoClaw JSON tag | Actual API field | Impact |
|---------------------|-----------------|------------------|--------|
| `zaloBotInfo.Name` | `"name"` | `"display_name"` | Bot name always empty in logs |
| `zaloFrom.Username` | `"username"` | `"display_name"` | Sender name never captured |
| `zaloChat.Type` | `"type"` | `"chat_type"` | Chat type always empty |

Verified via live `getMe` response:
```json
{"id":"...","account_name":"bot.hkMESnCs","account_type":"BASIC","display_name":"Bot An Nhiên Safety"}
```

### P2 — HTTP client timeout too tight for long-polling

`Timeout: 60s` global with 30s server poll timeout. Reference library uses `Timeout: 0` with per-call context timeouts + 7s headroom.

## Fix Plan

1. Change `getUpdates` to deserialize single `*zaloUpdate` object
2. Align `zaloBotInfo`, `zaloFrom`, `zaloChat` JSON tags to match actual API
3. Add dedicated poll HTTP client with per-call context timeout
4. Update all tests

**Scope:** `internal/channels/zalo/` — no API changes, no migrations, no UI changes.

utafrali:
## Summary
Fixes #922 - Zalo OA bot was silently dropping all incoming messages due to JSON field mapping mismatches.

## Type
- [x] Bug fix

## Target Branch
- [x] dev

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass with race detector
- [x] No hardcoded secrets
- [x] SQL uses parameterized queries (N/A)
- [x] i18n keys added for new user-facing strings (N/A)
- [x] Migrations updated (PG + SQLite) if schema changed (N/A)

## Test Plan
Tested with Zalo sandbox API - messages now come through properly with sender names and chat IDs.

---

The Zalo struct tags didn't match the actual API response. When messages came in, the JSON unmarshal would fail silently looking for `name` (actually `display_name`), `username` (also `display_name`), and `type` (actually `chat_type`). This left sender and chat info empty, so the bot would drop messages.

Updated the tags to match what the API actually sends. Also split the HTTP client since long-polling needs to wait indefinitely for updates - can't have a timeout. Added a separate `pollClient` with no timeout. Fixed the getUpdates unmarshal too, turns out the API sends a single update object, not an array.

### DIFF
--- a/internal/channels/zalo/zalo.go
+++ b/internal/channels/zalo/zalo.go
@@ -44,6 +44,7 @@ type Channel struct {
 	blockReply *bool
 	stopCh     chan struct{}
 	client     *http.Client
+	pollClient *http.Client
 	// pairingService, pairingDebounce are inherited from channels.BaseChannel.
 }
 
@@ -74,6 +75,7 @@ func New(cfg config.ZaloConfig, msgBus *bus.MessageBus, pairingSvc store.Pairing
 		blockReply:  cfg.BlockReply,
 		stopCh:      make(chan struct{}),
 		client:      &http.Client{Timeout: 60 * time.Second},
+		pollClient:  &http.Client{Timeout: 0},
 	}
 	ch.SetPairingService(pairingSvc)
 	return ch, nil
@@ -398,7 +400,7 @@ type zaloAPIResponse struct {
 
 type zaloBotInfo struct {
 	ID   string `json:"id"`
-	Name string `json:"name"`
+	Name string `json:"display_name"`
 }
 
 type zaloMessage struct {
@@ -414,12 +416,12 @@ type zaloMessage struct {
 
 type zaloFrom struct {
 	ID       string `json:"id"`
-	Username string `json:"username"`
+	Username string `json:"display_name"`
 }
 
 type zaloChat struct {
 	ID   string `json:"id"`
-	Type string `json:"type"`
+	Type string `json:"chat_type"`
 }
 
 type zaloUpdate struct {
@@ -428,6 +430,10 @@ type zaloUpdate struct {
 }
 
 func (c *Channel) callAPI(method string, body any) (json.RawMessage, error) {
+	return c.callAPIWith(context.Background(), c.client, method, body)
+}
+
+func (c *Channel) callAPIWith(ctx context.Context, client *http.Client, method string, body any) (json.RawMessage, error) {
 	url := fmt.Sprintf("%s/bot%s/%s", apiBase, c.token, method)
 
 	var reqBody io.Reader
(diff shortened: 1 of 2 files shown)


mrgoonie:
Review below

vanducng:
@utafrali Please fix this and make sure the CI passes.

mrgoonie:
## 🔍 Code Review — fix(zalo): correct json field mappings

### 🎯 Overview

Fix Zalo OA bot silently dropping messages due to JSON field mapping mismatches between struct tags and actual API response.

**Scope:** 2 files, ~30 lines changed

---

### ✅ What's Good

1. **Correct root cause** — Zalo API returns `display_name` instead of `name`/`username`, `chat_type` instead of `type` → silent unmarshal failure → bot drops all messages. Fix is in the right place.

2. **Separate poll client** — `pollClient` with timeout = 0 for long-polling, `client` with 60s for normal API calls. Correct pattern, avoids mid-wait timeouts.

3. **Context propagation** — `callAPIWith(ctx, client, ...)` instead of `http.NewRequest` → requests are cancelable on channel stop.

4. **getUpdates fix** — Zalo API returns a single object (`result: {...}`), not an array (`result: [...]`) → unmarshal into `zaloUpdate` then wrap as slice. Correct.

5. **Tests updated** — Mirror the actual API response format, test names are clearer (`ParsesSingleUpdate` instead of `ParsesUpdateArray`).

---

### ⚠️ Minor Issue

**🟢 LOW — `callAPI` falls back to `context.Background()`**

```go
func (c *Channel) callAPI(method string, body any) (json.RawMessage, error) {
    return c.callAPIWith(context.Background(), c.client, method, body)
}
```

Should pass context from caller instead of `Background()` for cancelability. But this is an internal helper with low impact — can refactor later.

---

### 💡 Recommendation

**🟢 APPROVE**

Clean bug fix, clear root cause, tests pass. Great work @utafrali! 🚀

PR #925, opened 2026-04-16
fix(zalo): OA bot wire type mismatch + poll client + context propagation

### LINKED ISSUE #922: fix(zalo): OA bot silently drops all messages — wire type and deserialization mismatch

vanducng:
## Summary
- Align wire types with actual Zalo Bot API (`display_name`, `chat_type`, `is_bot`, `account_name`, `account_type`)
- Fix `getUpdates` deserialization: single object instead of array
- Add dedicated `pollClient` with no global timeout for long-poll; per-call `context.WithTimeout` with 7s headroom
- Propagate `context.Context` through `sendMessage`, `sendPhoto`, `downloadMedia` for graceful shutdown

Resolves #922

## Test plan
- [x] `go build ./...` (PG)
- [x] `go build -tags sqliteonly ./...` (Desktop)
- [x] `go vet ./internal/channels/zalo/`
- [x] `go test -count=1 ./internal/channels/zalo/` — 35 tests pass
- [x] Regression tests: array format rejection, null result, context cancellation

### DIFF
--- a/internal/channels/zalo/zalo.go
+++ b/internal/channels/zalo/zalo.go
@@ -116,7 +116,7 @@ func (c *Channel) Stop(_ context.Context) error {
 }
 
 // Send delivers an outbound message to a Zalo chat.
-func (c *Channel) Send(_ context.Context, msg bus.OutboundMessage) error {
+func (c *Channel) Send(ctx context.Context, msg bus.OutboundMessage) error {
 	if !c.IsRunning() {
 		return fmt.Errorf("zalo bot not running")
 	}
@@ -132,13 +132,13 @@ func (c *Channel) Send(_ context.Context, msg bus.OutboundMessage) error {
 			if end > 0 {
 				photoURL := msg.Content[start+7 : start+end]
 				caption := strings.TrimSpace(msg.Content[:start] + msg.Content[start+end+1:])
-				return c.sendPhoto(msg.ChatID, photoURL, caption)
+				return c.sendPhoto(ctx, msg.ChatID, photoURL, caption)
 			}
 		}
 	}
 
 	// Send as text, chunking if over 2000 chars
-	return c.sendChunkedText(msg.ChatID, msg.Content)
+	return c.sendChunkedText(ctx, msg.ChatID, msg.Content)
 }
 
 // --- Polling ---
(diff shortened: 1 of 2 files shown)


vanducng:
Closed as duplicate of #924 which covers the same fix.

Issue #917, opened 2026-04-15
bug: vault_search excludes shared docs (agent_id=NULL) when agent searches


teexiii:
## Summary

Docs uploaded via UI **without selecting an agent** are permanently invisible to agent `vault_search`. This affects all shared-scope documents.

## Root Cause

When docs are uploaded without an agent selected, they are stored with `agent_id = NULL, scope = 'shared'`.

When an agent calls `vault_search`, the SQL filter adds a strict equality check:

```sql
AND agent_id = 'agent-uuid'   -- NULL != 'uuid' → shared docs excluded
```

SQL `NULL = 'uuid'` evaluates to `NULL` (not `TRUE`), so all shared docs are silently excluded.

## Affected Files

| File | Line | Issue |
|------|------|-------|
| `internal/store/pg/vault_documents.go` | 546, 585 | `AND agent_id = $N` excludes NULL rows in `ftsSearch` and `vectorSearch` |
| `internal/http/vault_handler_upload.go` | upload handler | Upload without agent selection → `scope=shared, agent_id=NULL` |

## Steps to Reproduce

1. Upload docs via Vault UI without selecting an agent
2. Create an agent and call `vault_search` with any query
3. Observe: docs are not returned despite being in the vault

## Expected Behavior

Agent should be able to find shared docs (`scope = 'shared'`, `agent_id = NULL`) in addition to their own personal docs.

## Actual Behavior

Only docs explicitly assigned to the agent (`agent_id = agent-uuid`) are returned. Shared docs with `agent_id = NULL` are never returned.

## Proposed Fix

Change the agent_id filter in `ftsSearch` and `vectorSearch` to include NULL (shared) docs:

```go
// internal/store/pg/vault_documents.go (line 546, 585)
// Before:
if agentID != nil {
    q += fmt.Sprintf(" AND agent_id = $%d", p)
    args = append(args, *agentID)
    p++
}

// After:
if agentID != nil {
    q += fmt.Sprintf(" AND (agent_id = $%d OR agent_id IS NULL)", p)
    args = append(args, *agentID)
    p++
}
```

## Impact

Any user who uploads docs to the vault without selecting a specific agent (the default behavior) cannot have those docs found by agent search. This silently breaks the expected knowledge base workflow.

minhhoang2705:
PR created: #928

viettranx:
[...]

PR #928, opened 2026-04-16
fix(vault): include shared docs in search results

minhhoang2705:
Fixes issue 917

### DIFF
--- a/internal/store/pg/vault_documents.go
+++ b/internal/store/pg/vault_documents.go
@@ -155,7 +155,7 @@ func (s *PGVaultStore) GetDocument(ctx context.Context, tenantID, agentID, path 
 		if err != nil {
 			return nil, fmt.Errorf("vault get document: agent: %w", err)
 		}
-		q += fmt.Sprintf(" AND agent_id = $%d", p)
+		q += fmt.Sprintf(" AND (agent_id = $%d OR agent_id IS NULL)", p)
 		args = append(args, aid)
 		p++
 	}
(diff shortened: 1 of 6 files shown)


viettranx:
Closing as duplicate — issue #917 was already fixed in `dev` by commit 7cfcbbf9 ("fix(vault): include shared docs in agent read paths (#917)") with broader scope.

Problems with this PR if merged on top of current `dev`:

1. **Duplicate fix** — 7cfcbbf9 already patches `Search`, `ListDocuments`, `CountDocuments`, `ListTreeEntries` and adds a CHECK invariant (migration 000055) to prevent future drift.
2. **Migration number collision** — this PR adds `000053_vault_docs_shared_index`, but `dev` already uses 000053/000054/000055 for agent_hooks + vault scope check.
3. **Schema version regression** — `RequiredSchemaVersion` is bumped **down** from 55 → 53, which would break boot on any DB already migrated to 55.
4. **Deletes 364 lines of integration tests** that were merged with 7cfcbbf9 (shared-doc visibility matrix, cross-tenant isolation, delete-strict guard).
5. **Weakens auth on `DeleteDocument`** — changing the agent filter to `(agent_id = $N OR agent_id IS NULL)` means an agent-scoped delete could wipe tenant-wide shared docs. `dev` intentionally keeps DELETE / GetDocument / GetByBasename strict for auth intent (`TestStoreVault_Delete_RemainsStrict` would fail).

If you'd like to contribute the `idx_vault_docs_shared` partial index (the only piece not already in `dev`), please open a fresh PR using migration number **000056** and leave the store code + schema version untouched.

PR #967, opened 2026-04-19
fix(security): fail-closed on pairing DB errors across all channels

srinis76:
## Summary

- `CheckDMPolicy` and `CheckGroupPolicy` in `internal/channels/channel.go` returned `PolicyAllow` when `IsPaired()` returned a DB error, silently granting access to unpaired senders
- Telegram DM and group pairing checks in `internal/channels/telegram/handlers.go` set `paired = true` on DB error, with the same effect
- 3 test cases in `policy_test.go` asserted the incorrect fail-open behavior

## Root Cause

The browser pairing path in `internal/gateway/router.go` already handles this correctly with an explicit comment:
```go
// Fail-closed: deny access on DB error instead of granting operator role.
```

The channel pairing paths were inconsistent with this — using fail-open instead.

This also violates the project's stated security principle (CLAUDE.md):
> **Fail-Closed** — Allowlists block by default

## Fix

- `channel.go`: `return PolicyAllow` → `return PolicyDeny` on `IsPaired` error (2 locations)
- `telegram/handlers.go`: `paired = true` → `paired = false` (left as default) on `IsPaired` error (4 locations)
- `policy_test.go`: updated 3 test cases to assert `PolicyDeny` on pairing service error

## Test plan

- [ ] Existing unit tests in `internal/channels/` pass with updated assertions
- [ ] Verify that normal paired users are unaffected (DB errors during pairing check are rare/transient)
- [ ] Verify that unpaired senders receive pairing reply instead of access during DB outage

### DIFF
--- a/internal/channels/channel.go
+++ b/internal/channels/channel.go
@@ -321,9 +321,9 @@ func (c *BaseChannel) CheckDMPolicy(ctx context.Context, senderID, dmPolicy stri
 		if c.pairingService != nil {
 			paired, err := c.pairingService.IsPaired(ctx, senderID, c.name)
 			if err != nil {
-				slog.Warn("security.pairing_check_failed, assuming paired (fail-open)",
+				slog.Warn("security.pairing_check_failed, denying access (fail-closed)",
 					"sender_id", senderID, "channel", c.name, "error", err)
-				return PolicyAllow
+				return PolicyDeny
 			}
 			if paired {
 				return PolicyAllow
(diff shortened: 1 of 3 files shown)


thieung:
@srinis76 Could you please rebase with origin/dev and resolve conflicts?
Also, please config your editor to avoid commit unrelated changes, I'm currently seeing so many changes on `internal/channels/telegram/handlers.go`
Once you've addressed, this PR is ready to merge.
Good job!

thieung:
Consolidated into #1185. Note: the diff here is ~99% CRLF→LF line-ending churn (real logic is ~7 lines) — in #1185 the same fail-closed logic is applied LF-only to keep `git blame` intact. Credited via `Co-authored-by`. Recommend closing in favor of #1185. Tip: add a `.gitattributes` with `*.go text eol=lf` to avoid CRLF in future.

PR #989, opened 2026-04-21
feat(pipeline): add tool call authorization mechanism

ba0f3:
## Summary
Tool deny rules were only applied when building the model-visible tool list, not at execution time, so `exec` could still be called via exec/bash paths.

This fix adds per-iteration runtime authorization before every tool execution (sequential and parallel), keeps deferred tool activation but re-checks deny policy, and hardens IsDenied to include alias resolution (bash -> exec).

Result: denied tools are now blocked reliably at runtime, even if the model still attempts the call.

- Introduced `AuthorizeToolCall` function to validate tool calls against a runtime allowlist.
- Updated `ToolStage` to utilize the new authorization check before executing tool calls.
- Enhanced `PipelineDeps` and `ToolState` to support the new authorization logic.
- Implemented lazy activation for deferred tools with policy checks to prevent unauthorized access.
- Ensured that the `ThinkStage` initializes the allowed tools list based on tool definitions.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/internal/agent/loop_pipeline_adapter.go
+++ b/internal/agent/loop_pipeline_adapter.go
@@ -137,6 +137,7 @@ func (l *Loop) buildPipelineDeps(req *RunRequest, bridgeRS *runState) pipeline.P
 		ExecuteToolCall:   cb.executeToolCall,
 		ExecuteToolRaw:    cb.executeToolRaw,
 		ProcessToolResult: cb.processToolResult,
+		AuthorizeToolCall: cb.authorizeToolCall,
 		CheckReadOnly:     cb.checkReadOnly,
 
 		// Observe: drain InjectCh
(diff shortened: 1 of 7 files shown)


thieung:
Consolidated into #1185, with a fix for a correctness regression: agents configured with `toolCallPrefix` had every tool call wrongly blocked (the gate looked up the raw prefixed name against an allowlist keyed by canonical names). #1185 resolves the name before the lookup and adds the gate's first direct test coverage. Credited via `Co-authored-by`. Recommend closing in favor of #1185.

ba0f3:
Thank you!

PR #1003, opened 2026-04-22
feat(tools): github_complete_check — close PR review check runs

tarrencev:
## Summary

Adds the goclaw side of cartridge-gg/internal's GitHub Checks integration. Until now, every "Code Review" check run that internal creates on a PR webhook sat stuck in `in_progress` forever — the yellow spinner on the PR Checks tab never resolved because nothing ever PATCHed the check to a terminal state. This PR adds the tool that closes it.

## What's in here

- **`internal/tools/github_complete_check.go`** — the tool. HMAC-SHA-256 signs the body (`X-Cartridge-Signature-256: sha256=<hex>`, same scheme as the inbound webhook in `internal/channels/pancake/webhook_handler.go`), retries via `providers.RetryDo`, posts under a **detached 30s ctx** so an agent-ctx cancellation right as the review finishes doesn't strand the check run. Handles the `json.Decode`-into-`map[string]any` float64 quirk for `check_run_id` so large IDs (e.g. `72,551,876,077`) don't render as scientific notation in the URL.
- **`internal/tools/context_keys.go`** — new `WithWakeMetadata` / `WakeMetadataFromCtx` helpers following the existing context-key pattern.
- **`internal/http/wake.go`** — propagates `wakeRequest.Metadata` into ctx via `WithWakeMetadata` so downstream tools can read it without the agent parsing it out of the message string. The handler already accepted `metadata` in the JSON body; it just wasn't being forwarded anywhere. Small `agentGetter` interface extracted for test stubbing.
- **`cmd/gateway_tools_wiring.go`** — registers the tool. Config is env-driven via `CARTRIDGE_INTERNAL_BASE_URL` + `CARTRIDGE_INTERNAL_WEBHOOK_SECRET`. The tool registers either way but `Execute` short-circuits with a clear error when unset, so non-integrated deployments aren't affected.

## Tests

- **`github_complete_check_test.go`** — 9 behaviour cases (happy path, no wake metadata, missing check_run_id, invalid conclusion, missing config, 500 with retry exhaustion, signature-contract pin, float64 large-ID coercion, parent-ctx-cancelled detached-ctx survival, annotation forwarding) + `TestReadCheckRunID` (11 inputs: nil/neg/fractional/string/json.Number/int/int64) + `TestSplitRepoSlug` (valid + 5 malformed).
- **`wake_test.go`** — metadata propagates through the HTTP handler into the ctx seen by `loop.Run`, nil-metadata leaves ctx unchanged, missing-message still rejected.

All green against the diff. The pre-existing `TestShellAbort_ProcessGroupKilled` flake is unrelated (fails on `origin/dev` too).

## Companion PR

Requires a companion PR on **cartridge-gg/internal** that fixes the wake payload shape — currently the service sends a raw `map[string]any`, but goclaw's handler expects `{message, metadata}` and 400s the current shape. Without that fix, wake has never actually reached Gillen for PR reviews and the tool I added here is unreachable.

**Landing order is free** — both PRs are buildable in isolation and neither regresses existing behavior when landed alone. The full loop only works once both deploy.

## Verification after both land

1. Open a PR on cartridge-gg/internal.
2. "Code Review" check run appears `in_progress` within ~15s (already working from #4378/#4380).
3. Gillen receives wake with `metadata.check_run_id`, runs `/review`, posts the comment, then calls `github_complete_check`.
4. Check run flips to `completed` with the chosen conclusion, annotations render inline on the PR diff.

Co-Authored-By: Claude Opus 4.7 (1M context)


tarrencev:
Closing — this is cartridge-fork-specific work (CARTRIDGE_INTERNAL_* env vars). Re-opening on cartridge-gg/goclaw.

PR #1004, opened 2026-04-22
fix(provider/cursor): persist chat ID to disk for session continuity across restarts

ba0f3:
## Summary

- **Problem**: Cursor CLI agent sessions weren't persisting across server restarts — chat ID was stored only in-memory
- **Solution**: Persist chat ID to `<workDir>/.cursor/chatid` file
- **Flow**: Check memory cache → check disk → call `agent create-chat` if missing → write to disk
- **Reset**: `/reset` clears both memory and disk file so next message creates a new chat

### Changes

1. `cursor_cli.go` — add `sync.Map` registry to track sessionKey → chatID per baseWorkDir
2. `cursor_cli_session.go` — refactored `buildArgs(chatID)` to take explicit chatID; added `getOrCreateChatID()`, `createChat()`, `ResetCursorCLISession()` 
3. `cursor_cli_chat.go` — call `getOrCreateChatID()` before building args in both `Chat()` and `ChatStream()`
4. `gateway_consumer_handlers.go` — call `ResetCursorCLISession()` on `/reset`
5. Updated tests

## Test

```bash
go build ./... && go test ./internal/providers/ -run TestCursor -v
```

### DIFF
--- a/cmd/gateway_consumer_handlers.go
+++ b/cmd/gateway_consumer_handlers.go
@@ -413,6 +413,7 @@ func handleResetCommand(
 	deps.SessStore.Reset(ctx, sessionKey)
 	deps.SessStore.Save(ctx, sessionKey)
 	providers.ResetCLISession("", sessionKey)
+	providers.ResetCursorCLISession("", sessionKey)
 	slog.Info("inbound: /reset command", "session", sessionKey)
 
 	return true
--- a/cmd/gateway_providers.go
+++ b/cmd/gateway_providers.go
@@ -198,6 +198,23 @@ func registerProviders(registry *providers.Registry, cfg *config.Config, modelRe
 		slog.Info("registered provider", "name", "claude-cli")
 	}
 
+	// Cursor CLI provider (subscription-based, no API key needed)
+	if cfg.Providers.CursorCLI.CLIPath != "" {
+		cliPath := cfg.Providers.CursorCLI.CLIPath
+		var opts []providers.CursorCLIOption
+		if cfg.Providers.CursorCLI.Model != "" {
+			opts = append(opts, providers.WithCursorCLIModel(cfg.Providers.CursorCLI.Model))
+		}
+		if cfg.Providers.CursorCLI.Mode != "" {
+			opts = append(opts, providers.WithCursorCLIMode(cfg.Providers.CursorCLI.Mode))
+		}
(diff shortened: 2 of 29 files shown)


ba0f3:
## Review fixes applied

### P0 — Race condition in getOrCreateChatID
- Added `chatIDMu sync.Map` to `CursorCLIProvider` struct (per-session mutex, separate from `sessionMu` to avoid deadlocks with `lockSession`)
- `getOrCreateChatID` now uses double-check pattern: fast path (registry) → acquire per-session mutex → second check (registry + disk) → `createChat`
- This protects against concurrent `createChat` calls even if the method is called outside `lockSession`

### P1 — createChat discards stderr
- Changed from `cmd.Output()` to explicit `cmd.Stderr = &stderr`
- Error messages now include stderr content: `cursor-cli create-chat: %w (stderr: %s)`

### P2 — chatIDFile directory creation silently ignored
- Restructured: `MkdirAll` failure is now logged at `Error` level (was `Warn`)
- If `MkdirAll` fails, the chat ID is still returned so the session works, but it's clear from logs that persistence failed
- `WriteFile` failure also logged at `Error` level

### Tests added
- `TestResetCursorCLISession` — verifies registry entry and disk file are both cleared
- `TestGetOrCreateChatID_LoadsFromDisk` — pre-writes a chat ID file and verifies it's loaded without calling `createChat`

All tests pass: `go build ./...`, `go vet ./...`, `go test ./internal/providers/ ./cmd/`

ba0f3:
## Additional change: team workspace in sandbox

The sandbox now mounts the team workspace alongside the personal workspace when running in team context.

### Changes

1. `sandbox.Config` — new `TeamWorkspace` field
2. `SandboxConfigFromCtx` — injects team workspace from context into sandbox config (returns a copy so the original config isn't mutated)
3. `newDockerSandbox` — mounts team workspace at `/team-workspace` (same rw/ro permissions as primary workspace)
4. `DockerManager.Get` — compares both primary and team workspace when deciding whether to reuse an existing sandbox container; recreates if either changed

### Behavior

- In team sessions, sandboxed tools (`exec`, `shell`, `write_file`, etc.) can access both:
  - `/workspace` — personal workspace (primary mount)
  - `/team-workspace` — team shared workspace (additional mount)
- The system prompt already tells agents the team workspace path; sandboxed exec/shell commands can now access it

PR #1002, opened 2026-04-22
feat: native image_generation for Codex + OpenAI-compat providers

kaitranntt:
Closes #1000

## Summary

Enables real ChatGPT image generation for agents routed through ChatGPT (Codex) OAuth accounts. The existing `create_image` builtin tool now routes to a new `NativeImageProvider` interface when the chain points at an OAuth-backed provider, so image generation works without the tool demanding static API credentials. Default model is `gpt-image-2` (the motivation of this PR); `gpt-image-1.5` is available as a legacy fallback via a whitelisted UI selector.

Verified end-to-end against a live backend with a real Codex OAuth account — see the visual trace below.

## Visual trace

[Full UX report (htmlpreview)](https://htmlpreview.github.io/?https://raw.githubusercontent.com/kaitranntt/goclaw/8dd1590fd0d1c7f6d55b3d6beb33752d95562bf1/.github/pr-assets/1002/index.html)

### 1 · Inline result — tool completes, image + prompt caption render

Tool flow, create_image Done, generated infographic rendered inline with prompt caption beneath

`create_image` returns `Done`. Generated Vietnamese Red Fox encyclopedia infographic renders inline via `MediaGallery`. The italic muted line beneath the image is the new **prompt caption** — rendered from `MediaRef.Prompt` and also embedded into the PNG's `tEXt` chunk on write, so downloaded files carry their own provenance.

### 2 · Expanded view — click to open the lightbox

Full 1024x1792 infographic in MediaGallery lightbox overlay

Full-resolution render through the existing `MediaGallery` lightbox. Download button top-right; filename resolved from `filename_hint` + timestamp (`cao-do-infographic-vietnamese_YYYYMMDD-HHmmss_….png`). The image lives in the assistant's workspace at `{workspace}/media/{sha256}.{ext}`, deduplicated on hash.

### 3 · Where to configure the image model

Create Image Provider Chain dialog with openai-codex card expanded, Image model dropdown open showing Default and Legacy options

**Built-in Tools → create_image → Settings → openai-codex → Settings (expand) → Image model.**
Whitelist: `gpt-image-2` (default, recommended) · `gpt-image-1.5` (legacy). Anything else rejected server-side by `ValidateImageModel` with a clear error. Stored as `params.image_model` in the chain entry JSON; threaded to the native tool via `NativeImageRequest.ImageModel`. Also visible in this screenshot: **Timeout 600s / Retries 1** — the new chain defaults that match realistic gpt-image-2 runtime.

## What changed

[...]

PR #1005, opened 2026-04-22
fix: correct STT routing and shell security runtime persistence

jahrulnr:
## Summary
This PR fixes two runtime behavior bugs in GoClaw: (1) `read_audio` STT routing now correctly uses transcription path semantics instead of falling through to chat/completions behavior in wrong cases, and (2) Shell Security Groups changes now persist and apply at runtime (no restart required), including reducing model pre-refusal bias so `exec` is attempted and policy decides allow/deny per turn.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [ ] `go build ./...` passes
- [ ] `go vet ./...` passes

## Test Plan
1. Automated checks run:
   - `go test ./cmd`
   - `go test ./internal/agent -run TestNoneModeSections`
   - `go test ./internal/tools`-level focused tests added:
     - `internal/tools/read_audio_resolve_test.go`
     - `internal/tools/shell_global_deny_groups_test.go`
2. Build verification:
   - `GOOS=linux GOARCH=arm64 make build-full` (includes web build + embed UI + Go binary build)
3. Manual runtime verification:
   - Updated Shell Security Groups in `/config`, saved, and confirmed changes apply without process restart.
   - Verified `sudo` behavior follows current deny-group config (no stale strict policy after save).
   - Verified agent prompt behavior no longer over-generalizes credentialed CLI approval wording to normal shell commands.
4. Deployment verification (ARM):
   - Binary deployed to `/usr/local/bin/goclaw`
   - Process restarted and verified running new binary hash.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ compose.d/*
 *.test
 goclaw-patched-linux-amd64
 ui/web/nginx.staging.conf
+dist
--- a/cmd/gateway_lifecycle.go
+++ b/cmd/gateway_lifecycle.go
@@ -84,6 +84,24 @@ func (d *gatewayDeps) runLifecycle(
 		deps.webFetchTool.UpdatePolicy(updatedCfg.Tools.WebFetch.Policy, updatedCfg.Tools.WebFetch.AllowedDomains, updatedCfg.Tools.WebFetch.BlockedDomains)
 	})
 
+	// Reload global shell deny-group toggles on config changes via pub/sub.
+	// Without this, Config page changes only apply after process restart.
+	d.msgBus.Subscribe("shell-deny-groups-config-reload", func(evt bus.Event) {
+		if evt.Name != bus.TopicConfigChanged {
+			return
+		}
+		updatedCfg, ok := evt.Payload.(*config.Config)
+		if !ok {
+			return
+		}
+		if execTool, ok := d.toolsReg.Get("exec"); ok {
+			if et, ok := execTool.(*tools.ExecTool); ok {
+				et.SetGlobalShellDenyGroups(updatedCfg.Tools.ShellDenyGroups)
+				slog.Info("shell deny groups reloaded via pub/sub", "groups", len(updatedCfg.Tools.ShellDenyGroups))
+			}
+		}
+	})
+
 	// Reload TTS providers on config changes via pub/sub.
 	d.msgBus.Subscribe("tts-config-reload", func(evt bus.Event) {
 		if evt.Name != bus.TopicConfigChanged {
--- a/cmd/gateway_setup.go
+++ b/cmd/gateway_setup.go
@@ -14,11 +14,11 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/bootstrap"
 	"github.com/nextlevelbuilder/goclaw/internal/bus"
 	"github.com/nextlevelbuilder/goclaw/internal/config"
+	"github.com/nextlevelbuilder/goclaw/internal/edition"
 	mcpbridge "github.com/nextlevelbuilder/goclaw/internal/mcp"
 	"github.com/nextlevelbuilder/goclaw/internal/permissions"
 	"github.com/nextlevelbuilder/goclaw/internal/providers"
 	"github.com/nextlevelbuilder/goclaw/internal/sandbox"
-	"github.com/nextlevelbuilder/goclaw/internal/edition"
 	"github.com/nextlevelbuilder/goclaw/internal/skills"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
 	"github.com/nextlevelbuilder/goclaw/internal/store/pg"
@@ -211,6 +211,9 @@ func setupToolRegistry(
 	// Exception: .goclaw/skills-store/ is allowed (skills may contain executable scripts).
 	if execTool, ok := toolsReg.Get("exec"); ok {
 		if et, ok := execTool.(*tools.ExecTool); ok {
+			// Global shell deny-group toggles from Config page (tools.shellDenyGroups).
+			// Per-agent shell_deny_groups still override per group at runtime.
+			et.SetGlobalShellDenyGroups(cfg.Tools.ShellDenyGroups)
 			et.DenyPaths(dataDir, ".goclaw/")
 			// Allow skills execution: master-tenant skills-store + all tenant-scoped skills-store dirs.
 			et.AllowPathExemptions(
@@ -595,4 +598,3 @@ func setupSkillsSystem(
 
 	return skillsLoader, skillSearchTool, globalSkillsDir, bundledSkillsDir, builtinSkillsDir
 }
-
--- a/internal/agent/resolver.go
+++ b/internal/agent/resolver.go
@@ -458,7 +458,7 @@ func NewManagedResolver(deps ResolverDeps) ResolverFunc {
 			evoMetricsStore = deps.EvolutionMetricsStore
 		}
 
-		restrictVal := true // always restrict agents to their workspace
+		restrictVal := ag.RestrictToWorkspace
 		loop := NewLoop(LoopConfig{
 			ID:                     ag.AgentKey,
 			DisplayName:            ag.DisplayName,
--- a/internal/agent/systemprompt.go
+++ b/internal/agent/systemprompt.go
@@ -335,7 +335,7 @@ func BuildSystemPrompt(cfg SystemPromptConfig) string {
 	}
 
 	// 2. ## Tooling
-	lines = append(lines, buildToolingSection(cfg.ToolNames, cfg.SandboxEnabled, cfg.ShellDenyGroups)...)
+	lines = append(lines, buildToolingSection(cfg.ToolNames, cfg.SandboxEnabled, cfg.ShellDenyGroups, !isNone)...)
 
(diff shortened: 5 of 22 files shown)


clark-cant:
## 🔍 Code Review — fix: correct STT routing and shell security runtime persistence

### 🎯 Overview

This PR fixes two main bugs: (1) `read_audio` STT routing now uses the correct transcription path instead of falling through to chat/completions, and (2) Shell Security Groups changes from the Config page are applied at runtime via pub/sub without requiring a restart. Additionally includes system prompt adjustments to reduce pre-refusal bias and TTS tenant provider support.

**Scope:** 22 files, +405/-127 lines

---

### ✅ Positive Points

1. **Shell deny groups hot-reload via pub/sub** — follows GoClaw's established pattern correctly. Subscribes to `TopicConfigChanged` in `gateway_lifecycle.go`, consistent with the existing TTS reload pattern. Clean and idiomatic.

2. **`RestrictToWorkspace` bug fix** — the hardcoded `restrictVal := true` was a real bug. Now correctly uses `ag.RestrictToWorkspace` from config. Important fix.

3. **System prompt pre-refusal reduction** — the `includeExecNudge` flag properly gates nudge messages for none-mode, avoiding agent confusion in limited mode. Good separation of concerns.

4. **TTS tenant provider with fallback chain** — the `ResolveTenantProvider` → try tenant → fallback global → error handling flow is clear and well-tested.

5. **Comprehensive tests** — 4 new test files: `read_audio_resolve_test.go`, `shell_global_deny_groups_test.go`, `context_keys_test.go`, `tts_agent_params_test.go`. Good coverage.

---

### ⚠️ Issues to Address

#### 1. [MEDIUM] — `config_channels.go` whitespace reformat is too large

**File:** `internal/config/config_channels.go`

**Issue:**
66 additions / 66 deletions are purely struct field indentation alignment (Telegram, Providers, Gateway, Tools, Browser, ToolPolicySpec). No logic changes in this file.

**Suggestion:**
Split into a separate commit `chore: normalize config struct indentation` to make the diff easier to review. Mixing formatting changes with logic changes makes it harder to spot real bugs.

---

#### 2. [MEDIUM] — `goto synthDone` is not idiomatic Go

**File:** `internal/tools/tts.go:288`

```go
if err == nil {
    goto synthDone
}
```

**Issue:**
`goto` is rarely needed in Go. This pattern can be refactored into a helper function or early return.

**Suggested fix:**
Extract the global fallback chain into a separate method:
```go
if tenantProvider, tenantName, _, ok := mgr.ResolveTenantProvider(ctx); ok && tenantProvider != nil {
    tenantOpts := opts
    if adapted := audio.AdaptAgentParams(genericAgentParams, tenantName); len(adapted) > 0 {
        tenantOpts.Params = mergeParams(opts.Params, adapted)
    }
    result, err = tenantProvider.Synthesize(ctx, text, tenantOpts)
    if err != nil {
        slog.Warn("tts tenant provider failed, trying global fallback", "provider", tenantName, "error", err)
        result, err = t.resolveAndSynthesize(ctx, mgr, text, opts)
    }
} else {
    result, err = t.resolveAndSynthesize(ctx, mgr, text, opts)
}
```

---

#### 3. [LOW] — `.gitignore` missing trailing newline

**File:** `.gitignore`

`+dist\ No newline at end of file` — should add a trailing newline to avoid git warnings.

---

#### 4. [LOW] — `SetGlobalShellDenyGroups` needs thread safety review

**File:** `internal/tools/shell.go`

`SetGlobalShellDenyGroups` is called from two places:
- `gateway_setup.go` — during initialization (single-threaded)
- `gateway_lifecycle.go` — from pub/sub callback (potentially concurrent with exec calls)

Need to verify that `globalShellDenyGroups` is protected by a mutex or uses atomic swap. If it's an `atomic.Pointer` or copy-on-write pattern, it's fine. But if it's a plain map assignment, there could be a race condition.

---

### 📊 Summary

| Severity | Count | Issues |
|----------|-------|--------|
| 🔴 CRITICAL | 0 | — |
| 🟡 HIGH | 0 | — |
| 🟡 MEDIUM | 2 | Config whitespace noise, goto pattern |
| 🟢 LOW | 2 | Missing newline, thread safety review |

---

### 💡 Recommendation

**🟢 APPROVE WITH CHANGES**

The core logic changes are on the right track and address root causes correctly. The issues above are improvements — not blocking for merge but should be fixed before release. Especially recommend splitting the config whitespace into a separate commit for cleaner git history.

Great work @jahrulnr! 🚀 Let me know if you want to discuss any of these points.

viettranx:
Thanks for the work here, @jahrulnr. After reviewing carefully, we've decided to **close this PR** and land the valuable fixes through targeted follow-ups. Breakdown:

## What we'll take (cherry-pick as separate focused PRs)

1. **STT routing fix** — `internal/tools/read_audio_resolve.go`. Correctly fails fast when a transcription-named model is used without credentials, and routes `openai_compat` transcription models to `/v1/audio/transcriptions` instead of falling through to chat/completions. Clean root-cause fix.
2. **Shell deny-groups runtime reload** — `cmd/gateway_lifecycle.go` + `cmd/gateway_setup.go` + `internal/tools/shell.go` + the `ShellDenyGroups` field in `ToolsConfig`. Follows the existing `TopicConfigChanged` pub/sub pattern; agent-context overrides still win per-group. We'll add a unit test asserting the subscriber actually updates `ExecTool.globalDenyGroups` on event dispatch (missing from this PR).
3. **Credentialed CLI disambiguation** — the wording change in `internal/tools/credential_context.go` scoping the "blocked" guidance to `[CREDENTIALED EXEC]` errors. Addresses the real over-generalization issue without polluting the prompt.

## What we're dropping and why

4. **`restrict_to_workspace` per-agent mutation** (`resolver.go`, `context_keys.go`, `http/agents.go`, `http/validate.go`, related tests). This is the biggest concern:
   - Not mentioned in the PR description.
   - The prior hardcoded `return true` in `effectiveRestrict()` and `restrictVal := true` in the resolver were explicit multi-tenant security hardening ("Agents must not access files outside their tenant-scoped workspace").
   - This change lets any tenant admin `PATCH /v1/agents/:id {"restrict_to_workspace": false}` to break an agent out of its workspace — a silent tenant-isolation regression with no added gating (no `IsOwnerRole` / `IsCrossTenant` check, no audit log, no UI affordance).
   - If there's a legitimate desktop/lite usecase, please open a separate PR that gates the field behind owner scope and documents the security model.
5. **System prompt "do not pre-refuse" nudges** (`systemprompt.go` exec/read_file nudges + `systemprompt_sections.go` execution-bias file-access line). These are band-aids placed in the stable (cached) prompt, costing tokens every turn forever, and the `exec` nudge isn't even gated on tool presence while the `read_file` one is — inconsistent. The better fix for the underlying symptom ("model sees old denial in history and refuses next turn even after policy reload") is either an episodic prompt-refresh hint after a config reload, or keeping it in the dynamic post-cache-boundary section. We'll explore that separately.
6. **`includeExecNudge` slim tooling mode**. The parameter name is misleading (gates sandbox, package, media, and "Ignore \"not available\" in history" too), and crucially `!isNone` means Minimal mode (subagent/cron) also loses the authoritative-tool-list guidance, while PromptNone mode silently drops the same line — which directly undermines the PR's own anti-pre-refusal goal. Not worth the complexity.
7. **`read_file` description hedging** ("may be workspace-only"). Weak language for LLMs. If we ever support per-agent restriction toggles, we'll compute the description dynamically from the actual config rather than hedging.
8. **TTS tenant provider preference** (`tts.go`). The idea (prefer tenant-scoped provider since tenant secrets aren't in the global env registry) is valid, but the implementation uses `goto synthDone` and mixes explicit-provider and fallback logic. Please resubmit as a standalone PR with cleaner control flow and tests covering (a) tenant provider succeeds, (b) tenant fails → global fallback, (c) explicit provider name matches tenant.
9. **`MethodSessionsCompact` in `isWriteMethod`** — already on dev via #1002 (commit `4b02c273`) and the duplicate was removed in `9576bbe0`. No-op now.
10. **`config_channels.go` formatting-only churn** — pure gofmt noise unrelated to the fix, makes review harder; dropped.

## Issues worth calling out for next time

- **Scope creep**: the PR bundled ~6 different concerns. Please split — each of (1)(2)(3)(4)(8) above was a valid standalone PR.
- **Stale base**: branch is forked from `a7962d18` and CI is red because later integration-test fixes (`e7c78263`, `5db4e063`) haven't been rebased in. Always rebase on dev before pushing for review.
- **Commit hygiene**: commit `4ff47365` concatenates 5 fixes into one message. Conventional commit per change please — no AI-generated concatenations.
- **Missing regression test for the headline fix**: shell-deny-groups runtime reload is one of the two stated fixes, but no unit test exercises the subscriber dispatch. Manual `/config` → save → verify is fine as smoke, not a substitute.

Closing now. Will land (1)(2)(3) shortly as focused PRs — feel free to follow up on (8) when ready. Thanks again for surfacing the STT and shell-persistence issues — those were real bugs.

viettranx:
Closing per review — see detailed breakdown above. Valuable STT routing, shell deny-groups runtime reload, and credentialed-CLI wording fixes will be cherry-picked as focused follow-up PRs. Other changes dropped due to silent security regression (restrict_to_workspace), scope creep, or being band-aids.

PR #1023, opened 2026-04-23
feat(discord): real-time voice-channel transcription

tarrencev:
## Summary
- New opt-in voice-channel join + per-speaker transcription subsystem for the Discord adapter.
- Bot listens for `VoiceStateUpdate` in the configured guild, joins the voice channel when humans are present, demuxes per-user Opus streams via `VoiceConnection.OpusRecv`, packages each utterance into Ogg/Opus (via `pion/webrtc/v4/pkg/media/oggwriter` — no PCM decode, no CGO), and posts `<DisplayName>: <transcript>` lines to a configured text channel through `audio.Manager.Transcribe`.
- Distinct from the pre-existing `voice_agent_id` field (which routes *uploaded* voice-message attachments, not real-time voice channels). Naming uses `VoiceChannel*` to disambiguate.

## Enabling the feature

Per-instance config on `discordInstanceConfig` / `DiscordConfig`:
- `voice_channel_enabled` (bool, required)
- `voice_channel_guild_id` (required)
- `voice_channel_id` (required)
- `voice_channel_transcript_channel_id` (required)
- `voice_channel_idle_leave_seconds` (default 60)
- `voice_channel_min_utterance_ms` (default 400)
- `voice_channel_max_utterance_ms` (default 10000)
- `voice_channel_daily_cap_seconds` (default 7200 = 2h/day — bounds runaway STT cost)

Intents: `IntentsGuilds | IntentsGuildVoiceStates` are now added unconditionally (additive change, existing text handling unaffected — regression test included). Bots that opt into voice must also toggle "Server Voice State" in the Discord Developer Portal.

## Design constraints (locked in code + tests)

- **Drain-only consumer.** `OpusRecv` is buffered to 2 packets in `bwmarrin/discordgo`; any I/O in the consumer drops frames silently. Stress test verifies zero-drop at 1000 packets/sec.
- **`deaf=false` on join.** Common footgun — `deaf=true` silently disables `OpusRecv`.
- **Explicit `VoiceConnection.Disconnect()` and goroutine drain on Stop.** discordgo otherwise leaks voice loops.
- **`safego.Recover` on every new goroutine.** A voice-path panic must not take down text handling.
- **SSRC → UserID race.** Opus packets can arrive before the matching `VoiceSpeakingUpdate`; buffers created by early packets get `userID` retroactively attached.
- **SSRC type cast guarded.** `VoiceSpeakingUpdate.SSRC` is `int`, `Packet.SSRC` is `uint32` — explicit cast with regression test.

## Operational controls

- Daily cap (`voice_channel_daily_cap_seconds`) stops transcribing mid-day if the budget is exceeded; bot stays joined and logs.
- Local LRU+TTL cache (1h) avoids a `GuildMember` REST call per transcript line.
- STT error taxonomy:
  - Transient (5xx, network): drop with warn.
  - Quota (429): open 60s circuit, drop in-flight.
  - Auth (401/403): disable STT for session, loud log (requires config reload / pod restart).
- Orphan tmpfile sweeper runs every 5min.
- Message truncation at 1900 chars with `...` suffix (well under Discord's 2000 cap).

## Dependency added

- `github.com/pion/webrtc/v4` — solely for `pkg/media/oggwriter`. Pure Go, MIT, narrow transitive graph (only `pion/rtp` + `pion/randutil`). Chose this over hand-rolling RFC 7845 page encoding per review discussion.

## Test plan

- [x] `go test -race -short ./internal/channels/discord/...` passes
- [x] `go test -race -run Test_drain_stress ./internal/channels/discord/voice/...` passes (10K packets, zero capacity drops)
- [x] `go vet ./internal/channels/discord/...` clean
- [x] Intents-bitmask regression test asserts the union of required flags
- [x] 35+ unit tests across state machine, error taxonomy, cache, daily-cap rollover, ogg round-trip
- [ ] Integration test against a real Discord bot token in a test guild (to run after merge on the consumer side)
- [ ] End-to-end with ElevenLabs Scribe — smoke test only per review decision; real WER evaluation is captured as a follow-up TODO


tarrencev:
Opened against upstream by mistake; reopening against cartridge-gg/goclaw (our maintained fork).

PR #1050, opened 2026-04-27
chore(deps): bump cartridge-gg/discordgo to v0.29.1-dave.13 (DAVE OP4 key_package)

tarrencev:
## Summary
- Bump discordgo pin from `dave.12` → `dave.13`.
- Picks up [cartridge-gg/discordgo#841fbfc](https://github.com/cartridge-gg/discordgo/commit/841fbfc): proactively send the MLS key package via voice opcode 26 when OP4 (session description) carries a non-zero `dave_protocol_version`.

## Why
Without this, the voice gateway connects on v8 + AEAD-rtpsize, but the MLS handshake never starts. Discord doesn't initiate it on its own; the client has to push its key package first. Result: every audio frame reaches STT as inner-DAVE ciphertext, which the transcription model hallucinates into multilingual sound-effect descriptions ("French beatbox", "Korean radio static", etc.).

After this bump, the bot enters the MLS group, receives commits/welcomes, derives the per-epoch key ratchet, and the inner DAVE decrypt produces real Opus frames. That's the last piece needed for the prod #chill voice channel transcripts to come out as actual English speech.

## Test plan
- [x] discordgo unit tests pass (`TestDecryptAEAD*`, `TestParseDAVEBinaryHeader`, etc.) on the new tag.
- [ ] Image rebuild + deploy (will follow in internal.voice runtime/goclaw.Dockerfile bump).
- [ ] Smoke: join #chill, speak, confirm transcript channel shows English text rather than hallucinated SFX.


tarrencev:
Wrong remote — re-opening against cartridge-gg/goclaw

PR #1063, opened 2026-04-28
fix(providers): resolve HTTP 400 DeepSeek reasoning passback error during tool calls

### LINKED ISSUE #1047: [Bug] DeepSeek V4 Pro: Lỗi 400 do thiếu reasoning_content trên kênh Zalo (OpenAI-compatible)
### Tóm tắt
- Khi dùng model `deepseek-v4-pro/flash` qua kênh Zalo (hoặc các adapter OpenAI-compatible),  API DeepSeek trả lỗi:
```
The `reasoning_content` in the thinking mode must be passed back to the API.
```
- Nguyên nhân: GoClaw (hoặc adapter) **không giữ lại và gửi lại reasoning_content** trong lịch sử hội thoại khi build request tiếp theo. OpenAI message format không chứa trường này nên bị mất khi truyền qua các channel như Zalo.

### Phạm vi ảnh hưởng
- Mọi kênh sử dụng OpenAI-compatible message format với DeepSeek V4/V4 Pro (Zalo, ...).

### Đề xuất hướng xử lý
- Khi nhận response từ DeepSeek có reasoning_content, cần lưu/thêm reasoning_content vào assistant messages khi xây lại request history cho turn tiếp theo trên các kênh dùng OpenAI message format.

### Mức độ ưu tiên
- High: lỗi chặn multi-turn/kịch bản agent khi dùng DeepSeek qua các messaging channel.

nagaame:
## Description
This PR fixes an HTTP 400 `invalid_request_error` encountered when using DeepSeek models (e.g., `deepseek-v4-pro`, `deepseek-reasoner`) with Thinking Mode enabled during multi-turn tool calling.
The DeepSeek API enforces a strict schema requirement: any `assistant` message in the context history that includes `tool_calls` **must** also include the `reasoning_content` field, even if the model did not emit any reasoning tokens in that specific turn. Previously, WorkBuddy would either omit this field when it was empty, or improperly drop the reasoning content entirely inside the provider layer when `stripThinking` was active.
### Changes Made
1. **Preserve Reasoning History (`internal/providers/openai_chat.go`)**: 
   - Ensure `result.Thinking` is always accumulated regardless of the `stripThinking` flag. This guarantees that the chain of thought is correctly added to the message history for subsequent API passback.
2. **Strict Schema Compliance (`internal/providers/openai_request.go`)**: 
   - When building the request payload for DeepSeek models, if an assistant message's `Thinking` is empty, explicitly inject `reasoning_content: ""` into the JSON. This satisfies DeepSeek's strict JSON schema validation.
3. **Safe UI Output Stripping (`internal/agent/loop_run.go`)**: 
   - Moved the responsibility of hiding leaked reasoning from the UI out of the provider layer. `result.Thinking` is now explicitly cleared from the final `completedPayload` at the end of the agent loop if `reasoningDecision.StripThinking` is true.
4. **Unit Testing (`internal/providers/openai_request_test.go`)**: 
   - Added specific tests (`TestBuildRequestBody_DeepSeekReasoningContent`) to ensure `reasoning_content` is accurately injected for DeepSeek models and safely omitted for standard OpenAI models (like GPT-4o).
## Related Issues
Fixes #1047

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/internal/agent/loop_run.go
+++ b/internal/agent/loop_run.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/google/uuid"
 
+	"github.com/nextlevelbuilder/goclaw/internal/providers"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
 	"github.com/nextlevelbuilder/goclaw/internal/tools"
 	"github.com/nextlevelbuilder/goclaw/internal/tracing"
@@ -216,6 +217,25 @@ func (l *Loop) Run(ctx context.Context, req RunRequest) (*RunResult, error) {
 		if isChildTrace && l.traceCollector != nil && traceID != uuid.Nil {
 			l.traceCollector.SetTraceStatus(ctx, traceID, store.TraceStatusCompleted)
 		}
+
+		provider := l.provider
+		if req.ProviderOverride != nil {
+			provider = req.ProviderOverride
+		}
+		model := l.model
+		if req.ModelOverride != "" {
+			model = req.ModelOverride
+		}
+		reasoningDecision := providers.ResolveReasoningDecision(
+			provider, model,
+			l.reasoningConfig.Effort,
+			l.reasoningConfig.Fallback,
+			l.reasoningConfig.Source,
+		)
+		if reasoningDecision.StripThinking {
+			result.Thinking = ""
+		}
+
 		completedPayload := map[string]any{"content": result.Content}
 		if result.Thinking != "" {
 			completedPayload["thinking"] = result.Thinking
--- a/internal/providers/openai_chat.go
+++ b/internal/providers/openai_chat.go
@@ -28,14 +28,11 @@ func (p *OpenAIProvider) Chat(ctx context.Context, req ChatRequest) (*ChatRespon
 		}
 	}
 
-	// Drop user-visible reasoning for models flagged as leakers (e.g. Kimi,
-	// DeepSeek-Reasoner). Usage.ThinkingTokens is preserved so billing stays
-	// correct (Phase 1 depends on this).
-	if resp != nil {
-		if strip, _ := req.Options[OptStripThinking].(bool); strip {
-			resp.Thinking = ""
-		}
-	}
+	// Note: We used to drop user-visible reasoning here for leaky models.
+	// However, DeepSeek strictly requires `reasoning_content` to be passed back
+	// on tool call iterations. Clearing `resp.Thinking` here causes HTTP 400s
+	// on subsequent passbacks. The responsibility of hiding leaked reasoning
+	// from the UI is now handled in the Agent loop (e.g. final payload stripping).
 
 	return resp, err
 }
@@ -126,9 +123,9 @@ func (p *OpenAIProvider) ChatStream(ctx context.Context, req ChatRequest, onChun
 		if reasoning == "" {
 			reasoning = delta.Reasoning
 		}
-		if reasoning != "" && !stripThinking {
+		if reasoning != "" {
 			result.Thinking += reasoning
-			if onChunk != nil {
+			if !stripThinking && onChunk != nil {
 				onChunk(StreamChunk{Thinking: reasoning})
 			}
 		}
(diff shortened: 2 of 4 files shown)


thieung:
Reviewed alongside #1101 (which addresses the same upstream issue #1047 from a different angle). Verified locally on the PR branch:

- `go build ./...` ✓
- `go build -tags sqliteonly ./...` ✓
- `go vet ./...` ✓
- `go test -race ./internal/providers/... ./internal/agent/...` ✓

### What's correct

- Removing the `resp.Thinking = ""` strip in `openai_chat.go:Chat` and accumulating `result.Thinking` unconditionally in `ChatStream` is the right move: persistence of Thinking is needed for the next-turn `reasoning_content` passback to DeepSeek/Kimi. Stripping responsibility moves cleanly into the agent loop.
- Gating the post-Run strip on `ResolveReasoningDecision.StripThinking` is correct — it only fires for known leakers (Kimi / deepseek-reasoner / gemma-4) with `effective_effort == off`, so non-leaky callers are unaffected.

### Concerns

**Important**

1. **`reasoning_content: ""` may not actually satisfy DeepSeek.** The error in #1047 is *\"The `reasoning_content` in the thinking mode must be **passed back** to the API\"* — \"passed back\" implies echoing the model's prior reasoning, not just any value. Empty-string compliance is plausible but unverified. Please confirm against a real DeepSeek thinking-mode endpoint (e.g. `deepseek-v4-pro` / `deepseek-reasoner`) with an iteration where the prior assistant turn legitimately has no reasoning — does the API actually accept `\"\"`, or does it require a non-empty placeholder? #1101 takes the safer route by injecting a non-empty string into the synthetic post-summary message.

2. **Empty `reasoning_content` injection applies to non-thinking DeepSeek variants too.** The new branch in `openai_request.go` triggers for any model whose name contains `deepseek` and where `m.Thinking == \"\"`. That includes `deepseek-chat` (non-thinking) where the API may not expect the field at all. Consider narrowing to thinking-mode DeepSeek (e.g. `deepseek-reasoner`, `deepseek-v4-pro`, or via `modelLeaksReasoning`/an explicit allowlist) instead of any `deepseek*` model.

3. **Test name and coverage gap.** `TestBuildRequestBody_DeepSeekReasoningContent` is named to suggest the injection is conditional on `tool_calls`, but the production code is not — it injects whenever assistant Thinking is empty. The test passes ToolCalls in both subcases but never asserts the no-tool_calls case explicitly. Recommend either (a) renaming to drop the tool_call implication, or (b) adding a no-tool_calls assistant-turn case to lock down the actual contract.

4. **Test plan checkbox `go test -race ./...` is not checked.** This is a multi-file change touching the agent loop and provider wire path; full suite should pass before merge. Ran it locally on the PR branch and it passes — please check the box.

**Suggestion**

5. **Double-resolve of `ResolveReasoningDecision`.** The same call is made inside `callLLM` (`internal/agent/loop_pipeline_callbacks.go:244`) to set `OptStripThinking` for the request, and now again in `Loop.Run` (`loop_run.go`) to clear `result.Thinking`. Two call sites can drift over time. Consider either (a) propagating the resolved decision through `RunResult` or `runState`, or (b) checking `result.Thinking != \"\"` plus an explicit `provider+model` strip flag set when `OptStripThinking` was actually applied to the request. Minor — not blocking.

6. **Asymmetry with Anthropic stream.** `internal/providers/anthropic_stream.go:93` still gates `result.Thinking += ev.Delta.Thinking` on `!stripThinking`. After this PR, OpenAI accumulates Thinking unconditionally; Anthropic does not. In practice Anthropic isn't a known leaker so `StripThinking` rarely fires for it, but the two providers now follow different conventions. Worth a follow-up to make them consistent if `StripThinking` for Anthropic is ever expected to also need history passback.

### Coordination with #1101

These PRs are complementary:

- **#1101** fixes the synthetic post-summary assistant ack in `loop_history.go` (Thinking was unset because no LLM call produced it). Uses a non-empty placeholder string. Independent of provider.
- **This PR** fixes the response-side strip (`openai_chat.go`) and adds defensive `reasoning_content: \"\"` for empty-Thinking assistant turns in DeepSeek requests.

If empty-string `reasoning_content` is actually accepted by DeepSeek (concern #1 above), this PR alone could close #1047 — the defensive fallback would catch the synthetic case too. If empty string is **not** accepted, then #1101 is required for the synthetic-message path.

Worth aligning on which behavior is canonical before both merge, to avoid one of them becoming dead defensive code.

PR #1074, opened 2026-05-01
Pass locale through cron run path so meta lines render in user's language

codebit0:
## Summary

Cron-triggered agent runs had no inbound channel locale (no WS connect
param, no `Accept-Language` header), so the system prompt and the cron
extra-context block were always built in English regardless of the
user's preferred language. The English meta lines — sitting at the
beginning and end of the prompt, where models weight instructions most
— pushed responses into English even when context files explicitly
asked for another language.

This change wires locale through the cron path using **only existing
data structures**: `tenants.settings` and `tenant_users.metadata` jsonb
columns. No new environment variable, no new `agents.other_config`
key, no schema change.

## Resolution chain

```
cron run
  ├─ tenant_users.metadata->>'locale'   (per-user override; jsonb field already exists)
  ├─ tenants.settings->>'locale'        (tenant default; jsonb field already exists)
  └─ i18n.DefaultLocale                 (= LocaleEN, unchanged fallback)
```

Empty resolution skips `store.WithLocale` entirely, so existing
deployments with no `locale` set anywhere keep getting English — no
behavioural regression.

## Changes

- `internal/i18n`: add `LocaleKO` plus `catalog_ko.go`. Only the new
  prompt-meta keys are translated; everything else falls through to the
  English catalog.
- `internal/i18n/keys.go`: five new keys —
  `MsgSysChannelGreeting{Direct,Group,GroupTitled}` and
  `MsgSysCronJobMeta{Deliver,NoDeliver}`.
- `internal/agent/systemprompt.go`: add `SystemPromptConfig.Locale` and
  replace the hardcoded *"You are a personal assistant running in
  X (a direct chat)."* line with `i18n.T()`.
- `internal/agent/loop_history.go`: feed
  `store.LocaleFromContext(ctx)` into `SystemPromptConfig.Locale`.
- `cmd/gateway_cron.go`: `resolveCronLocale()` cascade and an
  `i18n.T`-driven `extraPrompt`. Cron context picks up `WithLocale`
  before scheduling the agent loop.
- `cmd/gateway_heartbeat.go`: pass `pgStores.Tenants` to the cron
  handler factory.

## Operator usage

Set the tenant default once:
```sql
UPDATE tenants SET settings = settings || '{"locale":"ko"}' WHERE slug='master';
```

Per-user override (e.g. one Vietnamese member in an otherwise Korean
tenant):
```sql
UPDATE tenant_users SET metadata = metadata || '{"locale":"vi"}' WHERE …;
```

## Test plan

- [ ] `go vet ./...`
- [ ] `go build -tags embedui ./...`
- [ ] `go test ./internal/i18n/...`
- [ ] With `tenants.settings.locale = "ko"`, force-run a cron via
      `cron.run` WS RPC and verify `cron_run_logs.summary` is in
      Korean.
- [ ] Force-run with `tenants.settings.locale` unset and confirm the
      summary stays in English (regression guard).
- [ ] Force-run with `tenant_users.metadata.locale = "vi"` for the
      cron's `user_id` and confirm the summary is in Vietnamese
      (override path).

### DIFF
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"strings"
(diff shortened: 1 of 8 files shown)


codebit0:
Closing — agent-level locale support is out of scope; data-layer (md context file) is the right place for this.

PR #1072, opened 2026-05-01
feat(acp): full Gemini CLI integration — tool exposure + spec compliance + watchdog + GEMINI.md

codebit0:
## Summary

Consolidated ACP-mode agent integration covering everything verified end-to-end against Gemini CLI 0.40.1: tool exposure, spec compliance, vendor default args, prompt watchdog, and GEMINI.md context. Supersedes #1069's ACP-architecture portions.

### Relationship to #1069

This PR carries the canonical ACP architecture going forward. #1069's ACP commits (parallel implementation in ACPSettings struct style) are superseded by this branch and will be closed; #1069's non-ACP parts (ko i18n, ops hardening, CLI header fix) are being split into focused follow-up PRs.

### 1. ACP agent tool exposure

- `internal/skills/dep_checker.go` — raise `depCheckTimeout` 5s→30s; on `DeadlineExceeded` log warning and assume present (was triggering reinstall loop every ~30s under load).
- `cmd/gateway_builtin_tools.go` — seed six tools that ACP-mode agents could not see (`datetime`, `delegate`, `mcp_tool_search`, `list_group_members`, `vault_read`, `vault_search`). Native-provider agents bypass this and were unaffected, masking the gap.
- `internal/mcp/bridge_server.go` — replace hardcoded `BridgeToolNames` whitelist with `BuiltinToolStore.ListEnabled()` lookup. UI builtin-tools toggle is now the single source of truth; `bridgeAlwaysExcluded` keeps internal-only tools (`spawn`, `create_forum_topic`, `heartbeat`) off the bridge.
- `cmd/gateway_setup.go` — always construct `mcp.Manager` so `mcp_tool_search` is available even with zero external MCP servers configured.

### 2. ACP spec compliance for Gemini CLI 0.40.1

- **Snake_case method aliases** — Gemini CLI 0.38.x+ sends spec-compliant `fs/read_text_file`, `fs/write_text_file`, `terminal/wait_for_exit`, `session/request_permission`. Each tool-bridge case now accepts both snake_case and the legacy camelCase. Without this, requests fall through to "unknown method" and Gemini's catch path stringifies the rejection (non-Error JS object) as `[object Object]` in `tool_call_update.content`.
[...]

PR #1069, opened 2026-04-30
feat: ACP standardization + ko i18n + ops hardening (full branch)

codebit0:
## Summary

Comprehensive branch covering ACP (Agent Client Protocol) provider work, Korean (ko) localization, and several operational fixes. Built on top of `origin/main` (merged in) so it reflects the full state we have been running internally.

This PR is broader in scope than #1067 (which extracts the ACP-only subset). Reviewers may prefer to land #1067 first and rebase this PR on top, or split it further.

#### 1. ACP protocol standardization and hardening
- Handle `session/request_permission` for Gemini CLI 0.38.x
- Prompt inactivity watchdog + stealth mode option
- Unified `ACPSettings` config struct shared by both registration paths (config.json `providers.acp` and DB `llm_providers.settings` JSONB)
- ACP provider hot-reload on DB updates (no gateway restart)
- `GEMINI.md`-based system prompt + context restoration on session reset
- ACP permission-mode wiring through tool bridge
- Unit tests for `extractACPContent` and `writeGeminiMD`

#### 2. Korean (ko) localization
- New `internal/i18n/catalog_ko.go` with full key coverage
- `LocaleKO` constant and `IsSupported` extended
- All UI namespaces translated under `ui/web/src/i18n/locales/ko/*.json`
- `SUPPORTED_LANGUAGES` and language label tables updated

#### 3. Provider verify (#1034 refactor)
- `model` is now required (ping-mode entrypoint kept as a fallback in this branch — see file diff)
- Response shape: `{success, models[]}` for richer feedback
- CLI: `goclaw providers verify <id>` prints the available model list

#### 4. Gemini native embedding provider
- New `internal/memory/embedding_gemini.go` (`memory.NewGeminiEmbeddingProvider`)
- `buildEmbeddingProvider` adds `ProviderGeminiNative` branch

#### 5. Operational fixes
- Bot token masking in channel logs (Telegram, etc.)
- Episodic memory worker timeout raised to 120s
- `path_escape` log corrections
- Channel watchdog retry on transient failures
- Missing `X-GoClaw-User-Id` header on gateway HTTP client
- Heartbeat FK migration (#1034) and tenant-scoped `DeleteProvider` retained from upstream main

#### 6. Docs / repo hygiene
- `GEMINI.md` moved to parent directory
- KIS integration guidelines added to `DEVNOTE.md`
- CLAUDE.md / DEVNOTE.md scoped out of earlier PRs (per repository convention)

### Out of scope / known caveats

[...]

PR #1078, opened 2026-05-02
fix(acp): tool exposure + spec compliance + vendor default args (consolidated)

codebit0:
## Summary

Consolidated branch covering ACP-mode agent fixes verified end-to-end against Gemini CLI 0.40.1. Combines the tool-exposure work from #1072 with the spec-compliant ACP standardization, in the same provider-layer architecture as #1069.

### Relationship to existing PRs

- **#1072 (fix/acp-tool-exposure)** — three commits here mirror its content (`fix(skills): prevent infinite pip install loop`, `fix(tools): seed missing builtin tools and wire mcp_tool_search`, `fix(mcp): expose builtin_tools-seeded tools via dynamic bridge lookup`). If this PR lands, #1072 can close as superseded.
- **#1069 (merge/2026-04-29)** — that PR carries the broader ACP standardization plus ko i18n + ops hardening. The ACP-architecture parts here are equivalent in final shape (provider-layer `WithIncludeDirectories` option + `applyVendorDefaultArgs` helper); ko i18n and other ops items in #1069 are not duplicated and would still need to land separately.

Reviewers may prefer to merge #1069 first and rebase this on top, or pick one branch to consolidate everything.

### 1. ACP agent tool exposure (from #1072)

- `internal/skills/dep_checker.go` — raise `depCheckTimeout` 5s→30s; on `DeadlineExceeded` log a warning and assume packages present (the previous behaviour falsely flagged everything missing under load and triggered a reinstall loop every ~30s).
- `cmd/gateway_builtin_tools.go` — seed six tools that ACP-mode agents could not see (`datetime`, `delegate`, `mcp_tool_search`, `list_group_members`, `vault_read`, `vault_search`). Native-provider agents read the registry directly and were unaffected, masking the gap.
- `internal/mcp/bridge_server.go` — replace the hardcoded `BridgeToolNames` whitelist with `BuiltinToolStore.ListEnabled()` lookup. The UI builtin-tools toggle is now the single source of truth for ACP exposure; `bridgeAlwaysExcluded` keeps internal-only tools (`spawn`, `create_forum_topic`, `heartbeat`) off the bridge.
- `cmd/gateway_setup.go` — always construct an `mcp.Manager` so `mcp_tool_search` is available even when no external MCP servers are configured (returns "no tools found" until servers are added; no redeploy on add).

### 2. ACP spec compliance for Gemini CLI 0.40.1

- **Snake_case method aliases** — Gemini CLI 0.38.x+ sends spec-compliant `fs/read_text_file`, `fs/write_text_file`, `terminal/wait_for_exit`, `session/request_permission`. Each tool-bridge switch case now accepts both snake_case and the legacy camelCase. Without this, requests fall through to "unknown method", which Gemini's catch path stringifies as `[object Object]` in `tool_call_update.content` because the rejection isn't a JS `Error` instance.
- **Kind-based permission selection** — new `SessionRequestPermissionRequest`/`Response` types match the spec's nested-outcome shape (`{outcome: {outcome: "selected", optionId}}` / `{outcome: {outcome: "cancelled"}}`). `handleSessionPermission` selects by `PermissionOption.Kind` (`allow_once` / `allow_always` / `reject_once` / `reject_always`) rather than agent-defined OptionID strings, since Kind is stable across Gemini versions while OptionID values are not. Approve-all prefers `allow_once` so operator policy still gates each tool call.

### 3. Provider-layer vendor and include-directories injection

Removes the cmd-level `enrichGeminiACPArgs` in favour of two cleaner primitives in the providers package:

- `WithIncludeDirectories(dirs []string) ACPOption` — callers pass candidate skill directories; `NewACPProvider` stat-filters and emits `--include-directories <dir>` pairs only for the gemini binary.
- `applyVendorDefaultArgs(binary, args)` — appends per-binary CLI defaults that goclaw's deployment requires unconditionally. Currently injects `--skip-trust` for gemini so MCP discovery runs even when the per-session cwd inherits `DO_NOT_TRUST` from `~/.gemini/trustedFolders.json`. ACP sessions always run inside a goclaw-managed sandbox, so the user-facing trust gate is moot. New vendor rules go here rather than scattering binary-name checks across call sites.

`cmd/gateway_providers.go` now only enumerates candidate skill directories via `acpSkillCandidateDirs(workspace)` and passes them as the `WithIncludeDirectories` option; binary gating and vendor flag emission live in the provider.

## Test plan

- [x] `go build ./...` (PG)
- [x] `go build -tags sqliteonly ./...` (Desktop / SQLite)
- [x] `go vet ./...`
- [x] `go test ./internal/providers/acp/...`
- [x] `go test ./internal/skills/...`
- [x] Live verified against Gemini CLI 0.40.1: ACP `session/new` → `session/prompt` → `session/request_permission` (Kind-based, approved with `allow_once`) → MCP `tools/call` for `datetime` → tool result text returned to LLM → final user-facing message contains the correct timestamp. No `[object Object]` content in `tool_call_update`.
- [x] `mcp.bridge: tools registered count=37 skipped_excluded=1` — the seeded tools appear in `tools/list` results.
- [ ] Reviewer to run integration tests with pgvector pg18 on port 5433.
- [ ] Reviewer to verify the dep-check loop is gone under skill install load.

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -286,6 +286,7 @@ func runGateway() {
 	server.SetDB(pgStores.DB)
 	server.SetPolicyEngine(permPE)
 	server.SetPairingService(pgStores.Pairing)
+	server.SetBuiltinToolsStore(pgStores.BuiltinTools)
 	server.SetMessageBus(msgBus)
 	server.SetOAuthHandler(httpapi.NewOAuthHandler(pgStores.Providers, pgStores.ConfigSecrets, providerRegistry, msgBus))
 
--- a/cmd/gateway_builtin_tools.go
+++ b/cmd/gateway_builtin_tools.go
@@ -112,6 +112,26 @@ func builtinToolSeedData() []store.BuiltinToolDef {
 		{Name: "team_tasks", DisplayName: "Team Tasks", Description: "View, create, update, and complete tasks on the team task board", Category: "teams", Enabled: true,
 			Requires: []string{"managed_mode", "teams"},
 		},
+
+		// general — utility tools available to every agent (incl. ACP via MCP bridge)
+		{Name: "datetime", DisplayName: "Date / Time", Description: "Get current date and time with optional IANA timezone (returns UTC + local).", Category: "general", Enabled: true},
+
+		// subagents (continued)
+		{Name: "delegate", DisplayName: "Delegate Task", Description: "Delegate a task to another linked agent (inter-agent orchestration).", Category: "subagents", Enabled: true},
+
+		// skills (continued)
+		{Name: "mcp_tool_search", DisplayName: "MCP Tool Search", Description: "BM25 search over deferred MCP tool catalog to discover relevant tools by query.", Category: "skills", Enabled: true},
+
+		// messaging (continued)
+		{Name: "list_group_members", DisplayName: "List Group Members", Description: "List members in a group chat (Telegram group, etc.).", Category: "messaging", Enabled: true},
+
+		// memory — Knowledge Vault
+		{Name: "vault_read", DisplayName: "Vault Read", Description: "Read a document from the Knowledge Vault by path or [[wikilink]].", Category: "memory", Enabled: true,
+			Requires: []string{"vault"},
+		},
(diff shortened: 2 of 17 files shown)


codebit0:
Superseded — folding the same content (consolidated ACP work) directly into #1072 instead of opening a parallel PR.

PR #1088, opened 2026-05-04
fix(teams): propagate task result in team task completion event

khanhtranhd:
## Summary

- Add `Result string` field to `TeamTaskEventPayload` protocol struct (populated on `team.task.completed` events)
- Add `WithResult(r string)` option helper in `team_event_helpers.go`
- Pass `WithResult(result)` when broadcasting `EventTeamTaskCompleted` in `executeComplete()`
- Update completion notification format to include result summary (truncated at 300 chars)

## Problem

When a member agent completes a task, the `EventTeamTaskCompleted` bus event carried no result data. Downstream subscribers (notification, audit) had no way to surface the completion output without re-querying the task store.

## Solution

Populate `Result` directly in the broadcast payload at completion time. Two consumer paths benefit:

1. **External notify** — completion message now includes the result snippet instead of just `"✅ Task #N completed"`
2. **Bus subscribers** — can read `payload.Result` without an extra DB round-trip

## Test plan

- [ ] Member agent completes a task with a non-empty result string
- [ ] Verify `EventTeamTaskCompleted` payload carries the result
- [ ] Verify notification includes result snippet (≤300 chars) followed by `…` if truncated
- [ ] Verify empty result still produces the original `"✅ Task #N completed"` format (no trailing newline)

### DIFF
--- a/cmd/gateway_events.go
+++ b/cmd/gateway_events.go
@@ -214,7 +214,15 @@ func (d *gatewayDeps) wireTeamProgressNotifySubscriber() {
 		case protocol.EventTeamTaskAssigned:
 			content = fmt.Sprintf("📋 Task #%d \"%s\" → assigned to %s", payload.TaskNumber, payload.Subject, agentName)
 		case protocol.EventTeamTaskCompleted:
-			content = fmt.Sprintf("✅ Task #%d \"%s\" completed", payload.TaskNumber, payload.Subject)
+			if payload.Result != "" {
+				result := payload.Result
+				if len(result) > 300 {
+					result = result[:300] + "…"
+				}
+				content = fmt.Sprintf("✅ Task #%d \"%s\" completed\n%s", payload.TaskNumber, payload.Subject, result)
+			} else {
+				content = fmt.Sprintf("✅ Task #%d \"%s\" completed", payload.TaskNumber, payload.Subject)
+			}
 		case protocol.EventTeamTaskProgress:
 			if payload.ProgressStep != "" {
 				content = fmt.Sprintf("⏳ Task #%d \"%s\": %d%% — %s", payload.TaskNumber, payload.Subject, payload.ProgressPercent, payload.ProgressStep)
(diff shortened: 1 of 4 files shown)


mrgoonie (changes requested):
Summary: This adds the completion result to the explicit `team_tasks(action="complete")` event path, but it misses the post-turn auto-complete path that emits the same `team.task.completed` event.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: clear
- Project standards: passed
- Strategic necessity: clear value
- CI/checks: green

Findings:
- Important: `cmd/gateway_consumer_post_turn.go` still broadcasts `EventTeamTaskCompleted` without the computed `result` after auto-completing a member run. That path is explicitly used when an agent turn ends without a terminal tool action, and it stores the result via `CompleteTask(ctx, meta.TaskID, meta.TeamID, result)` immediately before broadcasting. With the current diff, explicit `team_tasks complete` notifications include `payload.Result`, but auto-completed tasks still emit the old payload with no result, so downstream notification/audit subscribers remain inconsistent and still need a DB re-query for a common completion path. Please pass the same result through `tools.WithResult(result)` in the post-turn auto-complete broadcast.
- Suggestion: Add `result?: string` to `ui/web/src/types/team-events.ts` and update `docs/13-ws-team-events.md` so the public event contract matches the new payload field.

Verdict: REQUEST_CHANGES

clark-cant:
Maintainer nudge (2026-06-26): 1 day since review. The finding is specific — `cmd/gateway_consumer_post_turn.go` broadcasts `EventTeamTaskCompleted` without `tools.WithResult(result)` in the auto-complete path, while the explicit `team_tasks complete` path already includes it. Fix scope: pass `tools.WithResult(result)` into the post-turn auto-complete broadcast, then update `ui/web/src/types/team-events.ts` and `docs/13-ws-team-events.md` to document the new `result` field. No duplicate PR found. This is mergeable once the finding is addressed.

PR #1101, opened 2026-05-05
fix(agent): add Thinking field to synthetic post-summary assistant message

### LINKED ISSUE #1047: [Bug] DeepSeek V4 Pro: Lỗi 400 do thiếu reasoning_content trên kênh Zalo (OpenAI-compatible)

hoakhongmau98:
## Summary

After session compaction, the agent loop injects a synthetic assistant ack ("I understand the context from our previous conversation. How can I help you?") into history but leaves `Thinking` empty. For thinking-mode providers (DeepSeek v4-pro, Kimi), the OpenAI-compat request builder requires `reasoning_content` on every assistant turn — the empty `Thinking` causes HTTP 400 `reasoning_content in the thinking mode must be passed back to the API` on the very next turn after compaction.

This PR populates `Thinking` on the synthetic message so the existing wire logic in `internal/providers/openai_request.go` emits a non-empty `reasoning_content` for DeepSeek/Kimi families.

Fixes #1047.

## Root cause

- `internal/agent/loop_history.go:249-259` appends a synthetic assistant turn after the user's post-summary "Got it." message. Before this change, only `Role` and `Content` were set.
- `internal/providers/openai_request.go:59-61` writes `reasoning_content = m.Thinking` for assistant turns when `openAIWireAssistantReasoningContent(model)` returns true (DeepSeek, Kimi, OpenAI reasoning families). With `Thinking == ""`, the field is omitted and DeepSeek rejects the request.

## Change

Single 4-line diff in `internal/agent/loop_history.go` — add a `Thinking` value to the synthetic message. No provider-layer changes, no new helpers, no new tests required (existing `internal/providers/openai_test.go:336` already asserts that DeepSeek wires `reasoning_content` from `Thinking`).

## Relation to #1063

PR #1063 (`nagaame:dev`) addresses a related-but-distinct symptom: it ensures `reasoning_content` is preserved across tool-call turns and injects an empty string for DeepSeek when `Thinking` is absent in the provider layer. The two fixes are complementary:

- **#1063** fixes the tool-calling history path and adds defensive provider-layer scaffolding.
- **This PR** fixes the synthetic post-compaction ack — a code path #1063 does not touch.

Either PR alone leaves the other failure mode open. They can be merged independently; this one is the minimal, scoped fix for the post-summary case.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test -race ./internal/providers/...` passes (existing DeepSeek wire test covers the path)
- [ ] Manual verification on a session that triggered compaction with DeepSeek v4-pro

### DIFF
--- a/internal/agent/loop_history.go
+++ b/internal/agent/loop_history.go
@@ -255,6 +255,10 @@ func (l *Loop) buildMessages(ctx context.Context, history []providers.Message, s
 		messages = append(messages, providers.Message{
 			Role:    "assistant",
 			Content: "I understand the context from our previous conversation. How can I help you?",
+			// Synthetic reasoning for thinking-mode providers (DeepSeek v4-pro, Kimi)
+			// that reject assistant turns missing reasoning_content. Wired in
+			// internal/providers/openai_request.go via openAIWireAssistantReasoningContent.
+			Thinking: "Reviewed previous conversation summary. Ready to continue from this context.",
 		})
 	}
 
--- a/internal/agent/loop_pipeline_adapter.go
+++ b/internal/agent/loop_pipeline_adapter.go
@@ -36,7 +36,11 @@ func (l *Loop) runViaPipeline(ctx context.Context, req RunRequest) (*RunResult, 
 	if err != nil {
 		return nil, err
 	}
-	return convertRunResult(pResult), nil
+	result := convertRunResult(pResult)
+	// Bridge Thinking into runState so finalizeRun emits a non-empty Thinking
+	// on the assistant message — required for DeepSeek/Kimi reasoning_content.
+	bridgeRS.finalThinking = result.Thinking
+	return result, nil
 }
 
 // buildPipelineDeps maps Loop fields + methods to PipelineDeps callbacks.


hoakhongmau98:
## Update: second fix appended

Trace `019df751-908e-70f3-b7d9-f3617e81e8ca` confirms the original 4-line fix on `loop_history.go` is necessary but **not sufficient** — 3/3 sub-agents spawned via the `batch-research` skill still failed at iteration 2 with the same HTTP 400.

### Root cause of the residual failure

The synthetic post-summary fix only covers history rebuilt after compaction. For sub-agents on a fresh session, the failure is on **iteration 2** itself: the assistant's *real* (LLM-generated) thinking from iteration 1 never reaches the next request because `runState.finalThinking` is never populated from the v3 pipeline result.

Flow:

1. `internal/providers/openai_http.go:72` — `result.Thinking = msg.ReasoningContent` ✅
2. `internal/pipeline/observe_stage.go:45` — `state.Observe.FinalThinking = resp.Thinking` ✅
3. `internal/pipeline/run_state.go:59` — `RunResult.Thinking: rs.Observe.FinalThinking` ✅
4. `internal/agent/loop_pipeline_adapter.go:254` (`convertRunResult`) — `Thinking: pr.Thinking` ✅
5. **But** `bridgeRS.finalThinking` is never assigned from the pipeline result ❌
6. `internal/agent/loop_finalize.go:106` reads `rs.finalThinking` → always `""` → `reasoning_content` missing → DeepSeek 400

### Change in commit 2138a32b

One-line bridge in `runViaPipeline` (`internal/agent/loop_pipeline_adapter.go`):

```go
result := convertRunResult(pResult)
bridgeRS.finalThinking = result.Thinking
return result, nil
```

`bridgeRS` is the same `runState` instance that `finalizeRun` later reads when constructing the assistant message persisted to history. Backward compatible — non-thinking providers (Anthropic, OpenAI non-reasoning) ignore `Thinking`.

### Verification
- [x] `go build ./...` passes
- [x] `go vet ./internal/agent/...` passes

Manual reproduction on the cron-triggered sub-agent path is still pending.

thieung:
Commit 1 (`c7def3ce`) looks correct and targets the right root cause. However, commit 2 (`2138a32b`) appears to be a **no-op** in the current code path — please double-check before merging.

### Issue in `internal/agent/loop_pipeline_adapter.go:42`

```go
bridgeRS.finalThinking = result.Thinking
return result, nil
```

1. `bridgeRS` is a local variable and goes out of scope immediately after `return`.
2. `syncBridgeToState` (`loop_pipeline_tool_callbacks.go:167`) — the only function that propagates `bridgeRS` into the pipeline `RunState` — does **not** copy `finalThinking`. It only copies `loopKilled / asyncToolCalls / deliverables / finalContent / bootstrap+team flags`. And it only runs *while* `p.Run()` is still executing, not after.
3. `Loop.finalizeRun` (referenced by the commit message at `loop_finalize.go:38`) has **no callers** anywhere in the repo (`grep '\.finalizeRun('` returns nothing). The only run path in `loop_run.go:168` is `runViaPipeline`. That function is dead v2 legacy code — it never executes.
4. The actual source of `Thinking` on the assistant message flushed to the session is `state.Observe.FinalThinking`, set at `observe_stage.go:45` from `resp.Thinking` and read at `finalize_stage.go:88`. This flow is entirely independent of `bridgeRS`.

### Suggested options

- **(a)** Drop commit 2 — keep only commit 1, which is the correct fix for the synthetic post-summary message.
- **(b)** If the intent was also to cover the case where the LLM itself returns `Thinking == \"\"` on a real assistant turn (not the synthetic one), the fix belongs in `ObserveStage`: when `resp.Thinking == \"\"` and the model is DeepSeek/Kimi, set a placeholder into `state.Observe.FinalThinking` so `FinalizeStage` persists the assistant message with a non-empty `reasoning_content`. That is the correct pipeline boundary and would protect every subsequent turn. (This may overlap in scope with #1063 — worth coordinating.)

### Bonus

Consider adding a small agent-level test asserting that `buildMessages` injects the synthetic assistant message with a non-empty `Thinking` whenever `summary != \"\"` — locks the invariant from commit 1 in place.

Verified locally on the PR branch:
- `go build ./...` ✓
- `go vet ./...` ✓
- `go test -race ./internal/providers/...` ✓

mrgoonie (changes requested):
Summary: This is a valid, small fix direction for the post-compaction synthetic assistant message, but the current PR still includes a dead-path bridge change and lacks a regression test for the actual invariant.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: related/overlap with #1063, but not a direct duplicate for the synthetic post-summary ack path.
- Project standards: issue found — one commit changes dead v2 bridge state and the core compaction path has no regression test.
- Strategic necessity: clear value; DeepSeek/Kimi reasoning-mode sessions should survive compaction without dropping required assistant reasoning content.
- CI/checks: GitHub checks missing/no check rollup reported; local targeted `go test ./internal/agent/... ./internal/pipeline/... ./internal/providers/...` passes.

Findings:
- Important: The `internal/agent/loop_pipeline_adapter.go` change is not effective in the current pipeline path. `bridgeRS.finalThinking` is assigned after `p.Run` returns, but `finalize_stage.go` persists the final assistant message from `state.Observe.FinalThinking`; `syncBridgeToState` does not copy `finalThinking`, and `Loop.finalizeRun` has no callers in the current run path. Please drop this commit or move any needed fallback into the actual pipeline boundary (`ObserveStage`/`FinalizeStage`) with tests.
- Important: The real `loop_history.go` fix needs a regression test. Please add an agent/history test proving that when a summary is injected, the synthetic assistant message has non-empty `Thinking`, so OpenAI-compatible thinking providers can emit `reasoning_content` on the next request. This is the invariant this PR is meant to protect.

Verdict: REQUEST_CHANGES

Next step: keep the minimal `loop_history.go` synthetic-ack fix, remove the dead-path `loop_pipeline_adapter.go` bridge change unless you can prove a live caller, and add the compaction-history regression test before re-review.

*Posted by /github-maintain at 2026-06-23T23:16:34Z*

PR #1103, opened 2026-05-05
fix(skills): normalize backslash in ZIP entry names for Windows compatibility

buidackim:
## Problem

ZIP archives created on Windows may use backslash (`\`) as path separator instead of forward slash (`/`) in entry names. When these archives are extracted on Linux, the backslash is treated as a literal character in the filename, causing files like `scripts\search.py` to be created as a **single flat file** instead of the proper `scripts/search.py` directory structure.

This breaks skill execution when skills contain subdirectories (e.g. `scripts/`, `references/`).

## Root Cause

The [ZIP specification (PKZIP APPNOTE)](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) requires forward slashes, but many Windows tools (Python `zipfile`, .NET `ZipArchive`, older archivers) produce non-conforming archives with backslashes.

The current extraction logic in `skills_upload.go` uses `filepath.Clean()` which does **not** convert backslashes to forward slashes on Linux (since `\` is a valid filename character on Linux).

## Fix

Added `strings.ReplaceAll(entryName, \, /)` normalization before `filepath.Clean()` during skill ZIP extraction. This is a single-line defensive fix that handles the edge case without affecting correctly-formed archives.

## Testing

- Uploaded a Windows-created ZIP containing `scripts\search.py` entry
- Before fix: file created as `scripts\search.py` (flat) → skill execution fails
- After fix: file correctly extracted as `scripts/search.py` (subdirectory) → skill works

### DIFF
--- a/internal/http/skills_upload.go
+++ b/internal/http/skills_upload.go
@@ -199,15 +199,16 @@ func (h *SkillsHandler) handleUpload(w http.ResponseWriter, r *http.Request) {
 				continue
 			}
 		}
-		// Skip macOS/system artifacts
-		if skills.IsSystemArtifact(entryName) {
-			continue
-		}
 		// Normalize Windows backslash separators in ZIP entries.
 		// ZIPs created on Windows may use `\` instead of `/` in paths,
 		// causing files like "scripts\search.py" to be extracted as a
 		// single flat file instead of scripts/search.py on Linux.
 		entryName = strings.ReplaceAll(entryName, "\\", "/")
+		// Skip macOS/system artifacts after separator normalization so
+		// entries like "__MACOSX\\._SKILL.md" are filtered too.
+		if skills.IsSystemArtifact(entryName) {
+			continue
+		}
 		// Security: prevent path traversal
 		name := filepath.Clean(entryName)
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Reviewed the current diff and pushed one follow-up commit to cover the behavior with a regression test.

Summary: this is a small, targeted fix for non-conforming Windows ZIP entries that use backslash separators. Current upload extraction used filepath.Clean on the raw ZIP name, which preserves backslash as a literal filename on Linux; normalizing to / before traversal checks makes scripts/search.py extract as a directory path instead of a flat scripts\search.py file.

Gates checked:
- Duplicate/prior work: no existing issue/PR found for skill ZIP backslash extraction.
- Project standards: change stays inside the existing skill upload path and preserves existing traversal checks.
- Strategic value: clear DX/compatibility value for uploaded skills packaged on Windows.

Validation:
- Added TestHandleUpload_NormalizesBackslashZipEntryPaths.
- Ran: go test ./internal/http -run TestHandleUpload_NormalizesBackslashZipEntryPaths

Risk: low. Approving.

PR #1104, opened 2026-05-05
fix(skills): instruct agent to use exact absolute path for skill files

buidackim:
## Problem

The system prompt contains conflicting guidance that causes weaker models to mangle skill file paths:

1. **Skills section** says: *read its SKILL.md at the location path with read_file*
2. **Workspace section** says: *Use relative paths — do not guess absolute paths*

Models like MiniMax-M2.7 try to reconcile both instructions by stripping the absolute prefix from the skill path (e.g. `/app/data/skills-store/gpt-image-2-pro-max/1/SKILL.md` → `skills/gpt-image-2-pro-max-main/SKILL.md`), causing `file not found` errors.

## Fix

Resolves the contradiction by clarifying the boundary:

- **Workspace section**: relative paths for *user files*, exact absolute paths for *skills*
- **Skills section** (inline + search modes): adds `(use the EXACT absolute path provided)`
- **skill_search result**: adds the same emphasis in the ACTION REQUIRED instruction

## Files Changed

- `internal/agent/systemprompt.go` — 3 string changes in prompt guidance
- `internal/tools/skill_search.go` — 1 string change in search result instruction

## Impact

Minimal prompt size increase (~30 tokens). Helps weaker models follow skill paths correctly without affecting stronger models.

### DIFF
--- a/internal/agent/systemprompt.go
+++ b/internal/agent/systemprompt.go
@@ -640,7 +640,7 @@ func buildSkillsSection(skillsSummary string, hasSkillSearch, hasSkillManage boo
 			"## Skills (mandatory)",
 			"",
 			"Before replying, scan `<available_skills>` below.",
-			"If a skill clearly applies, read its SKILL.md at the `<location>` path with `read_file`, then follow it.",
+			"If a skill clearly applies, read its SKILL.md at the `<location>` path with `read_file` (use the EXACT absolute path provided), then follow it.",
 			"If multiple could apply, choose the most specific one. Never read more than one skill up front.",
 			"If none apply, proceed normally.",
 			"",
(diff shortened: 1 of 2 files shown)


mrgoonie (changes requested):
Summary: This is the right problem to fix, but the patch currently breaks the prompt-size test budget in `internal/agent`.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; I only found this PR for the exact skill absolute-path prompt guidance.
- Project standards: issue found — prompt wording changes need to stay within the tested mode size budgets.
- Strategic necessity: clear value; the workspace-vs-skill path ambiguity is real and worth removing.
- CI/checks: missing on GitHub; local targeted test failed.

Findings:
- Important: `go test ./internal/agent ./internal/tools` fails on this branch with `TestNoneModeSections`: `none mode too large: 3112 chars`. The added wording is small, but it still pushes an existing prompt-size guard over its limit. Please either tighten the new wording enough to keep the none-mode budget under the test threshold, or update the test budget only if that larger prompt is an intentional product decision.

Verdict: REQUEST_CHANGES

Next step: make the wording more compact (for example, avoid repeating the full “use the EXACT absolute path provided” phrase in multiple emitted sections) and rerun `go test ./internal/agent ./internal/tools` before re-review.

*Posted by /github-maintain at 2026-06-22T21:09:30Z*

Issue #915, opened 2026-04-15
Bug - Telegram group: `write_file` checks writer permission against `group:telegram:<id>` instead of sender user


rongxanh:
## Summary
In Telegram **group chats**, `write_file` is being authorized with:

- `userId = group:telegram:<group_id>`

instead of the actual sender's Telegram user ID.

This breaks file-writer authorization in groups and causes `write_file` to fail before file creation.

---

## Affected versions
Reproduced on:

- `v3.6.0`
- `v3.7.1`

---

## Expected
For Telegram group messages:

- group should remain the **session/scope**
- sender should remain the **acting user**

So file-writer checks should use:

- scope = `group:telegram:<group_id>`
- user = `<sender_user_id>`

---

## Actual
`write_file` fails with:
`permission denied: only file writers can modify files in this group. Use /addwriter to get write access`

Runtime/tool context shows the wrong principal:

```
{
  "userId": "group:telegram:<groupid>",
  "channel": "telegram",
  "sessionKey": "agent:<agent>:telegram:group:<groupid>"
}
```
This suggests group runs are assigning `user_id` to the **group principal** instead of the sender user.

---

## Minimal evidence
```
[INFO] inbound: scheduling message (main lane) ... peerkind=group ... userid=group:telegram:<groupid>
[INFO] tool call ... tool=writefile
[WARN] tool error ... permission denied: only file writers can modify files in this group ...
Tool result:json
{
  "type": "tool.result",
  "payload": {
    "name": "writefile",
    "iserror": true,
    "result": "permission denied: only file writers can modify files in this group. Use /addwriter to get write access"
  },
  "userId": "group:telegram:<group_id>",
  "channel": "telegram"
}
```
---

## Likely root cause
Telegram group inbound/runtime context appears to collapse:

- **group scope/session**
- **acting user**

into the same principal.

`commands_writers.go` seems to treat:
- group = scope
- sender = actor

But the tool runtime appears to treat:
- group = scope
- group = actor

---

## Suggested fix
For Telegram group runs, keep these separate:

- session/scope → `group:telegram:<group_id>`
- acting user → actual Telegram sender user ID

Then run `write_file` auth against the sender user within the group scope.

viettranx:
[...]

PR #1105, opened 2026-05-06
fix(perms): propagate sender/role through MCP bridge & delegate (#915)

mozaa-solana:
## Summary

Group-scoped permission checks (`CheckCronPermission` / `CheckFileWriterPermission`) deny with `senderID=""` when tools are invoked through:

1. **MCP bridge** (external CLI providers like `claude-cli`, `opencode-go`) — the bridge middleware never injected `SenderID`/`Role` into ctx, so even though the agent loop sets them on `RunRequest`, the bridge layer dropped attribution. Result: any group cron / file_writer mutation called via the bridge denies as "system context".
2. **`delegate` tool** (`gateway_managed.go` `delegateRunFn`) — built `RunRequest` without copying `SenderID`/`Role`/`Channel`/`ChatID`/`PeerKind` from `DelegateRequest`, so the target agent's `loop_context.injectContext` skipped `WithSenderID` and downstream group cron/file_writer mutations denied with `senderID=""`.

## Changes

- **`internal/providers/claude_cli.go`** — new `OptSenderID`, `OptRole`.
- **`internal/providers/claude_cli_mcp.go`** — `BridgeContext` carries `SenderID`+`Role`; `writeMCPConfig` emits `X-Sender-ID`/`X-Role` headers and signs them in the HMAC payload (extras: `localKey | sessionKey | senderID | role`). `VerifyBridgeContext` gains a backward-compat fallback that drops the last 2 extras for sessions written pre-this-change.
- **`internal/providers/claude_cli_session.go`** — `bridgeContextFromOpts` reads new opts.
- **`internal/gateway/server.go`** (`bridgeContextMiddleware`) — read `X-Sender-ID`+`X-Role` headers (HMAC-protected), inject `WithSenderID`/`WithRole` into ctx.
- **`internal/agent/loop_pipeline_callbacks.go`** — pass `req.SenderID`/`req.Role` into `chatReq.Options`.
- **`cmd/gateway_managed.go`** (`delegateRunFn`) — propagate `SenderID`/`Role`/`Channel`/`ChatID`/`PeerKind` from `DelegateRequest` into `RunRequest`.
- **`internal/tools/team_tool_dispatch.go`** + **`cmd/gateway_consumer_normal.go`** — `slog.Warn` when sender is empty in a group-target dispatch so admins can find upstream callers losing attribution.
- **`internal/store/config_permission_store.go`** — surface diagnostic-rich error messages (which user/scope/configType denied) instead of generic "permission denied".
- **`internal/tools/cron.go`** — forward the diagnostic message to LLM result so the agent can self-correct.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/providers/... ./internal/store/... ./internal/gateway/... ./internal/agent/...` passes (incl. `TestVerifyBridgeContext` HMAC fallbacks)
- [x] Reproduced the cron-in-Telegram-group denial, applied this fix, retested, cron now works for whitelisted users
- Backward compatibility verified: pre-existing MCP config sessions (without senderID/role headers) still match the older HMAC fallback in `VerifyBridgeContext`.

## Related

Follow-up to #915. Diagnostic logs added in this PR are the ones I used to localize the root cause; happy to drop them if reviewers prefer.

### DIFF
--- a/cmd/gateway_consumer_normal.go
+++ b/cmd/gateway_consumer_normal.go
@@ -358,20 +358,39 @@ func processNormalMessage(
 
 	// Resolve effective sender: prefer MetaOriginSenderID when the on-wire
 	// SenderID is an internal/synthetic one (e.g. "notification:progress",
-	// "ticker:system", "system:escalation", "session_send_tool"). This lets
+	// "ticker:system", "system:escalation", "session_send_tool") OR is
+	// empty (some publish paths drop SenderID even though the originating
+	// turn carried a real user via MetaOriginSenderID). This lets
 	// system-initiated turns that DO have a real user behind them (because
 	// they propagated the origin via metadata) attribute actions to that user
-	// — e.g. for CheckFileWriterPermission in group chats (#915). Synthetic
-	// senders without propagation keep their on-wire value and hit F1's
-	// deny-in-group rule (safe default).
+	// — e.g. for CheckFileWriterPermission / CheckCronPermission in group
+	// chats (#915). Synthetic / empty senders without propagation keep their
+	// on-wire value and hit F1's deny-in-group rule (safe default).
 	effectiveSenderID := msg.SenderID
-	if bus.IsInternalSender(effectiveSenderID) {
+	if effectiveSenderID == "" || bus.IsInternalSender(effectiveSenderID) {
 		// Defense-in-depth: if a propagation bug ever writes a synthetic
 		// value into MetaOriginSenderID, do NOT honour it. We want only real
-		// user senders to override the on-wire synthetic.
+		// user senders to override the on-wire synthetic/empty.
 		if realSender := msg.Metadata[tools.MetaOriginSenderID]; realSender != "" && !bus.IsInternalSender(realSender) {
 			effectiveSenderID = realSender
 		}
+	}
+	// Diagnostic: warn when we end up with an empty sender for a group
+	// context. This is almost always a propagation bug somewhere upstream
+	// (channel handler, re-ingress publisher, or skill that lost ctx).
+	// CheckFileWriterPermission / CheckCronPermission will deny this turn.
+	if effectiveSenderID == "" && peerKind == string(sessions.PeerGroup) {
+		mk := make([]string, 0, len(msg.Metadata))
+		for k := range msg.Metadata {
+			mk = append(mk, k)
+		}
+		slog.Warn("group inbound has empty SenderID — group-scoped permission checks will deny",
+			"channel", msg.Channel,
+			"chat_id", msg.ChatID,
+			"on_wire_sender", msg.SenderID,
+			"meta_origin_sender", msg.Metadata[tools.MetaOriginSenderID],
+			"meta_keys", mk,
+		)
 	}
 	// Role propagation: carry the RBAC role of the originating actor so
 	// permission checks during the re-ingress turn can bypass per-user
--- a/cmd/gateway_managed.go
+++ b/cmd/gateway_managed.go
@@ -420,10 +420,25 @@ func wireExtras(
 				SessionKey:    sessionKey,
 				Message:       req.Task,
 				UserID:        req.UserID,
-				Channel:       "delegate",
+				// Propagate real acting sender + RBAC role through delegate so that
+				// downstream group-scoped permission checks (CheckCronPermission /
+				// CheckFileWriterPermission) attribute to the original user instead
+				// of an empty principal. Without these two lines, any cron/file
+				// mutation inside a delegated turn in a group context fails with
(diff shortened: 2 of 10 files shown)


clark-cant:
## PR Review

Verdict: REQUEST CHANGES

### Summary
This PR fixes a real attribution bug: group-scoped permission checks lose the acting sender/role when tool calls go through the MCP bridge or delegate path, causing cron/file-writer mutations to deny with senderID empty.

The intent is good, but the MCP bridge HMAC fallback currently allows unsigned X-Sender-ID / X-Role headers to be trusted for old-session signatures. That is a security issue because these headers power group-scoped permission checks.

Risk level: high, because this touches permission context propagation across the bridge.

### Findings

1. [HIGH] internal/gateway/server.go:246-282 + internal/providers/claude_cli_mcp.go:302-309 — fallback verification can trust unsigned sender/role headers.

bridgeContextMiddleware always passes localKey, sessionKey, senderID, roleHdr into VerifyBridgeContext, then injects senderID and roleHdr whenever ok == true.

But VerifyBridgeContext has a backward-compat fallback that drops the last 2 extras:

preSender := SignBridgeContext(..., extra[:len(extra)-2]...)
if hmac.Equal(...) { return true, true }

That means a request with a valid pre-sender/role signature over only localKey|sessionKey can still include arbitrary X-Sender-ID / X-Role headers and pass verification via fallback. The middleware then injects those unsigned values into context.

Impact: any bridge caller using an older MCP config signature could forge the sender or role headers for group-scoped permission checks. Since this PR specifically uses sender/role for cron/file-writer permission, these fields must not be trusted unless they were covered by the matched HMAC format.

Suggested fix: have VerifyBridgeContext return an additional signal for which context fields were covered, e.g. senderRoleVerified bool, or avoid injecting sender/role unless the full signature including senderID/role matched.

Expected behavior:
- Current full match over localKey|sessionKey|senderID|role: ok=true, tenantVerified=true, senderRoleVerified=true
- Pre-sender fallback over localKey|sessionKey: ok=true, tenantVerified=true, senderRoleVerified=false
- Middleware only calls WithSenderID / WithRole when senderRoleVerified == true

Please add a regression test: sign with old extras only (localKey, sessionKey), send non-empty senderID/role, assert bridge middleware does not inject them, or assert VerifyBridgeContext reports they are not verified.

2. [MEDIUM] internal/providers/claude_cli_mcp_test.go — missing tests for new sender/role HMAC behavior.

Existing tests cover localKey/sessionKey extra params, but this PR adds security-sensitive fields and a fallback mode. Please add explicit tests for:

- current signature including senderID/role validates
- tampering senderID or role fails full verification
- old pre-sender signature remains accepted for backward compatibility but does not authorize sender/role injection

### Anti-AI-Slop Check
- Intent clear: yes — concrete group permission attribution bug and reproduction.
- Diff scope justified: mostly yes, but permission context propagation needs sharper verification semantics.
- Refactor justified: not applicable.

### Tests / CI
- Local targeted check: go test ./internal/providers ./internal/gateway ./internal/store ./internal/tools ✅
- PR status checks: none reported at review time.
- Test gap: no coverage proving sender/role cannot be spoofed through fallback signatures.

### Final Notes
The delegate propagation side looks reasonable. The bridge side needs one more security boundary: do not trust sender/role unless the matched HMAC actually covered sender/role. Fix that and add regression tests, then this should be in good shape.

mrgoonie (changes requested):
Maintainer recheck: still requesting changes. The bridge HMAC fallback issue from the prior review is still present at head `b20d051e`.

Blocking finding:
- `internal/gateway/server.go` always passes `senderID` and `roleHdr` into `VerifyBridgeContext`, then injects them into context whenever `ok == true`.
- `internal/providers/claude_cli_mcp.go` accepts a pre-sender/role fallback by dropping the last two extras and still returns `(true, true)`.
- Result: a request signed with the older `localKey|sessionKey` format can add arbitrary `X-Sender-ID` / `X-Role` headers and the middleware will trust them, even though those fields were not covered by the matched HMAC.

Why this matters: sender/role are exactly the fields used by group-scoped cron/file-writer permission checks. Backward compatibility for old MCP configs is fine, but old signatures must not authorize newly added identity headers.

Required fix:
1. Change verification semantics so callers can distinguish `context signature valid` from `sender/role covered by signature` — e.g. return `ok, tenantVerified, senderRoleVerified`.
2. Only call `store.WithSenderID` / `store.WithRole` when `senderRoleVerified == true`.
3. Add regression tests for:
   - current signature with sender/role validates and injects them;
   - tampering sender/role fails full verification;
   - old signature remains accepted for backward compatibility but does **not** inject sender/role.

No merge attempt: this is security-sensitive permission propagation and remains blocked until that boundary is fixed.

PR #1108, opened 2026-05-06
feat(web_search): add `provider` arg to force a specific engine

mozaa-solana:
## Summary

Currently `web_search` resolves a tenant-configured provider chain (e.g. tavily → exa → brave) and stops at the first success. That's the right default for cost/latency, but it makes **cross-engine corroboration impossible at the agent level**: the caller has no way to ask \"search the same query on Exa specifically\" once Tavily already returned a hit.

This PR adds an optional `provider` argument. When set, the chain is narrowed to the named provider (case-insensitive); other providers are not queried. First-success-wins fallback is preserved when the arg is omitted.

## Concrete use case

A research agent verifying source freshness:

1. Call A: `web_search(query=\"Solana DEX volume\", freshness=\"pd\", provider=\"tavily\")`
2. Call B: `web_search(query=\"Solana DEX volume\", freshness=\"pd\", provider=\"exa\")`
3. URLs in **both** result sets → high-confidence original primary sources.
4. URLs in only one engine → republish suspect; verify with `web_fetch` page metadata.

Without `provider`, both calls return the same cached Tavily payload and the corroboration is meaningless.

## Changes

- **`internal/tools/web_search.go`**
  - New optional `provider` arg in `Parameters()`.
  - When passed, the resolved chain is narrowed to the matching provider (case-insensitive). Unknown name returns an error listing what IS configured for the tenant.
  - Cache key now includes `requestedProvider` so per-engine results don't collide.
- **`internal/tools/web_search_provider_param_test.go`** (new)
  - 5 cases: narrowing, case-insensitivity, unknown provider error, default first-success preservation, cache isolation.

## Backward compatibility

Fully backward-compatible — `provider` is optional. Existing callers (and existing cached entries during rollout) continue to work because the cache key change only affects new writes; old keys age out via TTL.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/tools/...` passes (new + existing)
- [x] Unit-tested all four code paths: narrowed-chain, case-insensitive match, unknown provider error, default fallback
- [x] Cache isolation verified (same query, two providers → both engines run)

## Related

Used by the c02-choros content-research agent (Max) to implement multi-layer source-freshness verification — agent calls each engine separately and cross-checks URLs to detect republished/syndicated content slipping past Tavily/Exa's date filters. Currently shipped in our fork (`mozaa-solana/goclaw`); upstreaming so the capability is broadly available.

### DIFF
--- a/internal/tools/web_search.go
+++ b/internal/tools/web_search.go
@@ -163,6 +163,10 @@ func (t *WebSearchTool) Parameters() map[string]any {
 				"type":        "string",
 				"description": "Filter results by discovery time. Supports 'pd' (past day), 'pw' (past week), 'pm' (past month), 'py' (past year), and date range 'YYYY-MM-DDtoYYYY-MM-DD'.",
 			},
+			"provider": map[string]any{
+				"type":        "string",
+				"description": "Optional: force a specific provider (e.g., 'tavily', 'exa', 'brave', 'duckduckgo'). When omitted, the tenant's configured provider chain is used (first-success-wins). Use this to force cross-engine corroboration — call once with each provider and compare results.",
+			},
 		},
 		"required": []string{"query"},
 	}
@@ -183,6 +187,7 @@ func (t *WebSearchTool) Execute(ctx context.Context, args map[string]any) *Resul
 	searchLang, _ := args["search_lang"].(string)
 	uiLang, _ := args["ui_lang"].(string)
 	freshness, _ := args["freshness"].(string)
+	requestedProvider, _ := args["provider"].(string)
 
 	params := searchParams{
 		Query:      query,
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Approved. This is a focused, backward-compatible tool-schema improvement with a concrete research-agent use case.

Review evidence:
- Diff is limited to `internal/tools/web_search.go` plus `internal/tools/web_search_provider_param_test.go`.
- Existing default behavior is preserved when `provider` is omitted: the configured chain remains first-success-wins.
- Explicit provider selection is case-insensitive and fails closed with a clear list of configured providers when the requested engine is unavailable.
- Cache isolation now includes the requested provider, so cross-engine corroboration will not replay one provider's cached result for another provider.
- Duplicate/prior-work check found only this open PR for the provider-param behavior; prior web_search work covers tenant provider-chain resolution, not per-call engine forcing.
- Local verification passed: `go test ./internal/tools`.

No Critical or Important findings. GitHub shows no checks for this branch, but the PR is small and covered by targeted local tests.

PR #1139, opened 2026-05-11
fix(cron): invert stateless gate to match UI label

codebit0:
## Summary

- The cron handler reset the session when `stateless=false` and skipped reset when `stateless=true` — the opposite of every UI locale's label (en/ko/zh/vi all say *"each run starts fresh without loading previous messages"*).
- One-character logic fix in `cmd/gateway_cron.go` (`if !job.Stateless` → `if job.Stateless`) so runtime matches the UI contract. Comment block reworded for clarity.
- No DB migration: existing values were set by users based on the UI label, so they already encode the intended behavior.

## Impact

A daily ETL cron with `stateless=true` was observed accumulating 38 messages over 18 days, causing the LLM to short-circuit tool calls and replay prior assistant turns (silent ETL failure — cron reported \`status=ok\` while DB rows were never written).

After the fix:
- `stateless=true` → session reset every run (matches UI label, intended for ETL/one-shot jobs)
- `stateless=false` → session preserved across runs (multi-run dialog crons)

## Test plan

- [x] \`go build ./...\` (PG build)
- [x] \`go build -tags sqliteonly ./...\` (Desktop build)
- [x] \`go vet ./cmd/...\`
- [ ] After deploy: confirm a \`stateless=true\` cron's session row has \`messages = '[]'\` after the first post-fix run
- [ ] Confirm a \`stateless=false\` cron retains prior turns (regression check)

### DIFF
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -85,11 +85,14 @@ func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg 
 		defer cancelCron()
 		cronCtx = store.WithTenantID(cronCtx, job.TenantID)
 
-		// Reset session before each cron run to prevent tool errors from previous
-		// runs from polluting the context and blocking future executions (#294).
-		// Save() persists the empty session to DB so stale data won't reload after restart.
-		// Stateless jobs skip this — they intentionally carry no session history.
-		if !job.Stateless {
+		// Stateless jobs explicitly reset their session before each run — they
+		// carry no history between executions, matching the "Stateless" UI label
+		// and statelessHelp ("each run starts fresh without loading previous
+		// messages"). Stateful (non-stateless) jobs keep prior turns to enable
+		// multi-run dialog.
+		// Save() persists the empty session to DB so stale data won't reload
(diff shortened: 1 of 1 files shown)


mrgoonie (approved):
Summary: This is the narrower version of the cron stateless-session fix and it matches the UI/help contract: stateless jobs now reset before each run, while stateful jobs keep prior turns for dialog-style schedules.nnRisk level: LownnMandatory gates:n- Duplicate/prior implementation: overlap found with #1032, but #1139 is the better-scoped variant. #1032 resets all cron jobs and would break legitimate stateful cron conversations; this PR only flips the inverted stateless condition.n- Project standards: passed; small targeted change in cmd/gateway_cron.go, no migration needed because existing values were set against the UI label semantics.n- Strategic necessity: clear value; prevents stateless ETL/report jobs from accumulating stale messages and replaying old context.n- CI/checks: green (go and web).nnFindings:n- Critical: nonen- Important: nonen- Suggestion: follow-up test coverage would be useful for stateless=true resets vs stateless=false preservation, but the one-line runtime fix is still safe to ship.nnVerdict: APPROVE

PR #1145, opened 2026-05-13
fix(skills): resolve {baseDir} placeholder in DB-backed skill content

codebit0:
## Summary

Both PG and SQLite skill stores returned SKILL.md content without the `{baseDir}` substitution that `skills.Loader.LoadSkill` performs for file-based access. Agents reading SKILL.md via this RPC saw the literal placeholder, breaking script-invocation lines that expect absolute paths.

Mirror the loader's substitution in both store implementations so behavior is consistent regardless of access path.

## Test plan

- [ ] Read a SKILL.md with `{baseDir}` placeholder via the DB store (skills.GetContent or RPC) and verify the placeholder is resolved to the absolute path
- [ ] Verify the same content read via file-based loader (existing path) is unchanged
- [ ] Sanity: agent uses a skill with script-invocation lines and the script runs from the correct directory

### DIFF
--- a/internal/store/pg/skills_content.go
+++ b/internal/store/pg/skills_content.go
@@ -40,6 +40,11 @@ func (s *PGSkillStore) LoadSkill(ctx context.Context, name string) (string, bool
 	if err != nil {
 		return "", false
 	}
+	// Resolve {baseDir} placeholder so SKILL.md script-invocation lines hold
+	// absolute paths regardless of agent CWD. Mirrors skills.Loader.LoadSkill
+	// substitution; without it, agents reading SKILL.md via this RPC see the
+	// literal placeholder.
+	content = strings.ReplaceAll(content, "{baseDir}", info.BaseDir)
 	return content, true
 }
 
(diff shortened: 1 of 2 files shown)


mrgoonie:
Backlog review: valuable small fix, but I would add one regression test before merge. The PG and SQLite stores now mirror file-loader {baseDir} substitution, which is the right behavior, and go/web checks pass. Missing piece: a store-level test proving DB-backed skill content returns an absolute substituted path for both PG/SQLite or at least the shared behavior.

mrgoonie (approved):
Summary: This is a small, targeted parity fix: DB-backed skill content now resolves the {baseDir} placeholder the same way the file-based loader already does. I checked both PG and SQLite store paths plus the existing loader behavior; no duplicate implementation or broader scope issue found.nnRisk level: LownnMandatory gates:n- Duplicate/prior implementation: clear; only this PR targets DB-backed LoadSkill placeholder substitution.n- Project standards: passed; mirrors internal/skills/loader.go behavior and uses the existing store.SkillInfo.BaseDir computed from file_path/version metadata.n- Strategic necessity: clear value; prevents DB/RPC skill reads from returning literal {baseDir} in script-invocation lines.n- CI/checks: green (go, web).nnFindings:n- Critical: nonen- Important: nonen- Suggestion: add a focused store-level regression test for PG/SQLite LoadSkill placeholder replacement when the test harness is convenient, but the production change itself is straightforward and matches existing loader semantics.nnVerdict: APPROVE

PR #1157, opened 2026-05-19
fix(channels): ignore zero failure timestamps

### LINKED ISSUE #1143: Channel diagnostics card shown incorrectly when channel is healthy due to Go zero-value time serialization
### Problem

The "What happened" diagnostics card on the channel detail page is displayed even when the channel is in a `healthy` state. This makes it look like something is wrong when the channel is actually operating normally.

### Root Cause

The `showDiagnosticsCard` condition in `channel-detail-page.tsx` includes:

```typescript
const showDiagnosticsCard =
    status?.state === "failed" ||
    status?.state === "degraded" ||
    !!status?.remediation ||
    !!status?.consecutive_failures ||
    !!status?.first_failed_at;    // ← bug here
```

The backend (`mergeChannelHealth` in `internal/channels/health.go`) correctly resets `FirstFailedAt` to `time.Time{}` when the channel transitions back to healthy. However, Go's zero-value `time.Time{}` serializes to `"0001-01-01T00:00:00Z"` in JSON — it does not become `null` or get omitted (the struct tag has no `omitempty`, and `omitempty` doesn't work for `time.Time` in Go's standard library anyway).

On the frontend, `!!status?.first_failed_at` evaluates `!!"0001-01-01T00:00:00Z"` → `true`, so the diagnostics card is always shown for any channel that has ever experienced a failure, even if it has fully recovered.

### Fix

Replace the raw truthy check with the existing `formatRelativeTime()` helper, which already filters out year≤ 1 timestamps:

```typescript
import { formatRelativeTime } from "../channels-status-view";

const showDiagnosticsCard =
    status?.state === "failed" ||
    status?.state === "degraded" ||
    !!status?.remediation ||
    !!status?.consecutive_failures ||
    !!formatRelativeTime(status?.first_failed_at);
```

`formatRelativeTime("0001-01-01T00:00:00Z")` returns `null` because of its existing guard:

```typescript
if (date.getUTCFullYear() <= 1) return null;
```

This ensures the diagnostics card only appears when there is a meaningful active or recent failure.

evgyur:
## Summary
- Moves the channel detail diagnostics visibility check into a shared helper.
- Uses `formatRelativeTime(status?.first_failed_at)` so Go zero-value timestamps (`0001-01-01T00:00:00Z`) are treated as absent.
- Adds regression coverage for healthy recovered channels and active degraded/failed states.

## Test plan
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm exec vitest run src/pages/channels/channels-status-utils.test.ts`
- `corepack pnpm exec tsc -b --pretty false`
- `git diff --check`

Closes #1143

### DIFF
--- a/ui/web/src/pages/channels/channel-detail/channel-detail-page.tsx
+++ b/ui/web/src/pages/channels/channel-detail/channel-detail-page.tsx
@@ -19,6 +19,7 @@ import {
   getChannelRemediationMeta,
   getRenderableChannelStatus,
   getChannelStatusMeta,
+  shouldShowChannelDiagnosticsCard,
 } from "../channels-status-view";
 import { useChannelTimeline } from "./channel-detail-timeline-hook";
 import { ChannelDetailDialogs } from "./channel-detail-dialogs";
@@ -130,12 +131,7 @@ export function ChannelDetailPage({
 
   const timelineItems = useChannelTimeline(status, t);
 
-  const showDiagnosticsCard =
-    status?.state === "failed" ||
-    status?.state === "degraded" ||
-    !!status?.remediation ||
-    !!status?.consecutive_failures ||
-    !!status?.first_failed_at;
+  const showDiagnosticsCard = shouldShowChannelDiagnosticsCard(status);
 
   const neutralHealthNote =
     !showDiagnosticsCard &&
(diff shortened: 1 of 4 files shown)


mrgoonie (approved):
Summary: This cleanly fixes the channel diagnostics false-positive by routing the detail-page visibility check through the existing zero-time-aware formatter and covering the recovered-channel regression.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; only matching issue/PR is #1143/#1157.
- Project standards: passed; small UI helper extraction plus colocated Vitest coverage matches existing channel UI patterns.
- Strategic necessity: clear value; fixes misleading healthy-channel diagnostics and closes #1143.
- CI/checks: missing/not reported on this branch; review relied on focused diff/code inspection.

Findings:
- Critical: none
- Important: none
- Suggestion: none

Verdict: APPROVE

PR #1172, opened 2026-05-25
Feat/provider kimi coding

raihan0824:
## Summary

Adds Moonshot's **Kimi Coding** endpoint as a first-class LLM provider.

The endpoint is OpenAI-compatible on the wire but rejects any request
that doesn't carry the fixed header `User-Agent: claude-code/0.1.0`.
Rather than special-case it, this PR generalises the need via a small
`WithExtraHeaders` option on `OpenAIProvider` — any future provider
that needs pinned identity headers can reuse the same mechanism without
touching the request path.

### Reusable: static request headers on `OpenAIProvider`

- `internal/providers/openai_config.go` — new `extraHeaders map[string]string`
  field, `WithExtraHeaders(map[string]string)` builder, `ExtraHeaders()`
  getter. Repeat calls merge; passing an empty map is a no-op.
- `internal/providers/openai_http.go` — `doRequest` now writes every
  `extraHeaders` entry to the outgoing request after the standard
  Content-Type / Authorization / OpenRouter site headers.
- `internal/providers/adapter_openai.go` — `ToRequest` mirrors the same
  headers into the returned `http.Header` so adapter callers see the
  same shape as direct callers.

### Provider wiring: `kimi_coding`

- `internal/store/provider_store.go` — `ProviderKimiCoding = "kimi_coding"`
  constant + entry in `ValidProviderTypes` + companion default constants
  (`KimiCodingDefaultAPIBase`, `KimiCodingDefaultModel`,
  `KimiCodingRequiredUserAgent`).
- `cmd/gateway_providers.go` and `internal/http/providers.go` — new
  `case store.ProviderKimiCoding:` in both provider-registration
  switches. Both construct an `OpenAIProvider` against the default base
  (`https://api.kimi.com/coding/v1` when none is supplied) and inject
  the required User-Agent via `WithExtraHeaders`.
- `ui/web/src/constants/providers.ts` — new `Kimi Coding (Moonshot)`
  dropdown entry with the API base pre-filled.

### Tests

`internal/providers/openai_extra_headers_test.go` — 3 new unit tests:
1. Header reaches the real outgoing HTTP request alongside Bearer auth.
2. Adapter `ToRequest` path mirrors the same headers.
3. Empty-map / nil `WithExtraHeaders` calls are a no-op (no accidental
   map allocation).

## Admin flow

1. **Providers → Add provider**
2. Pick **Kimi Coding (Moonshot)** — API base auto-fills.
3. Paste the API key, save.

Every outbound request from that provider now carries
`Authorization: Bearer <key>` and `User-Agent: claude-code/0.1.0`.

## Migration

None — pure additive change. Existing providers and grants are
unaffected; the new constant is opt-in via the admin UI.

## Verification

- [x] `go build ./...` (PG) — clean
- [x] `go build -tags sqliteonly ./...` (desktop) — clean
- [x] `go vet ./...` — clean
- [x] New tests + full `internal/providers/` and `internal/store/`
  packages — green
- [x] Web UI `pnpm tsc --noEmit` — no new errors

## Notes for reviewers

- `WithExtraHeaders` is intentionally narrow: a `map[string]string` set
  on the provider, applied unconditionally on every request. It does
  **not** support per-request override (callers can already pass headers
  through `ChatRequest.Options` for that). The split keeps the static
  "this provider always needs this header" case from leaking into the
  hot path.
- The same mechanism could replace the existing `siteURL` / `siteTitle`
  OpenRouter pair, but I left those alone to keep the diff focused.

### DIFF
--- a/cmd/gateway_providers.go
+++ b/cmd/gateway_providers.go
@@ -440,6 +440,19 @@ func registerProvidersFromDB(registry *providers.Registry, provStore store.Provi
 			prov := providers.NewOpenAIProvider(p.Name, p.APIKey, base, store.BytePlusDefaultModel)
 			prov.WithProviderType(p.ProviderType)
 			registry.RegisterForTenant(p.TenantID, prov)
+		case store.ProviderKimiCoding:
+			// Moonshot Kimi Coding requires a fixed User-Agent on every request.
+			// OpenAI-compatible wire shape otherwise.
+			base := p.APIBase
+			if base == "" {
+				base = store.KimiCodingDefaultAPIBase
+			}
+			prov := providers.NewOpenAIProvider(p.Name, p.APIKey, base, store.KimiCodingDefaultModel)
+			prov.WithProviderType(p.ProviderType)
+			prov.WithExtraHeaders(map[string]string{
+				"User-Agent": store.KimiCodingRequiredUserAgent,
+			})
+			registry.RegisterForTenant(p.TenantID, prov)
 		default:
 			prov := providers.NewOpenAIProvider(p.Name, p.APIKey, p.APIBase, "")
 			prov.WithProviderType(p.ProviderType)
--- a/internal/http/provider_models.go
+++ b/internal/http/provider_models.go
@@ -15,8 +15,8 @@ import (
 
 // ModelInfo is a normalized model entry returned by the list-models endpoint.
 type ModelInfo struct {
-	ID        string                        `json:"id"`
-	Name      string                        `json:"name,omitempty"`
+	ID        string                         `json:"id"`
+	Name      string                         `json:"name,omitempty"`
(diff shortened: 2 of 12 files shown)


mrgoonie:
Review/fix loop result: Approve.

Iterations: 1 review/fix/CI loop.

Fixes pushed:
- Rebased the PR onto current dev and removed stale unrelated fork history with a guarded force-with-lease update.
- Preserved the Kimi Coding provider implementation.
- Added a follow-up fix so Kimi model-list calls also use the required User-Agent and empty-api-base Kimi rows default to https://api.kimi.com/coding/v1 instead of OpenAI.

Current head: 540dc1130174fc3b8de8425599ae2efeba5559b7
Merge/conflict state: CLEAN.
CI state: green.
- go: pass, 6m30s
- web: pass, 48s

Remaining findings: none.
Unresolved blockers/questions: none.

PR #1155, opened 2026-05-19
fix(sandbox): avoid shell in FsBridge writes

evgyur:
## Summary
- Fixes `FsBridge.WriteFile` command injection by removing the `sh -c` write path.
- Passes the sandbox target filename as a discrete argv entry to `tee` instead of interpolating it into a shell command string.
- Adds regression tests for command-substitution-style filenames and append mode.

## Security impact
A path such as `notes/$(touch /tmp/pwned).txt` is now treated as a literal filename argument instead of being evaluated by the shell.

## Test plan
- `gofmt -w internal/sandbox/fsbridge.go internal/sandbox/fsbridge_test.go`
- `go test ./internal/sandbox -count=1`
- `git diff --cached --check`

Closes #1121

PR #974, opened 2026-04-20
fix(security): prevent SSRF via DNS bypass in provider URL validation

Vo-Linh:
## Summary

- **SSRF vulnerability confirmed via live exploit** on deployment at `10.10.27.30:18790`
- `validateProviderURL` only checked literal IPs (`net.ParseIP` returns nil for hostnames), allowing wildcard DNS services (`nip.io`, `sslip.io`) or attacker-controlled domains resolving to private IPs to bypass all SSRF protections
- Exploited to: exfiltrate API keys via `Authorization` header, port-scan Docker-internal network (`172.18.0.2:5432` postgres), probe cloud metadata endpoints

## Root Cause

```go
ip := net.ParseIP(host)  // returns nil for ANY DNS name
if ip != nil {            // never entered for hostnames
    if ip.IsPrivate() ... // dead code for DNS-based URLs
}
```

No IP-level check exists at HTTP dial time either — `NewDefaultTransport()` uses Go's default dialer.

## Three-Layer Fix

1. **DNS resolution at validation time** (`providers.go`): resolve hostnames via `net.LookupHost` and check every resolved IP against private/loopback/link-local ranges. Blocks `*.nip.io`, `*.sslip.io`, and attacker-controlled domains
2. **Scheme enforcement for all provider types** (`providers.go`): local types (Ollama/CLI/ACP) previously skipped scheme validation entirely — now `file://`, `gopher://` etc. are blocked for all types
3. **SSRF-safe dialer** (`defaults.go`): `NewDefaultHTTPClient()` uses a custom `DialContext` that re-checks resolved IPs at connection time — defense-in-depth against DNS rebinding after provider creation

## Files Changed

| File | Change |
|---|---|
| `internal/http/providers.go` | Refactored `validateProviderURL` with DNS resolution, extracted `checkHostPrivate`/`checkIPPrivate` helpers, added `allowPrivateProviderURLsFn` env-var gate |
| `internal/providers/defaults.go` | Added `ssrfSafeDialContext` with IP-pinning, wired into `NewDefaultHTTPClient()` |
| `internal/http/providers_url_validate_test.go` | 31 test cases covering DNS bypass, literal IPs, scheme enforcement, local types, allow-private flag, dual-stack, unresolvable hosts |

## Test plan

- [x] `go test -run TestValidateProviderURL ./internal/http/` — 31/31 pass
- [x] `go build ./...` — PG build clean
- [x] `go build -tags sqliteonly ./...` — Desktop/SQLite build clean
[...]

PR #973, opened 2026-04-20
fix(security): enforce RBAC tenant scope on ft= file token requests

Vo-Linh:
## Summary

- **Fix cross-tenant file access via ft= signed tokens.** The `handleServe` ft= code path only checked workspace/dataDir boundaries but skipped the RBAC tenant-scope restriction, allowing signed URLs to bypass tenant isolation and be used without authentication.
- **Unify path isolation check** for both ft= and bearer-authenticated requests into a single code path that always applies RBAC tenant scoping when tenant context is available.
- **Expand `deniedFilePrefixes`** with `/home/`, `/Users/`, `/var/lib/`, `/var/www/`, `/opt/`, `/srv/`, `/.ssh/`, `/.aws/`, `/.kube/`, `/.gnupg/`, `/.config/`.
- **Fail-closed** when neither workspace nor dataDir is configured — reject all file requests instead of defaulting open.

## Vulnerability details

**Severity:** Critical (confirmed via live testing)

The `ft=` HMAC file token was designed as a standalone auth mechanism for browser-embedded file URLs. However:

1. The `auth` middleware passed ft= requests to `handleServe` without tenant context
2. `handleServe` had separate code paths for ft= (lines 164-173, boundary-only) vs bearer (lines 176-205, boundary + RBAC tenant scope)
3. The ft= path never ran the RBAC tenant-scope check at lines 190-196
4. Signed URLs worked for **unauthenticated** clients — no bearer token required

**Attack chain:** Admin calls `POST /v1/files/sign` → gets ft= URL → URL readable by anyone (no auth, no tenant check, 5-min TTL). Confirmed: confidential workspace files (docx, json, txt) exfiltrated without authentication.

### `internal/http/files.go`
- `auth()`: When ft= is valid and RBAC is enabled, also resolve tenant context from bearer token if present
- `handleServe()`: Replace split ft=/bearer boundary checks with unified check that always applies RBAC tenant scope when tenant context exists
- `deniedFilePrefixes`: Add 9 new sensitive directory prefixes
- Fail-closed: early-return 404 when `workspace==""` and `dataDir==""`

### `internal/http/files_path_security_test.go`
- `TestFilesHandleServe_HomeDirBlocked`: Verify all new deny-list entries are blocked
- `TestFilesHandleServe_EmptyWorkspaceAndDataDir_Denies`: Verify fail-closed behavior

## Test plan

- [x] All 10 existing + new unit tests pass (`go test ./internal/http/ -run TestFiles`)
[...]

PR #1185, opened 2026-06-03
fix(security): consolidate & enhance batched security fixes (#1155, #967, #972, #974, #989, #973)

thieung:
Consolidates the batch of security PRs that were waiting on review into one branch, re-implemented cleanly on current `dev` with fixes for the blocking issues found during review. Each commit maps to an original PR and credits its author via `Co-authored-by`.

## What's included

| Commit | Source PR | Notes |
|--------|-----------|-------|
| `fix(sandbox): avoid shell in FsBridge writes` | #1155 | Shell-free `tee -- <path>` + stdin. Dropped CRLF/dead-code; removed stale `dd` helper + tests. |
| `fix(security): fail-closed on pairing DB errors across channels` | #967 | ~7 lines of real logic, LF-only (original PR was 99% CRLF noise). Shared DM/group helpers + 4 Telegram inline sites. |
| `fix(security): harden provider URL validation against SSRF` | **#972 + #974 merged** | The two PRs touched the same `validateProviderURL` with contradictory local-type policy. Merged into one coherent fix. |
| `feat(pipeline): add fail-closed tool call authorization gate` | #989 | Ported + fixed prefix-bug + added gate tests. |
| `fix(security): expand file-serve deny-list defense-in-depth` | #973 (salvage) | Deny-list + fail-closed log only. |

## Enhancements over the original PRs

- **SSRF (#972/#974):** reuse `internal/security.IsBlocked` (now exported) as the single CIDR source of truth instead of a hand-rolled `IsLoopback||IsPrivate||IsLinkLocal` check — this also closes the `0.0.0.0` / `::` (unspecified) bypass #974 missed. Local provider types are restricted to an explicit localhost allowlist (#972) so the local-type path can't reach `169.254.169.254`; remote hostnames are DNS-resolved and every resolved IP is checked (closes nip.io / sslip.io / config-time rebinding). Operator opt-in via `GOCLAW_ALLOW_PRIVATE_PROVIDER_URLS`.
- **Tool authz (#989):** fixed a correctness regression where agents configured with `toolCallPrefix` had every tool call wrongly blocked (name is now resolved to its canonical form before the allowlist lookup). Added the gate's first direct test coverage.
- **Pairing (#967):** ported logic only, in LF — avoids the `git blame`-destroying CRLF reformat in the original PR.
- **Sandbox (#1155):** removed leftover dead `dd` helper and stale tests.

## Decisions

- **#973 is superseded** by `532ff91d` on `dev` (`inferredScopedFileRoot` + symlink/TOCTOU hardening). Only the safe defense-in-depth deny-list is salvaged here; the original PR's `handleServe` change was fail-open and is intentionally **not** ported. Original #973 can be closed.
- **Dial-time DNS-rebinding hardening is deferred** (follow-up). The validation layer blocks at provider create/update time. Wiring a dial-time guard requires separating local/remote in the shared `NewOpenAIProvider` constructor + adjusting httptest-based provider tests — out of scope for this consolidation. No regression vs `dev` (which has no dial guard either).

## Verification

- `go build ./...` and `go build -tags sqliteonly ./...` — clean
- `go vet ./...` — clean
- `go test ./...` on all touched packages (http, providers, security, sandbox, channels, pipeline, agent, tools) — pass

## Follow-ups

1. Dial-time SSRF rebinding guard for remote providers (local/remote split in provider client construction).
2. Optional: migrate the 2 ex-`dd` sandbox tests' intent is already covered by the new `fsbridge_test.go`.

Closes #1155, #967, #972, #974, #989. Supersedes #973.

### DIFF
--- a/docs/02-providers.md
+++ b/docs/02-providers.md
@@ -554,7 +554,7 @@ ClaudeCLIProvider can be configured in `config.json`:
 }
 ```
 
-Or via database `llm_providers` table with `provider_type = "claude_cli"`.
+Or via database `llm_providers` table with `provider_type = "claude_cli"`. For database providers, `api_base` is the CLI executable selector (`"claude"` or an absolute binary path), not an HTTP base URL, so provider URL SSRF opt-ins do not apply to Claude CLI.
 
 ### Session Management
 
--- a/internal/agent/loop_lazy_mcp_test.go
+++ b/internal/agent/loop_lazy_mcp_test.go
@@ -5,12 +5,14 @@ import (
 	"testing"
 
 	"github.com/nextlevelbuilder/goclaw/internal/config"
+	"github.com/nextlevelbuilder/goclaw/internal/pipeline"
+	"github.com/nextlevelbuilder/goclaw/internal/providers"
 	"github.com/nextlevelbuilder/goclaw/internal/tools"
 )
 
 // mockExecTool is a simple tool that records whether it was executed.
 type mockExecTool struct {
-	name    string
+	name     string
 	executed bool
 }
 
@@ -163,6 +165,24 @@ func TestLoop_LazyMCP_NilAllowedTools_AllowsAll(t *testing.T) {
 	}
 	if !tool.executed {
 		t.Error("tool should execute when allowedTools is nil")
+	}
+}
+
+func TestLoopAuthorizeToolCall_PrefixedNameCanonicalLookup(t *testing.T) {
+	loop := NewLoop(LoopConfig{
+		AgentToolPolicy: &config.ToolPolicySpec{ToolCallPrefix: "proxy_"},
+	})
+	gate := loop.makeAuthorizeToolCall()
+	state := &pipeline.RunState{}
+	state.Tool.AllowedTools = map[string]bool{"exec": true}
+
+	ok, reason := gate(context.Background(), state, providers.ToolCall{
+		ID:   "tc-proxy-exec",
+		Name: "proxy_exec",
+	})
+
+	if !ok {
+		t.Fatalf("prefixed tool call should resolve to canonical allowlist entry; reason: %q", reason)
 	}
 }
 
--- a/internal/agent/loop_lazy_mcp_test.go
+++ b/internal/agent/loop_lazy_mcp_test.go
@@ -24,13 +24,16 @@ func (m *mockExecTool) Execute(_ context.Context, _ map[string]any) *tools.Resul
 	return tools.NewResult("ok from " + m.name)
 }
 
-// simulateLazyActivationCheck mimics the allowedTools check from loop.go for one tool call:
+// simulateLazyActivationCheck mimics the runtime authorize gate from
+// makeAuthorizeToolCall (loop_pipeline_callbacks.go) for one tool call.
+// The real gate is a PipelineDeps.AuthorizeToolCall callback invoked by ToolStage;
+// this helper exercises the same logic in isolation so tests stay unit-level.
 //
-//	if allowedTools != nil && !allowedTools[tc.Name] {
-//	    if l.tools.TryActivateDeferred(tc.Name) { allowedTools[tc.Name] = true }
+//	if allowed != nil && !allowed[name] {
+//	    if reg.TryActivateDeferred(name) { allowed[name] = true }
 //	    else { result = ErrorResult(...) }
 //	}
-//	if result == nil { result = l.tools.ExecuteWithContext(...) }
+//	if result == nil { result = reg.ExecuteWithContext(...) }
 //
 // Returns (result, blocked).
 func simulateLazyActivationCheck(reg *tools.Registry, allowedTools map[string]bool, toolName string) (*tools.Result, bool) {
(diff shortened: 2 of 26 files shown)


thieung:
@clark-cant Please review this PR

clark-cant:
## PR Review

Verdict: REQUEST CHANGES

### Summary
This is a valuable consolidation of several security fixes: sandbox write hardening, fail-closed pairing checks, provider URL SSRF validation, runtime tool-call authorization, and file-serving deny-list expansion. The scope is large but the commits are reasonably separated and CI is green.

Risk level: high, because this touches security boundaries and provider/tool runtime behavior across multiple subsystems.

### Findings

1. [HIGH] `internal/http/providers.go` — new URL validation breaks Claude CLI provider custom binary paths.

`ProviderClaudeCLI` is included in `localProviderTypes`, and `registerInMemory` still treats `p.APIBase` as a Claude CLI executable path:

```go
cliPath := p.APIBase // reuse APIBase field for CLI path
if cliPath ==  {
    cliPath = claude
}
if cliPath != claude && !filepath.IsAbs(cliPath) {
    cliPath = claude
}
```

But `handleCreateProvider` / `handleUpdateProvider` now always call `validateProviderURL(p.APIBase, p.ProviderType)`, and the new `validateProviderURL` enforces `http`/`https` scheme even for local provider types. That means a previously valid Claude CLI custom path like `/usr/local/bin/claude` is now rejected as a non-HTTP URL.

This is not just stricter SSRF validation; it changes the meaning of `api_base` for Claude CLI from “optional executable path” into “HTTP URL”. Please special-case Claude CLI path validation separately from provider base URL validation, or stop storing CLI path in `APIBase` before applying URL validation.

2. [MEDIUM] `internal/http/providers.go` — local provider allowlist likely rejects valid Docker/container local endpoints such as `http://host.docker.internal:11434/v1` only because it is explicitly listed, but sidecar/container hostnames are now impossible without the private-URL opt-in applying to local types.

For `ollama`, the validation returns before `GOCLAW_ALLOW_PRIVATE_PROVIDER_URLS` is considered:

```go
if localProviderTypes[providerType] {
    for _, a := range allowedLocalHosts { ... return nil }
    return fmt.Errorf(...)
}
```

That means even if an operator opts into private provider URLs, local provider types still cannot use a valid LAN/container-local hostname like `ollama:11434`, `host.lan`, or a Docker compose service name. Maybe this is intentional, but it should be explicit because it is a behavior change from the previous skip-validation path for local providers.

If the policy is “local types must only ever be localhost/host.docker.internal”, please document it and add a test that the opt-in flag intentionally does not apply to local providers. If not, let the opt-in gate cover local providers too.

3. [LOW] `internal/pipeline/tool_stage_authorize_test.go` — prefix regression test does not exercise the real prefix stripping path.

`TestAuthorizeGate_PrefixedName_CanonicalLookup` passes the already-resolved name `exec` into the fake gate. That does not actually prove `makeAuthorizeToolCall` handles `proxy_exec -> exec`; it only proves `exec` is allowed when `exec` is in the allowlist.

Please add a test around the real `Loop.makeAuthorizeToolCall()` path, or at least a unit test that calls `tools.StripToolPrefix` / `resolveToolCallName` with an actual prefixed tool name and verifies the allowlist lookup succeeds.

### Anti-AI-Slop Check
- Intent clear: yes
- Diff scope justified: mostly yes; large but maps to known security PRs
- Refactor justified: yes
- Author explanation needed: no

### Tests / CI
- CI: go ✅, web ✅
- Merge state: clean / mergeable
- Test coverage is good overall, but the Claude CLI `APIBase` path regression needs coverage before merge.

### Final Notes
The security direction is good, but the provider URL validator now conflates HTTP provider endpoints with Claude CLI executable paths. That regression is blocking.

clark-cant (approved):
## Re-review

Verdict: APPROVE

### Summary
Re-reviewed the updated PR after commit `bb47d85b` (`fix(providers): allow claude cli executable paths`). The previous blocking Claude CLI regression is fixed: `claude_cli` now bypasses URL validation and validates `api_base` as either `claude` or an absolute executable path, with create/update regression coverage.

### Previously Raised Findings
1. [HIGH] Claude CLI custom executable path rejected by URL validation — fixed.
   - `internal/http/providers.go` now special-cases `store.ProviderClaudeCLI` via `validateClaudeCLIExecutablePath`.
   - Added create/update handler regression tests for absolute Claude CLI paths.

2. [MEDIUM] Local URL providers ignore `GOCLAW_ALLOW_PRIVATE_PROVIDER_URLS` — clarified as intentional by test coverage.
   - `TestValidateProviderURL_LocalTypesIgnoreAllowPrivateFlag` documents the behavior: Ollama/ACP stay restricted to localhost/host.docker.internal even with the opt-in enabled.
   - This is conservative and acceptable for this security hardening PR.

3. [LOW] Prefix regression test did not hit the real loop gate — fixed.
   - Added `TestLoopAuthorizeToolCall_PrefixedNameCanonicalLookup`, which exercises `Loop.makeAuthorizeToolCall()` with `proxy_exec` resolving to canonical `exec`.

### Verification
- GitHub CI: go ✅, web ✅
- Local targeted tests: `go test ./internal/http ./internal/agent ./internal/pipeline ./internal/tools` ✅
- Merge state: clean

### Final Notes
Security posture looks materially better now, and the follow-up commit addresses the review blockers without broad extra churn. Safe to merge from my side.

PR #1189, opened 2026-06-04
fix(feishu): parse bot/v3/info top-level bot object for mention detection

### LINKED ISSUE #895: Bug: Lark Channel ignores `requireMention` intermittently in group chats
# Bug: Lark Channel ignores `requireMention` intermittently in group chats

## Problem Summary
The Lark channel agent sometimes auto-responds to messages in group chats and threads even when `requireMention: true` is configured and the bot is **not** @mentioned.

This is an **intermittent bug** — it does not happen on every message, making it harder to reproduce but very noticeable in active groups.

## Reproduction
1. Configure Lark channel with `requireMention: true`
2. Add bot to a group chat
3. Send messages in the group **without** @mentioning the bot
4. **Expected:** Bot does not respond
5. **Actual:** Bot occasionally responds anyway (intermittent)

## Observations
- Bot responds correctly to direct @mentions
- Bot sometimes responds to messages with no mention at all
- Happens in both group chats and threads
- More likely to trigger with:
  - Rich text messages
  - Replies/Forward messages
  - Messages with special formatting
  - `@All` mentions (may include bot in mentions list)

## Hypothesis
The Lark event payload structure varies by message type. For certain message types (rich text, cards, replies), the `mentions` field might be:
- Missing from payload
- Structured differently
- Parsed incorrectly by GoClaw

When GoClaw cannot find or parse `mentions`, it may default to "treat as mention" instead of "ignore".

## Environment
- GoClaw: v3.0.0
- Channel: Lark (WebSocket mode)
- Config: `requireMention: true`

## Suggested Fix
1. Add debug logging for Lark event payloads when `requireMention` is enabled
2. Log the raw `mentions` field for every incoming message
3. Explicitly handle cases where `mentions` is missing/empty → should NOT reply
4. Verify `@All` mention behavior (should bot respond to @All?)

## Impact
- Bot spams responses in group chats
- Degrades user experience in teams using Lark
- Makes `requireMention` config unreliable

jason1804:
## Problem

With `requireMention` enabled on a Lark group, the bot replied to messages that mentioned **other users**, not the bot (#895).

## Root cause

The legacy `/open-apis/bot/v3/info` endpoint returns the bot object at the **top level** of the response — `{"code":0,"msg":"ok","bot":{...}}` — not inside `data` like newer Lark APIs. `GetBotInfo()` read `resp.Data` (nil) and always failed:

```
level=WARN msg="feishu bot probe failed (will continue)" error="fetch bot info: unmarshal response: unexpected end of JSON input"
```

With `botOpenID` left empty, the mention parser falls back to treating **any** mention as a bot mention (`bot_parse.go`), so the bot answered messages mentioning other users.

Verified against the live Lark API — real response shape:

```json
{"bot":{"app_name":"...","open_id":"ou_..."},"code":0,"msg":"ok"}
```

## Changes

- `apiResponse`: add top-level `bot` field (legacy endpoint shape)
- `GetBotInfo()`: read top-level `bot` first, fall back to `data.bot` defensively, clear error when both missing
- Startup probe: retry 3× with linear backoff so transient failures don't degrade mention detection
- Tests: fixtures now match the real API shape + new fallback/missing-object cases

## Testing

- `go test ./internal/channels/feishu/` — all pass
- `go build ./...` + `go build -tags sqliteonly ./...` — both builds OK
- Live API verified: `bot` confirmed at top level with valid `open_id`
- Production log confirmed the exact unmarshal error before the fix

After deploy, startup log should show `feishu bot connected bot_open_id=ou_...` instead of the WARN, and mentions of other users no longer trigger replies.

Fixes #895


mrgoonie (approved):
Summary: This is a targeted Feishu/Lark mention-detection fix. It corrects the legacy /open-apis/bot/v3/info response parsing so botOpenID is populated from the top-level bot object, keeps a defensive data.bot fallback, and adds retry coverage for transient startup probe failures.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear — I found only this PR for the Feishu/Lark bot v3 info + mention-detection root cause.
- Project standards: passed — small channel-local Go change with tests in the existing package.
- Strategic necessity: clear value — fixes #895 where requireMention can degrade when bot identity probing fails.
- CI/checks: missing on GitHub for this branch; local targeted test passed: go test ./internal/channels/feishu

Findings:
- Critical: none
- Important: none
- Suggestion: none

Verdict: APPROVE

Issue #879, opened 2026-04-13
[Feature Request] MCP OAuth Integration — Support OAuth 2.1 Authentication via Existing ChatGPT/ OpenAI Subscriptions


tduong-p:
## Problem

Currently, GoClaw requires users to provide an **OpenAI API key** to use GPT models. This creates two significant friction points:

1. **Separate billing**: Users with an active ChatGPT Plus/Pro/Team subscription must also pay for a separate OpenAI API plan to use GoClaw — effectively paying twice for the same capability.
2. **Key management overhead**: API keys must be generated, stored, rotated, and kept secure. There is no per-user, session-scoped authentication.

Meanwhile, **Claude Desktop's MCP connectors** already support OAuth 2.1 authentication flows, allowing users to connect third-party MCP servers using their existing subscriptions. OpenAI's Apps SDK also implements the [MCP authorization spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization) with OAuth 2.1, enabling ChatGPT to act as an OAuth client.

## Proposed Solution

Implement an **MCP OAuth Provider** in GoClaw that:

- Exposes GoClaw as an OAuth 2.1 protected resource server
- Enables users to authenticate via their **existing ChatGPT/OpenAI account** (Plus, Pro, Team, Enterprise)
- Issues access tokens scoped to the authenticated user's subscription
- Allows GoClaw to route LLM requests through the user's own subscription billing

This mirrors the OAuth flow that Claude Desktop MCP connectors already support — but for the OpenAI side.

---

## Background: How Claude Desktop MCP OAuth Works (Reference)

To understand the target behavior, here is how Claude Desktop currently handles OAuth with MCP servers:

### Components

| Component | Role |
|---|---|
| **MCP Server** | Protected resource server; exposes tools, verifies tokens on each request |
| **Authorization Server** | Identity provider (Auth0, Okta, etc.) that issues tokens and publishes discovery metadata |
| **Client** (ChatGPT/Claude Desktop) | Acts on behalf of the authenticated user; supports dynamic client registration + PKCE |

### OAuth Metadata Endpoints Required

1. **Protected resource metadata** at `GET /.well-known/oauth-protected-resource`:
```json
{
  "resource": "https://goclaw.example.com",
  "authorization_servers": ["https://auth.openai.com"],
  "scopes_supported": ["model:gpt-4o", "model:gpt-4o-mini", "files:read"],
[...]

PR #1196, opened 2026-06-09
feat(mcp): MCP OAuth 2.1 client for tool servers

thotam:
## Summary
Adds a complete MCP OAuth 2.1 authorization flow so agents can call MCP tool servers that require user-delegated access (RFC 9728 discovery → 8414/OIDC metadata → 7591 DCR → PKCE S256 / refresh / client_credentials), with encrypted token storage, per-user vs global isolation, SSRF-safe outbound calls, and automatic cleanup of stale tokens when a server's URL or OAuth config changes.

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] Tests pass: `go test -race ./...` — `-race` unavailable locally (no cgo/gcc on Windows dev box). Ran without `-race`: `internal/mcp/oauth`, `internal/http`, `internal/gateway`, `internal/agent` all pass. (Two pre-existing `store/pg` `TestBuildSkillInfo*` failures are Windows path-separator only, unrelated to this PR; CI runs `-race` on Linux.)
- [x] No hardcoded secrets or credentials (tokens AES-256-GCM encrypted at rest)
- [x] SQL queries use parameterized `$1, $2` (PG) / `?` (SQLite) — no string concat
- [x] New user-facing strings added to all 3 locales (en/vi/zh) — `internal/i18n` catalogs + `ui/web/src/i18n/locales/{en,vi,zh}`
- [x] Migration version bumped in `internal/upgrade/version.go` (RequiredSchemaVersion → 74; new migration `000074_mcp_oauth_tokens`)

## Test Plan

**Automated**
- `internal/mcp/oauth/*_test.go` — discovery fallback + 5-min cache, PKCE S256 math, RFC 7591 DCR, refresher (cache/expiry/refresh, per-user vs global, `InvalidateServer`).
- `internal/http/mcp_oauth_test.go` — all 5 endpoints, admin gating, WS `mcp.oauth_complete` event, callback HTML.
- `internal/http/mcp_update_oauth_purge_test.go` — tokens purged on URL change & OAuth-config change; NOT purged on unrelated update.
- `tests/integration/v3_mcp_oauth_*` — store CRUD + AES-256-GCM round-trip + tenant isolation, `DeleteServerOAuthTokens`, E2E `start → callback → DB` via httptest.
- `internal/gateway/event_filter_test.go` — `mcp.oauth_complete` routed only to initiating user / in-tenant admins, fail-closed across tenants.
- `internal/agent/loop_mcp_user_test.go` — Bearer token injection + graceful expiry + 401 cache purge.

**Manual**
- Configure an OAuth-protected MCP server (DCR and manual client_id), authorize via popup → token stored encrypted, agent calls succeed.
- `client_credentials` grant completes server-side (no popup) and reports authorized.
- Change server URL / client_id → status flips to "not authorized", old token removed from `mcp_oauth_tokens`, pool reconnects.
- `goclaw migrate up` applies `000074` cleanly on PostgreSQL; SQLite desktop build starts.

## Notes for reviewers
- Single squashed commit (`605d1e69`); 64 files, +6712/−126.
- Security surface: all outbound HTTP (discovery/DCR/token/refresh) goes through the SSRF-safe client with pinned IPs + response size limits; OAuth status/revoke are admin-gated; callback page builds its postMessage payload via `json.Marshal` (no reflected XSS from `error_description`).


thotam:
@mrgoonie This PR is ready for review. The branch has been rebased onto the latest `dev` and the previous merge conflict is resolved — `mergeStateStatus` is now `MERGEABLE`.

Key conflict resolutions during the rebase:
- Renumbered the migration to avoid the `000074` collision with upstream's `run_timeline_items`: PG migration is now `000084_mcp_oauth_tokens` (`RequiredSchemaVersion = 84`); SQLite migration is key `50` (`SchemaVersion = 51`).
- `config_channels.go`: kept upstream's `MCPAllowedHosts` + `ChatBehavior` and added the OAuth `PublicURL` field.
- `loop_tool_filter.go`: dropped this branch's older per-user MCP tool plumbing in favor of dev's `NewUserToolOverlay`/`userTools` mechanism (the OAuth token storage is independent of tool filtering).
- `schema.sql` / sqlite `factory.go`: unioned the `mcp_oauth_tokens` table and store wiring with upstream's new tables/stores.

Verified locally: `go build ./...`, `go build -tags sqliteonly ./...`, and `go test ./internal/config ./internal/agent ./internal/mcp/...` all pass. Could you take a look when you get a chance? Thanks!

mrgoonie:
Summary: I did the cron-safe merge-readiness pass after your rebase. This is not ready for full review/merge yet because the current GitHub `go` check is red.

Decision: blocked / deferred

Evidence:
- PR is now very large for cron review scope: 65 files, +7141/-132, so I did not deep-review the full diff in this pass.
- CI run `27900891131` has `release-versioning` and `web` green, but `go` failed.
- The failing integration test is `TestOAuthCallbackMissingState` in `v3_mcp_oauth_flow_test.go:290`: missing OAuth `state` should return 400, but currently returns 200.

Next step: fix the missing-state validation regression and get the `go` check green. After that, this needs a full MCP/OAuth security review because it touches token storage, callback validation, migrations, config, and tool-server auth flow.

thotam:
@mrgoonie Fixed in `af40ef2d`.

Root cause: `handleCallback` rendered a 200 HTML page for the whole missing-`code`/`state` branch. The fix distinguishes the two cases:
- auth **code present but `state` missing** → **400** (state is the CSRF token, so this is a malformed / CSRF-risk callback) — this is the integration regression you flagged.
- neither `code` nor `state` (user opening the URL directly), provider errors, and exchange failures → keep the **200** HTML popup page.

(My first attempt returned 400 for the whole branch, which broke the unit test `TestHandleCallbackMissingCodeAndState` that expects 200 for a bare hit — that's now corrected.)

Verified locally:
- `go test -tags integration -run 'TestOAuthCallbackMissingState|TestOAuthCallbackProviderError' ./tests/integration/` → pass
- `go test ./internal/http -run Callback` → all callback tests pass
- `go build ./...` green

Note: `TestOAuthHandlerManualCallbackRequiresSameRequester` (the ChatGPT/Codex `/v1/auth/...` handler, unrelated to this PR) fails on my local Windows box because it binds port 1455; it passes on the Linux CI runners and is not touched by this change.

Ready for the MCP/OAuth security review when you have time.

mrgoonie:
Summary: Thanks for the quick fix. The previous CI blocker is resolved now, but this PR is still too large/high-risk for a cron-safe full approval pass.

Decision: unblocked from CI / deferred for full MCP-OAuth security review

Evidence:
- Current head is `af40ef2d9b46fe5cbe7c84f0ecc75a049b14ae74`.
- GitHub checks are now green: `go`, `web`, and `release-versioning` all passed on run `27901882637`.
- `mergeStateStatus` is `CLEAN`.
- The PR remains a large MCP/OAuth change: 65 files, about +7.1k/-132 lines, touching OAuth callback validation, token storage, migrations, config, and tool-server auth flow.

Next step: this needs a full focused MCP/OAuth security review before approval/merge. The next review pass should cover callback state/CSRF guarantees, token encryption/storage and tenant/user scoping, migration/versioning, allowed-host/public-url boundaries, and tool-server credential isolation. I’m leaving it open and not approving/merging from the cron pass.

thotam:
## 🔒 MCP OAuth self-audit (to speed up the security review)

Pre-emptive self-audit against the five areas you flagged, with file:line citations on the current head. Not a substitute for your review — just a map.

### 1. Callback state / CSRF
- CSRF state: 16 bytes `crypto/rand` → base64url (128-bit), minted in `StartFlow` (`internal/mcp/oauth/flow.go`, `generateState`).
- One-time + expiry: `ExchangeCode` looks up `pending[state]` and **deletes it immediately** (replay protection), rejects if `time.Since(CreatedAt) > flowTTL` (10 min); a cleanup loop evicts stale flows every minute (`flow.go:154-163, 211-219`).
- PKCE S256 mandatory for every authorization-code flow (OAuth 2.1), verifier 32 bytes `crypto/rand` (`flow.go:96-108`).
- Callback validation: missing `state` with a `code` present → **400** (the regression fix); provider `error=` → 200 HTML popup (`internal/http/mcp_oauth.go:337-344`).
- The callback does **not** read tenant/user from the request: the token is persisted under `flow.TenantID/UserID/ServerID` captured at `StartFlow` time (by the authenticated admin), so a forged callback cannot inject scope.

### 2. Token encryption / storage
- AES-256-GCM: `access_token`, `refresh_token`, `dcr_client_secret` are `crypto.Encrypt`-ed before insert and decrypted on read (`internal/store/pg/mcp_oauth_tokens.go:114-130, 87-103`; `internal/crypto/aes.go`).
- All queries are parameterized (`$1,$2,$3`) — no string concatenation.

### 3. Tenant / user scoping & isolation
- Every get/upsert/delete carries `WHERE server_id=$1 AND tenant_id=$2`, with `user_id IS NULL` (global) or `=$3` (per-user) (`pg/mcp_oauth_tokens.go:43-56, 206-222`).
- `tenantID` always comes from `store.TenantIDFromContext(ctx)`, never from caller input.
- Endpoint auth: start/status/revoke/discover are all `requireAuth(RoleAdmin)`; only the callback is public (per OAuth, protected by the state token) (`mcp_oauth.go` `RegisterRoutes`).

### 4. Migration / versioning
- PG `000084_mcp_oauth_tokens` (after upstream's 83), `RequiredSchemaVersion=84`; SQLite migration key 50, `SchemaVersion=51`.
- Partial unique indexes `(server_id,tenant_id) WHERE user_id IS NULL` and `(server_id,tenant_id,user_id) WHERE user_id IS NOT NULL` avoid duplicate tokens from NULL-distinct UNIQUE semantics.
- FK `ON DELETE CASCADE` to `mcp_servers` and `tenants` cleans up tokens when a server/tenant is deleted.

### 5. Allowed-host / public-url boundaries
- Discovery, DCR, and token exchange all go through `security.Validate(url)` (pins the resolved IP, blocks private/metadata) + `security.NewSafeClient(15s)` (`internal/mcp/oauth/discovery.go:200-202`, `dcr.go:46`, `mcp_oauth.go:184,626`).
- `public_url` config is an explicit admin override for the callback URL.

### Two points worth your explicit confirmation (not claiming they're correct)
- `requireAuth(RoleAdmin)` + tenant-from-context on a tenant-scoped table — please confirm RoleAdmin here is tenant-aware as intended vs the `requireTenantAdmin` used on other tenant-scoped writes.
- Per-user token cache/pool eviction (`refresher.InvalidateCache(serverID, userID)`) — please confirm the cache key is keyed by (server, user) so a per-user token can't leak across users.

mrgoonie (changes requested):
Summary: I did the focused MCP/OAuth security pass on the current head (`af40ef2d9b46fe5cbe7c84f0ecc75a049b14ae74`). CI is green and the previous missing-state callback regression is fixed, but I do **not** consider this safe to merge yet.

Risk level: High

Mandatory gates:
- Duplicate/prior implementation: clear — related feature request #879 exists, but I did not find another merged/open implementation that already satisfies this OAuth client flow.
- Project standards: issue found — the PR violates the repo’s tenant-scope guard rule for tenant-scoped writes.
- Strategic necessity: clear value — MCP OAuth is valuable and fits the MCP/tool-server roadmap.
- CI/checks: green — GitHub `go`, `web`, `release-versioning` are green; local `go test ./internal/http ./internal/mcp/oauth ./internal/agent`, `go build ./...`, and `go build -tags sqliteonly ./...` passed.

Findings:
- Important: Tenant-scoped OAuth token writes are only gated by `requireAuth(RoleAdmin)`, not the repo’s tenant-admin guard. `RegisterRoutes` wires `POST /v1/mcp/oauth/start`, status, revoke, and discover through `requireAuth(permissions.RoleAdmin)` only (`internal/http/mcp_oauth.go:70-81`). But this feature writes/deletes rows in the tenant-scoped `mcp_oauth_tokens` table (`handleStart` persists tokens via `activateToken`; `handleRevoke` deletes by tenant/server/user at `internal/http/mcp_oauth.go:557-570`). The project standard in `CLAUDE.md` is explicit: `RoleAdmin` is not a tenant check; tenant-scoped writes must use `requireTenantAdmin` plus tenant-scoped SQL. The SQL scoping is present, but the tenant-admin membership gate is missing. Please add a tenant store dependency/middleware and enforce `requireTenantAdmin` for token mint/revoke/admin status operations, with tests covering a RoleAdmin caller that is not a tenant admin for the target tenant.
- Important: The web form exposes manual OAuth/DCR controls that the backend does not actually honor. The UI serializes `oauth.use_dcr`, `oauth.auth_endpoint`, and `oauth.token_endpoint` (`ui/web/src/pages/mcp/mcp-form-dialog.tsx:179-188`), and the update path fingerprints those fields for purge decisions (`internal/http/mcp.go:219-250`), but `handleStart` always runs discovery and only reads `client_id`, `client_secret`, `scope`, and `grant_type` from settings (`internal/http/mcp_oauth.go:190-230`). That means `use_dcr=false` and manually supplied endpoints are silently ignored. This is dangerous for operators because the UI implies manual endpoint configuration works, while the runtime may still discover/register against a different AS. Either wire the manual endpoint/use_dcr behavior end-to-end or remove/disable those UI fields until supported.

Confirmed during review:
- Callback state/CSRF mechanics look sound after `af40ef2d`: random state, one-time delete before exchange, TTL, PKCE S256, and callback persists using the captured flow tenant/user/server rather than request input.
- Token storage is parameterized and encrypted for access token, refresh token, and DCR client secret in both PG and SQLite.
- The refresher cache is keyed by `serverID:userID`; because `mcp_servers.id` is a primary key, this does not appear to leak per-user tokens across tenants. Pool keys include tenant/server/user.
- The previous missing-state bug is covered and passing.

Verdict: REQUEST_CHANGES — not safe to merge yet. Please fix the tenant-admin guard and the UI/backend OAuth config mismatch, then I can re-review/merge if the updated head stays green.

thotam:
Both findings addressed in `6f669c80`.

**Finding 1 — authorization.** You're right that `requireAuth(RoleAdmin)` is not a tenant check. I started with a blanket `requireTenantAdmin` on start/status/revoke, but that breaks per-user self-service that the UI already exposes: `MCPUserCredentialsDialog` shows an "Authorize" button to regular (non-admin) users for their *own* per-user credential, calling `/oauth/start` with their own `user_id`. A blanket tenant-admin gate would 403 that.

So I mirrored the existing per-user MCP credentials model (`resolveTargetUserID` in `mcp_user_credentials.go`):
- start/status/revoke now accept any authenticated user; each handler calls a new `authorizeOAuthScope`.
- a caller may manage their **own** per-user token (self-service); the **global/server token** (`user_id=""`) and **another user's** token require **tenant-admin** (owner bypass) — so a RoleAdmin who is not a tenant admin is rejected, which is your finding.
- `discover` stays admin-only (it only previews AS metadata).
- Added a `TenantStore` dep. Tests: self-service (200), on-behalf-of-another (403), global-by-non-tenant-admin (403).

If you'd rather make per-user OAuth admin-managed only (and hide the self-service button in the UI), I can switch to blanket tenant-admin — but the scope-based model matches how env/header per-user credentials already work in this repo.

**Finding 2 — manual OAuth config, end-to-end.**
- `use_dcr=false` (a `*bool`, so legacy/absent stays discover+DCR) now skips discovery/DCR and uses the operator's `auth_endpoint`/`token_endpoint`, SSRF-validated via `security.Validate` before use.
- `token_endpoint` is always required; `auth_endpoint` only for auth-code grants — `client_credentials` needs no authorization URL, matching the UI which hides that field for that grant.
- the refresher already refreshes against the stored `token_endpoint` and the callback persists it, so manual-mode tokens refresh correctly.
- `oauthFingerprint` now includes `use_dcr` (nil normalized to true) so toggling DCR mode purges stale tokens.
- the web form only serializes the manual endpoints when `use_dcr` is off, so stale form-state values never reach the backend/fingerprint.

Audited all MCP dialogs (form, global OAuth, per-user credentials, grants, tools) for consistency. Local `go build ./...`, `go build -tags sqliteonly ./...`, the web build, and the OAuth handler tests pass.

mrgoonie (approved):
## Summary

I re-reviewed the updated head `6f669c80e4de01408d9c2f38073dd9b77352fbcf` after your reply/fix. The two previous Important blockers are addressed, and I consider this merge-ready now.

## Risk level

High feature surface, but no unresolved Critical/Important findings after this pass.

## Mandatory gates

- **Duplicate/prior implementation:** clear — #879 is the related feature request, not a duplicate implementation.
- **Project standards:** passed — the tenant-scoped OAuth token operations now have scope-aware authorization and tenant-admin enforcement where required.
- **Strategic necessity:** clear value — MCP OAuth unlocks OAuth-protected tool servers and fits the MCP roadmap.
- **CI/checks:** green — GitHub `go`, `web`, and `release-versioning` are green on the current head.

## Verified fixes

### Authorization blocker

`internal/http/mcp_oauth.go` now wires a `TenantStore` and gates OAuth scope through `authorizeOAuthScope`:

- Own per-user token is self-service.
- Global/server token and another user's token require `requireTenantAdmin`.
- `discover` also requires tenant admin.

The new tests cover:

- Non-tenant-admin rejection.
- Self-service own-user authorization.
- On-behalf-of-another rejection.

This resolves my prior `RoleAdmin is not a tenant check` blocker while preserving the existing per-user credential UX.

### Manual OAuth / DCR blocker

`handleStart` now:

- Honors `use_dcr=false`.
- Uses operator-supplied `auth_endpoint` / `token_endpoint`.
- SSRF-validates the manual endpoints.
- Supports `client_credentials` with token endpoint only.
- Avoids discovery/DCR in manual mode.

Also:

- `extractOAuthFingerprint` now includes normalized `use_dcr`.
- The web form stops persisting stale manual endpoint values when DCR is enabled.

## Local verification

I ran fresh on this head:

```bash
go test ./internal/http ./internal/mcp/oauth ./internal/agent
go build ./...
go build -tags sqliteonly ./...
```

## Verdict

**APPROVE / MERGE_READY**

I am going to attempt squash merge per maintain rules.

PR #1209, opened 2026-06-11
fix: incorrect Bitrix portal domain check

thanh-nguyen-95:
## Summary

Some valid cloud and self-hosted (custom) Bitrix portal domains are not supported.

For example, when using `egany.bitrix24.vn` - a valid Bitrix portal domain, the UI shows an error:

image

**The solution:**

Add more support for Bitrix domains, which includes:

- Cloud domains: bitrix24.vn/.tr/.es/.com.br/.com.ar
- Self-hosted domains

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist

- [x] `go build ./...` passes

## Test Plan

- [x] Update cloud domain test
- [x] Add self-hosted domain test

### DIFF
--- a/internal/gateway/methods/bitrix_portals.go
+++ b/internal/gateway/methods/bitrix_portals.go
@@ -69,9 +69,9 @@ type bitrixPortalView struct {
 // Validation regexes. Kept package-level so they compile once and tests can
 // reference them directly.
 var (
-	// Bitrix24 cloud portal hosts. Matches *.bitrix24.{com,eu,ru,de,fr,jp,in,kz,ua,by}
+	// Bitrix24 cloud portal hosts. Matches *.bitrix24.{com,eu,ru,de,fr,jp,in,kz,ua,by,vn,tr,es,com.br,com.ar}
 	// plus self-hosted *.bitrix.info. Subdomain regex matches DNS label rules.
-	bitrixCloudDomainRegex = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\.(bitrix24\.(com|eu|ru|de|fr|jp|in|kz|ua|by)|bitrix\.info)$`)
+	bitrixCloudDomainRegex = regexp.MustCompile(`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?\.(bitrix24\.(com|eu|ru|de|fr|jp|in|kz|ua|by|vn|tr|es|com\.br|com\.ar)|bitrix\.info)$`)
 
 	// Valid hostname regex for self-hosted Bitrix24 instances (custom domains).
 	// Accepts any valid FQDN or hostname with optional port (e.g. bx.example.com, portal.internal:8443).
--- a/internal/gateway/methods/bitrix_portals.go
+++ b/internal/gateway/methods/bitrix_portals.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"log/slog"
+	"net"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/google/uuid"
@@ -13,6 +16,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/gateway"
 	"github.com/nextlevelbuilder/goclaw/internal/i18n"
 	"github.com/nextlevelbuilder/goclaw/internal/permissions"
+	"github.com/nextlevelbuilder/goclaw/internal/security"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
 	"github.com/nextlevelbuilder/goclaw/pkg/protocol"
 )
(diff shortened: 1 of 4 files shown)


clark-cant:
## PR Review

Verdict: REQUEST CHANGES

### Summary
This PR fixes a real UX/backend validation mismatch for Bitrix portal domains by adding missing Bitrix24 regional cloud domains and allowing self-hosted/custom domains.

The cloud-domain expansion looks good, but the new self-hosted regex is currently too broad for a value that is later used to make backend outbound HTTPS requests. It allows localhost, private/internal IPs, metadata IPs, and arbitrary internal hostnames without any explicit SSRF/private-network policy.

Risk level: medium-high, because this changes an admin-provided string into an outbound server-side request target.

### Findings

1. [HIGH] internal/gateway/methods/bitrix_portals.go:78 — self-hosted domain validation permits SSRF-style internal targets.

The new selfHostedDomainRegex accepts any DNS-looking hostname plus optional port:

- localhost
- 127.0.0.1
- 169.254.169.254
- 10.0.0.1
- internal service names like postgres, redis, metadata.google.internal if lowercased and syntactically valid
- arbitrary ports like portal.example.com:8443

This domain is later used by the Bitrix client to build outbound URLs like:

https://<domain>/rest/profile.json
https://<domain>/rest/<method>.json

So this is no longer just form validation; it controls a server-side network destination. Admin-only access reduces exposure, but it is still a tenant-scoped SSRF footgun and a behavior change from the previous strict Bitrix host allowlist.

Suggested fix: decide and encode the intended self-hosted policy explicitly:

- If private/internal self-hosted Bitrix is allowed: gate it behind an explicit admin/system opt-in, document it, and add tests showing private/internal hostnames are intentionally accepted only when enabled.
- If not allowed by default: run the constructed URL through the existing SSRF validator / blocked-CIDR logic, reject loopback/link-local/private/metadata targets, and add negative tests for 127.0.0.1, 169.254.169.254, 10.0.0.1, localhost, and .internal/.local as appropriate.

At minimum, add tests so this policy is not accidental.

2. [MEDIUM] internal/gateway/methods/bitrix_portals.go:78 — optional port is syntactically accepted without range validation.

The regex accepts any digit sequence after colon, including :0 or :99999. That will fail later at request construction or connect time, but the create RPC would persist the portal first. Please validate port range 1-65535 if ports are supported, and add tests for invalid ports.

3. [LOW] Tests no longer prove non-Bitrix arbitrary public domains are rejected.

TestBitrixPortals_Create_InvalidDomain changed from not-a-bitrix-domain.com to -invalid.com. Since arbitrary self-hosted public domains are now allowed, this is understandable, but please replace it with a set of policy-specific invalid cases: spaces, scheme/path, leading hyphen, empty label, invalid port, and any internal/private hosts per the decision above.

### Anti-AI-Slop Check
- Intent clear: yes — valid regional Bitrix cloud domain egany.bitrix24.vn was incorrectly rejected.
- Diff scope justified: mostly yes, but the self-hosted expansion needs explicit security policy.
- Refactor justified: not applicable.

### Tests / CI
- CI: go ✅, web ✅
- Local targeted check: go test ./internal/gateway/methods ✅
- Test gap: missing SSRF/private-network/port boundary tests for self-hosted domains.

### Final Notes
The regional cloud additions should merge. The self-hosted wildcard needs a clear outbound-network safety boundary before this lands.

clark-cant:
## Re-review

Verdict: REQUEST CHANGES

### Summary
Thanks for the follow-up. The new commits address most of the previous review: backend SSRF and port validation were added, frontend validation was mirrored, and tests now cover blocked literal IPs, invalid ports, and cloud domains.

However, one security gap remains in the hostname-resolution path.

### Findings

1. HIGH internal/gateway/methods/bitrix_portals.go:424-437 — hostname SSRF check only validates the first resolved IP.

validateSelfHostedDomain resolves hostnames with net.LookupHost, but only checks the first returned address. That leaves a bypass when a hostname resolves to multiple addresses and a blocked/private address is not first. DNS result ordering is not a security boundary and can vary by resolver, response, and client.

For SSRF validation, the safe rule is: reject if any resolved IP is blocked, not just the first.

Suggested fix: iterate over every resolved address, parse each IP, and return an error if any address is blocked by security.IsBlocked.

Please add a regression test with a stubbed resolver returning a public IP followed by a private IP, for example 8.8.8.8 and 10.0.0.1, and assert validation rejects it.

2. MEDIUM internal/gateway/methods/bitrix_portals_test.go — tests depend on real public DNS.

TestValidateSelfHostedDomain_ValidPublicDomain uses real domains like google.com and example.com:443. This makes the unit test dependent on CI network and DNS availability. The comment says failure without internet is acceptable, but for CI tests it usually is not.

Suggested fix: make the resolver injectable/package-level for tests, or split syntax validation from DNS validation and stub the DNS part. Then test public-domain success with deterministic fake results.

### Previously Raised Items
- Cloud regional domains: looks good.
- Literal private, loopback, and metadata IP blocking: added.
- .local and .localhost blocking: added.
- Port range validation: added.
- Frontend validation: added as UX mirror; backend remains authoritative.

### Tests / CI
- CI: go ✅, web ✅
- Local targeted check: go test ./internal/gateway/methods ✅

### Final Notes
This is close. Change DNS validation from checking only the first IP to checking every resolved IP, and make the DNS tests deterministic. Then I am good to approve.

clark-cant:
## Final re-review

Verdict: APPROVE

### Summary
Re-reviewed after commit 4c500055. The remaining SSRF review items are addressed.

### What changed
- validateSelfHostedDomain now checks every IP returned by DNS, not only the first one.
- DNS resolution is injectable via package-level lookupHost, so tests no longer depend on real public DNS.
- Added deterministic tests for:
  - all-public resolution success
  - public-then-private and private-then-public multi-IP rejection
  - metadata IP rejection
  - IPv6 loopback rejection
  - DNS failure, empty results, invalid IP strings

### Verification
- Local targeted check: go test ./internal/gateway/methods ✅
- CI: go ✅, web ✅
- Merge state: clean

### Final Notes
The original cloud domain fix is good, and the self-hosted domain expansion now has an explicit outbound-network safety boundary. Safe to merge from my side.

mrgoonie:
Review/fix loop complete.

- Reviewed #1209 and found one actionable follow-up: the UI validator accepted regional/self-hosted Bitrix domains, but existing i18n strings still described only `bitrix24.com`.
- Fix applied: pushed follow-up commit `00073113` on `dev` updating the en/vi/zh validation copy. Direct push to the fork PR head was blocked because maintainer edits were disabled / no permission on `EGANY-Team/goclaw`.
- Merge: PR merged into `dev` via merge commit `05f0357e`.
- Local verification: `go test ./internal/gateway/methods`, `go build ./...`, `pnpm -C ui/web install --frozen-lockfile`, `pnpm -C ui/web build`, plus later `go test -race ./internal/cron` and `go test -race ./...` while watching the dev CI recovery.
- CI: the original post-merge run for `00073113` was cancelled by later `dev` pushes after its core jobs had passed. The latest `dev` run `27524242260` on `faf3e282` is green: web, go, beta_version, linux binaries, Docker latest/full, beta alias promotion, and release publish all succeeded.
- Conflicts: none.
- Remaining blockers/questions: none.

PR #1234, opened 2026-06-17
feat(cron): per-job LLM provider/model override

zezaeoh:
## Problem

`makeCronJobHandler` builds the agent `RunRequest` with no provider/model, so **cron jobs always run on the agent default provider**. High-frequency scheduled jobs (e.g. periodic triage/digest) therefore can't be routed to a cheaper model — even though **heartbeats already can**, via `agent_heartbeats.provider_id`/`model` → `RunRequest.ProviderOverride`/`ModelOverride`.

## Change

Give cron jobs the same per-run override that heartbeats have, end-to-end (schema → store → handler → tool):

- **Schema**: add `provider_id` (FK → `llm_providers`, `ON DELETE SET NULL`) and `model` to `cron_jobs`, mirroring the existing `agent_heartbeats` columns.
  - Postgres: migration `000081`
  - SQLite: `schema.sql` + schema migration
  - **Bump `RequiredSchemaVersion` 80 → 81** — without this, `upgrade`/auto-upgrade stops at v80 and the new migration never applies (the column is added but unreachable at runtime).
- **Store**: read/scan the new columns (pg + sqlite); set them via `UpdateJob` (`CronJobPatch.ProviderID/Model`).
- **Handler**: resolve `job.ProviderID`/`job.Model` into `RunRequest.ProviderOverride`/`ModelOverride`, reusing the exact resolution pattern from `heartbeat/ticker.go`.
- **Tool**: the `cron` MCP tool's `add`/`update` now accept `provider` (provider **name**, resolved via `ProviderStore.GetProviderByName`) or `providerId` (UUID), plus `model`. `ProviderStore` is wired into the cron tool. This makes the feature usable by agents/automation without hand-editing the DB.

`NULL`/unset → agent default, so **existing jobs are unaffected**.

## Not included (follow-up)

Dashboard / web UI exposure. Today the override is settable via the `cron` tool (`provider`/`model`), the `UpdateJob` patch, or directly in the DB.

## Verification

- `go build ./...` and `go build -tags sqlite ./...` — exit 0; `go vet`, `gofmt` — clean
- cron unit tests pass on both pg and sqlite stores
- **Deployed on-box (v3.14.0-beta.3), DB migrated v73 → v81, exercised end-to-end through the `cron` tool:**
  - `add` with `model` → persisted on the job
  - `update` with `model` → updated in place
  - `add` with `provider: "groq"` → resolved to the provider's UUID (`providerId` set)
  - unknown provider name → clean error, **no job created** (resolution happens before insert)

### DIFF
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -12,6 +12,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/bus"
 	"github.com/nextlevelbuilder/goclaw/internal/channels"
 	"github.com/nextlevelbuilder/goclaw/internal/config"
+	"github.com/nextlevelbuilder/goclaw/internal/providers"
 	"github.com/nextlevelbuilder/goclaw/internal/scheduler"
 	"github.com/nextlevelbuilder/goclaw/internal/sessions"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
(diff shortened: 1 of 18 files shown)


mrgoonie (approved):
Summary: This cleanly extends the existing heartbeat provider/model override pattern to cron jobs end-to-end for storage, scheduling, and the agent-facing cron tool. The migration/version bumps and green CI cover the main compatibility risks.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: clear; no equivalent merged cron override implementation found, and this reuses the heartbeat override model rather than adding a parallel provider path.
- Project standards: passed; changes stay in existing cron store/handler/tool wiring and preserve nil/unset fallback behavior.
- Strategic necessity: clear value; lets high-frequency scheduled jobs run on cheaper or more appropriate models without changing agent defaults.
- CI/checks: green (release-versioning, go, web).

Findings:
- Critical: none.
- Important: none.
- Suggestion: follow up later with RPC/dashboard exposure if non-agent clients need to configure these fields too.

Verdict: APPROVE / MERGE_READY

Issue #993, opened 2026-04-22
MCP tool routing divergence: deferred MCP tools fail "unknown tool" 1ms on direct invoke (search-mode bug)


haidang1810:
## Symptom

Same MCP server (gdrive), same agent (Leader), 3 tools in 3 different states:

| Tool | mcp_tool_search visibility | Direct invoke result |
|------|---------------------------|----------------------|
| `updateGoogleSheet` | VISIBLE | WORKS (~983ms full MCP round-trip) |
| `getGoogleSheetContent` | VISIBLE | FAILS `unknown tool: mcp_gdrive__getGoogleSheetContent` (1-5ms fail-fast) |
| `uploadFile` | NOT VISIBLE | N/A |

A second agent (Director) on the same MCP server can invoke all 3 tools directly.

The 3-orders-of-magnitude timing gap (1-5ms vs 583-1002ms) proves the failure happens at the **tenant/registry router level**, before any request reaches the gdrive MCP server.

## Root cause investigation — TWO hypotheses

The actual root cause depends on whether gdrive has `settings.require_user_credentials = true`. The two paths through the code behave differently.

### Hypothesis A — gdrive is NOT user-cred (standard path)

If gdrive goes through `LoadForAgent` → `connectViaPool` → `maybeEnterSearchMode`, the bug is in two architectural defects:

**A1. Non-deterministic inline/deferred split** in `internal/mcp/manager.go:359-417`:

```go
func (m *Manager) maybeEnterSearchMode() {
    allNames := m.ToolNames()  // iterates m.servers (Go map = random order)
    if len(allNames) <= mcpToolInlineMaxCount { return }
    deferSet := ...
    for _, name := range allNames[mcpToolInlineMaxCount:] { // arbitrary slice
        deferSet[name] = struct{}{}
    }
}
```

Each Manager construction produces a different inline-vs-deferred split because Go map iteration is undefined-order. Two agents with identical grants disagree on which tools are directly callable. Threshold = 40 tools (`mcpToolInlineMaxCount` constant in same file, line 31). Director likely has < 40 tools total → no search mode → all-inline → all 3 tools work direct.

**A2. Lazy activator wired but never called from execution path.** `internal/agent/resolver.go:329-330`:

```go
// Search mode: too many tools — register mcp_tool_search meta-tool.
// Also wire lazy activator so deferred tools can be called by name directly.
toolsReg.SetDeferredActivator(mcpMgr.ActivateToolIfDeferred)
```

[...]

PR #1235, opened 2026-06-18
fix(agent): surface per-user MCP tools (require_user_credentials) to the LLM tool list

tech-synity:
## Problem

Per-user MCP servers (`require_user_credentials`) load and connect on every run
(`mcp.user_tools_loaded user=… tools=N`), but their tool definitions are **never
sent to the model**. The agent only ever sees the shared MCP server's tools and
reports the per-user `mcp_<prefix>__*` tools as unavailable — it can never call a
tool it doesn't see.

## Root cause

The per-user `BridgeTool` objects are intentionally **not** registered in the
shared tool registry, to avoid leaking one user's credentialed tool to other
users (see `getUserMCPTools`). But `PolicyEngine.FilterTools` only emits a
definition for a tool it can resolve from the registry, so per-user tools are
silently dropped:

```go
for _, name := range allowed {
    if tool, ok := registry.Get(canonical); ok { // per-user tools: ok=false → skipped
        defs = append(defs, ToProviderDef(tool))
    }
}
```

The call site captures the per-user tools (`userTools`) but only logs them — they
were never merged into the request's tool list. (Execution already works via
`executeToolForActor`, which resolves per-actor from `mcpUserTools`; the model
simply never received the definitions.)

## Fix

Surface per-user tools through a **request-scoped overlay registry** passed to the
policy engine:

- `tools.NewUserToolOverlay(base, userTools)` wraps the shared registry with the
  calling actor's per-user tools (overrides `Get`/`List`). It is built per request
  and discarded after filtering — the shared registry, and other users, are never
  mutated, so there is no cross-user credential leak. Returns `base` unchanged when
  there are no per-user tools.
- `buildFilteredTools` wraps the registry in the overlay when per-user tools are
  present, so `FilterTools` evaluates **and** emits them under the same
  profile/allow/deny policy as registry tools. The no-policy path appends the defs
  directly (all tools allowed).
- Fixes a stale `getUserMCPTools` docstring that claimed shared-registry
  registration.

A side benefit: because per-user tools now flow through `FilterTools`, an explicit
`deny` on a per-user tool name is honored (previously impossible) — covered by a
test.

## Tests

- `internal/tools/user_tool_overlay_test.go` — overlay List/Get, empty→base
  passthrough, policy emits by default, and **deny-by-name suppresses** (policy now
  applies to per-user tools).
- `internal/agent/loop_user_mcp_tools_surface_test.go` — per-user tools surface to
  the LLM list, final-iteration strip, dedupe.

`go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./internal/agent/... ./internal/tools/...`, and the new tests pass.

### DIFF
--- a/internal/agent/image_gen_gate_test.go
+++ b/internal/agent/image_gen_gate_test.go
@@ -55,7 +55,7 @@ func TestImageGenGate_AllTrue_ToolPresent(t *testing.T) {
 	prov := &imageCapableProvider{imageGen: true}
 	l := buildImageGenLoop(true, prov)
 
-	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil)
+	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil, nil)
 
 	if !hasImageGenTool(defs) {
 		t.Error("expected image_generation tool when all gate conditions are true")
@@ -68,7 +68,7 @@ func TestImageGenGate_ProviderNoCapability_ToolAbsent(t *testing.T) {
 	prov := &imageCapableProvider{imageGen: false}
 	l := buildImageGenLoop(true, prov)
 
-	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil)
+	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil, nil)
 
 	if hasImageGenTool(defs) {
 		t.Error("image_generation must NOT be in tools when provider does not advertise ImageGeneration")
@@ -82,7 +82,7 @@ func TestImageGenGate_ProviderNotCapabilitiesAware_ToolAbsent(t *testing.T) {
 	prov := &stubProvider{}
 	l := buildImageGenLoop(true, prov)
 
-	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil)
+	defs, _, _ := l.buildFilteredTools(&RunRequest{}, false, 1, 10, nil, nil)
 
 	if hasImageGenTool(defs) {
 		t.Error("image_generation must NOT be in tools when provider is not CapabilitiesAware")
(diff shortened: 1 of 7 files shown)


clark-cant:
## PR Review

Verdict: APPROVE

### Summary
- Fixes per-user MCP tools (`require_user_credentials`) not being surfaced to the LLM by adding a request-scoped overlay registry.
- Keeps credentialed tool objects out of the shared registry while allowing policy filtering to emit their provider definitions.
- Risk: low/medium because this touches tool exposure, but the implementation is scoped and covered by focused regression tests.

### Findings
- No blocking findings.

### Anti-AI-Slop Check
- Intent clear: yes — bug/root cause is specific and tied to per-user MCP tool visibility.
- Diff scope justified: yes — small, targeted changes around tool filtering + overlay.
- Refactor justified: yes — request-scoped overlay is the right boundary; no shared registry mutation.

### Tests / CI
- Local: `go test ./internal/tools ./internal/agent` ✅
- Added coverage for overlay List/Get, default policy emission, deny-by-name suppression, per-user surfacing, final-iteration stripping, and dedupe.
- PR status currently shows no check rollup in `gh pr view`; merge state is `UNSTABLE`, so still wait for GitHub CI/branch state before merge if required by repo policy.

### Final Notes
Looks good. The important bit is that deny policy now applies before per-user MCP defs reach the provider, so this does not reopen the cross-user leak this code was trying to prevent.

mrgoonie (approved):
Summary: This fixes the per-user MCP tool visibility gap by passing request-scoped user MCP tools into tool filtering through an overlay registry, without registering credentialed BridgeTools globally.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: clear; I found only this PR for the exact require_user_credentials tool-definition surfacing gap. Related #993 is adjacent but not the same failure mode.
- Project standards: passed; the change keeps per-user BridgeTools out of the shared registry and preserves per-actor execution through executeToolForActor.
- Strategic necessity: clear value; per-user MCP tools were loaded/executable but invisible to the model, making require_user_credentials servers unusable.
- CI/checks: local targeted tests passed; GitHub checks are missing/not reported on this PR.

Findings:
- Critical: none
- Important: none
- Suggestion: there is adjacent pre-existing fragility around native tool definitions with nil Function in some tool-definition loops; this PR does not introduce that pattern and keeps the native image_generation append after the filters it changes.

Verdict: APPROVE

PR #1243, opened 2026-06-18
fix(discord): enable binding-based agent routing

msaidf:
## Changes

Fixes a limitation where Discord channel instances always route to the instance's `agent_id`, ignoring `bindings` in `config.json`.

**Root cause:** The Discord handler sets `targetAgentID = c.AgentID()` unconditionally, which becomes `msg.AgentID` in the inbound message. The gateway consumer only checks `resolveAgentRoute(cfg, ...)` when `msg.AgentID == ""`. Since it's never empty, bindings are silently ignored for all DB-backed Discord channels.

**Fix:** Clear `targetAgentID` to `""` in `handleMessage` so the consumer's existing `resolveAgentRoute` logic matches via `cfg.Bindings`. If no binding matches, it falls back to `cfg.ResolveDefaultAgentID()`.

**File:** `internal/channels/discord/handler.go` (+12, -1)

## Testing

Verified on a live deployment with 9 agents and 10 Discord channel bindings — all channels route to their designated agents correctly.


mrgoonie (approved):
Summary: This is a small, targeted fix for Discord DB-backed channel routing. Clearing the default `AgentID` lets the existing gateway `resolveAgentRoute` binding logic run for normal Discord messages, while voice messages still override to `VoiceAgentID` when configured.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear — I found prior Discord fixes, but no existing merged/open fix for this binding-routing path.
- Project standards: passed — the change reuses the existing channel/bus routing contract where empty `InboundMessage.AgentID` means resolve from bindings/default agent.
- Strategic necessity: clear value — fixes multi-agent Discord deployments where `config.json` bindings were silently ignored.
- CI/checks: missing on GitHub; local targeted tests passed: `go test ./internal/channels/discord ./cmd`.

Findings:
- Critical: none
- Important: none
- Suggestion: consider adding a regression test for Discord normal-message routing with DB instance bindings when the channel test harness covers this path.

Verdict: APPROVE

Issue #1244, opened 2026-06-18
[DB] Missing FK constraints across 16+ tables — critical data integrity risk


bclermont:
## Summary

A full audit of all 80 migration files reveals **16+ tables with missing foreign key constraints** across `agent_id`, `tenant_id`, `trace_id`, `channel_id`, `team_id` and other relationship columns. Combined with missing transactions on multi-step writes, the database has zero referential integrity guarantees. Data silently corrupts on deletes, migrations, or any write bug.

This is not a single oversight — it is a systemic pattern across the entire schema.

---

## Confirmed Missing FK Constraints

| Table | Column | Should Reference | ON DELETE |
|---|---|---|---|
| `traces` | `agent_id` | `agents(id)` | SET NULL |
| `traces` | `parent_trace_id` | `traces(id)` | SET NULL |
| `spans` | `trace_id` | `traces(id)` | CASCADE |
| `spans` | `parent_span_id` | `spans(id)` | SET NULL |
| `spans` | `agent_id` | `agents(id)` | SET NULL |
| `agent_context_files` | `agent_id` | `agents(id)` | CASCADE |
| `agent_context_files` | `tenant_id` | `tenants(id)` | CASCADE |
| `hooks` | `tenant_id` | `tenants(id)` | CASCADE |
| `tenant_hook_budget` | `tenant_id` | `tenants(id)` | CASCADE |
| `webhooks` | `tenant_id` | `tenants(id)` | CASCADE |
| `webhooks` | `channel_id` | `channel_instances(id)` | SET NULL |
| `webhook_calls` | `tenant_id` | `tenants(id)` | CASCADE |
| `webhook_calls` | `agent_id` | `agents(id)` | SET NULL |
| `usage_snapshots` | `agent_id` | `agents(id)` | SET NULL |
| `usage_events` | `tenant_id` | `tenants(id)` | CASCADE |
| `usage_events` | `team_id` | `agent_teams(id)` | SET NULL |
| `usage_event_rollups` | `tenant_id` | `tenants(id)` | CASCADE |
| `subagent_tasks` | `spawned_by` | `subagent_tasks(id)` | SET NULL |
| `channel_instances` | `agent_id` | `agents(id)` | CASCADE |

[...]

PR #1246, opened 2026-06-20
fix(db): add FK constraints, missing indexes, and transaction fixes (migrations 081-082)

bclermont:
## Summary

Addresses https://github.com/nextlevelbuilder/goclaw/issues/1244

Staged implementation following the maintainer's guidance: audit queries first, then FK constraints in two targeted batches, with orphan verification before each batch.

---

## Step 1 — Audit SQL (`migrations/audit_fk_integrity.sql`)

Standalone runnable queries to verify orphans and tenant_id drift **before** applying migrations. All queries should return 0 rows on a clean database. Covers all tables targeted by migrations 081 and 082.

---

## Step 2 — Migration 081: Core FK constraints + transaction fixes

**Go transaction fixes (no migration required):**
- `HookStore.Create` (`internal/store/pg/hooks.go`): wrapped INSERT hooks + INSERT hook_agents loop in a single `BeginTx` transaction. Previously a partial failure left an orphaned hook row with incomplete agent links.
- `AgentStore.Create` (`internal/store/pg/agents.go`): replaced compensating `DELETE FROM agents` anti-pattern with a real transaction. Previously the compensating delete could itself fail, leaving a permanent ghost agent.

**9 FK constraints added (all verified zero-orphan on live DB before adding):**

| Table | Column | References | ON DELETE | Rationale |
|---|---|---|---|---|
| `hooks` | `tenant_id` | `tenants(id)` | CASCADE | Created in migration 052 after migration 027 added tenant FKs everywhere else — omission |
| `tenant_hook_budget` | `tenant_id` | `tenants(id)` | CASCADE | PK with no parent reference |
| `webhooks` | `tenant_id` | `tenants(id)` | CASCADE | Created in migration 059 without FK |
| `webhook_calls` | `tenant_id` | `tenants(id)` | CASCADE | Same |
| `webhook_calls` | `agent_id` | `agents(id)` | SET NULL | Log record, survives agent delete |
| `spans` | `trace_id` | `traces(id)` | CASCADE | NOT NULL column, dangling spans accumulate silently |
| `spans` | `agent_id` | `agents(id)` | SET NULL | Audit column |
| `traces` | `agent_id` | `agents(id)` | SET NULL | Audit column |
| `usage_events` | `team_id` | `agent_teams(id)` | SET NULL | Only remaining gap after migration 080 constrained the other columns |

---

## Step 3 — Migration 082: Remaining FK constraints + indexes + tenant isolation

**FK constraints added (all verified zero-orphan on live DB):**

| Table | Column | References | ON DELETE | Rationale |
|---|---|---|---|---|
| `webhooks` | `channel_id` | `channel_instances(id)` | SET NULL | FK and index both missing |
| `channel_contacts` | `merged_id` | `channel_contacts(id)` | SET NULL | Self-ref, no FK |
| `traces` | `parent_trace_id` | `traces(id)` | SET NULL | Self-ref, has indexes but no FK |
| `spans` | `parent_span_id` | `spans(id)` | SET NULL | Self-ref |
| `subagent_tasks` | `spawned_by` | `subagent_tasks(id)` | SET NULL | Self-ref, no FK and no index |

**ON DELETE behavior fix:**
- `team_task_attachments.created_by_agent_id`: changed from RESTRICT → SET NULL, consistent with every other `created_by_agent_id` column in the schema

**Tenant isolation fix:**
- `hook_executions`: added nullable `tenant_id` (backfilled from `hooks.tenant_id` via `hook_id`). This was the only audit table with no tenant isolation at the DB level.

**Indexes added:**
`idx_webhooks_channel_id`, `idx_subagent_tasks_spawned_by`, `idx_hooks_created_by`, `idx_hook_executions_hook_id`, `idx_hook_executions_tenant_id`, `idx_usage_events_team_id`

---

## Not addressed in this PR

- `hooks.created_by` FK: stores an OAuth-derived UUID with no corresponding DB table — no FK possible by design. Index added instead.
- Schema smells (hardcoded sentinel UUID in column default, `bitrix_portals.agent_id` type inconsistency, `kg_dedup_candidates.tenant_id` nullability, `vault_documents_scope_consistency` NOT VALID) — require larger refactor
- SQLite store: separate migration system, not updated here

See `docs/db-normalization-status.md` for full status.

### DIFF
--- a/internal/hooks/dispatcher.go
+++ b/internal/hooks/dispatcher.go
@@ -438,6 +438,7 @@ func (d *stdDispatcher) writeExec(ctx context.Context, cfg HookConfig, ev Event,
 	exec := HookExecution{
 		ID:         uuid.New(),
 		HookID:     &hookID,
+		TenantID:   &ev.TenantID,
 		SessionID:  ev.SessionID,
 		Event:      ev.HookEvent,
 		InputHash:  inputHash,
--- a/internal/hooks/types.go
+++ b/internal/hooks/types.go
@@ -186,7 +186,8 @@ type HookConfig struct {
 // error_detail (BYTEA) is AES-256-GCM encrypted before storage.
 type HookExecution struct {
 	ID          uuid.UUID  `json:"id"`
-	HookID      *uuid.UUID `json:"hook_id,omitempty"` // NULL when hook deleted (ON DELETE SET NULL)
+	HookID      *uuid.UUID `json:"hook_id,omitempty"`  // NULL when hook deleted (ON DELETE SET NULL)
+	TenantID    *uuid.UUID `json:"tenant_id,omitempty"`
 	SessionID   string     `json:"session_id"`
 	Event       HookEvent  `json:"event"`
 	InputHash   string     `json:"input_hash"`  // canonical-JSON sha256, 64 hex chars
--- a/internal/store/pg/agents.go
+++ b/internal/store/pg/agents.go
@@ -97,6 +97,12 @@ const agentSelectCols = `id, agent_key, display_name, frontmatter, owner_id, pro
 		 model_fallback, shell_deny_groups, kg_dedup_config,
 		 agent_type, is_default, status, budget_monthly_cents, created_at, updated_at, tenant_id`
 
+// sqlExecer is satisfied by both *sql.DB and *sql.Tx, allowing helper
+// functions to be called within or outside a transaction.
+type sqlExecer interface {
+	ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error)
+}
+
 func (s *PGAgentStore) Create(ctx context.Context, agent *store.AgentData) error {
 	if agent.ID == uuid.Nil {
 		agent.ID = store.GenNewID()
@@ -108,7 +114,14 @@ func (s *PGAgentStore) Create(ctx context.Context, agent *store.AgentData) error
 	if tenantID == uuid.Nil {
 		tenantID = store.MasterTenantID
 	}
-	_, err := s.db.ExecContext(ctx,
+
+	tx, err := s.db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("begin transaction: %w", err)
+	}
+	defer tx.Rollback() //nolint:errcheck
+
+	_, err = tx.ExecContext(ctx,
 		`INSERT INTO agents (id, agent_key, display_name, frontmatter, owner_id, provider, model,
 		 context_window, max_tool_iterations, workspace, restrict_to_workspace,
 		 tools_config, sandbox_config, subagents_config, memory_config,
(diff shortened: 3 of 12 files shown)


mrgoonie (changes requested):
Summary: This PR is the right direction for #1244, but it cannot merge in its current form because the migration numbering conflicts with an existing migration.

Risk level: High

Mandatory gates:
- Duplicate/prior implementation: clear intent, but migration version collision found
- Project standards: failed — migration filenames must be unique and monotonic
- Strategic necessity: clear value; this is important DB integrity hardening
- CI/checks: failing

Findings:
- Critical: CI fails in `internal/store/pg` before the migration suite can run: `migrate.New: failed to open source, file://../../../migrations: duplicate migration file: 000081_cron_provider_override.down.sql`. This PR adds `migrations/000081_add_fk_constraints.*`, but migration 081 is already used by the cron provider override work. Renumber these migrations to the next unused versions and bump `internal/upgrade/version.go` accordingly, then rerun the migration/store tests.

Verdict: REQUEST_CHANGES

mrgoonie (changes requested):
Summary: This is important DB-integrity work and the previous migration-number blocker is fixed. One SQLite-facing regression remains before this is safe to merge.

Risk level: High

Mandatory gates:
- Duplicate/prior implementation: clear; this continues #1244 and uses next migration numbers after `000081_cron_provider_override`
- Project standards: issue found — shared tracing behavior must not assume Postgres-only FK semantics
- Strategic necessity: clear value; staged FK hardening and transaction fixes are worthwhile
- CI/checks: green

Findings:
- Important: `internal/store/sqlitestore/tracing.go` now relies on `ON DELETE CASCADE` from `spans.trace_id` to `traces(id)`, but SQLite's schema defines `spans.trace_id TEXT REFERENCES traces(id) ON DELETE SET NULL` (`internal/store/sqlitestore/schema.go` / `schema.sql`). After this change, `DeleteTracesOlderThan` deletes old traces only, leaving old spans behind with `trace_id` nulled instead of deleting them. The Postgres migration adds cascade semantics, but SQLite was explicitly left outside this PR's migration work, so the manual span delete still needs to stay in the SQLite store or SQLite schema/migration must be updated with matching cascade behavior and tests.

Verdict: REQUEST_CHANGES

bclermont:
@mrgoonie I pushed after processing your latest comment

mrgoonie (approved):
Summary: Re-reviewed after the author pushed in response to the last review. The previous SQLite trace-cleanup blocker is fixed; Postgres/upgrade tests pass, GitHub CI is green, and the remaining notes are non-blocking cleanup.

Risk level: High

Mandatory gates:
- Duplicate/prior implementation: clear; this directly addresses #1244 and I did not find an overlapping PR covering the same FK/transaction hardening.
- Project standards: passed; migration numbers now follow the existing sequence and RequiredSchemaVersion is bumped to 83.
- Strategic necessity: clear value; this is useful DB integrity hardening with staged audit queries, FK batches, and transaction fixes.
- CI/checks: green on GitHub. Local `go test ./internal/store/pg ./internal/upgrade` passes. Local `go test -tags sqlite ./internal/store/sqlitestore` still fails in migration-v49 upgrade tests with `duplicate column name: provider_id`, but that is outside this PR's touched SQLite trace-cleanup change and not evidence against the FK PR itself.

Findings:
- Critical: none.
- Important: none blocking. The prior Important finding about SQLite `DeleteTracesOlderThan` is fixed by keeping the explicit span delete in `internal/store/sqlitestore/tracing.go`.
- Suggestion: update stale migration comments/PR text that still say migration 081/082 while the files are now `000082_*` and `000083_*`. It is cosmetic, but it will reduce future audit confusion.

Verdict: APPROVE

PR #1247, opened 2026-06-21
fix(mcp): expose delegate tool via MCP bridge

zezaeoh:
## Problem

`BridgeToolNames` (the allowlist of GoClaw tools exposed over the MCP bridge) omitted `delegate`. As a result, claude-cli / Claude-Code-backed agents never received the `delegate` tool — even when their orchestration mode correctly resolved to `delegate` with active `agent_links`.

The native agent loop injects `delegate` per orchestration mode (`orchModeDenyTools`), but bridge-backed sessions only ever see the curated `BridgeToolNames` set. So two linked agents would show `delegate` mode + delegate targets in the dashboard, yet inter-agent delegation was impossible from any bridge-backed (claude-cli) session.

## Fix

Add `delegate` to `BridgeToolNames`.

It is safe to expose unconditionally, mirroring `team_tasks` (already bridged):
- It **self-gates** via `CanDelegate` / `agent_links` — an agent with no delegate targets simply can't delegate.
- It resolves its **source agent from the `X-Agent-ID` header context** (`store.AgentIDFromContext`), the same mechanism `team_tasks` uses, which the bridge handler already populates.

## Test

- `go build ./...` ✅
- `go test ./internal/mcp/` ✅
- gofmt clean

No migration / schema change.

### DIFF
--- a/internal/mcp/bridge_server.go
+++ b/internal/mcp/bridge_server.go
@@ -17,6 +17,8 @@ import (
 
 // BridgeToolNames is the subset of GoClaw tools exposed via the MCP bridge.
 // Excluded: spawn (agent loop), create_forum_topic (channels).
+// delegate is included: it self-gates via CanDelegate/agent_links and resolves
+// its source agent from the X-Agent-ID header context, same as team_tasks.
 var BridgeToolNames = map[string]bool{
 	// Filesystem
 	"read_file":  true,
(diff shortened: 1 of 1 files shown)


mrgoonie (approved):
Summary: This is the right minimal bridge fix: `delegate` was absent from `BridgeToolNames`, and the added entry is consistent with the existing context-driven bridge tools.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear for this small bridge allowlist patch. I found adjacent ACP tool-exposure work (#1072), but no open/merged PR that already makes this exact `BridgeToolNames` delegate addition.
- Project standards: passed. One localized change in `internal/mcp/bridge_server.go`; no schema/config/API expansion beyond exposing an existing self-gated tool.
- Strategic necessity: clear value. Bridge-backed agents with active `agent_links` need parity with native agent delegation; otherwise dashboard orchestration mode can say delegate while the bridge session cannot actually call it.
- CI/checks: not fully green yet; release-versioning and web pass, go is still pending.

Findings:
- Critical: none.
- Important: none.
- Suggestion: after go CI finishes, this is mergeable if it stays green.

Verdict: APPROVE

Issue #1070, opened 2026-04-30
GoClaw MCP test succeeds but save fails for local/private MCP URLs due to SSRF validation mismatch


iamkokono:
# GoClaw MCP test succeeds but save fails for local/private MCP URLs due to SSRF validation mismatch

## Summary

When adding a Streamable HTTP MCP server hosted on a local/private network address, GoClaw can successfully test the connection and discover tools, but saving the same MCP server fails with SSRF validation errors.

This appears to be a validation mismatch between:

- `POST /v1/mcp/servers/test`
- `POST /v1/mcp/servers`

The test endpoint can connect and discover tools, while the create/update endpoint rejects the same URL because it resolves to a private IP.

## Environment

- GoClaw dashboard: `http://localhost:18790`
- GoClaw running in WSL/Linux
- MCP server: Windows-hosted Streamable HTTP server
- MCP health endpoint: `http://127.0.0.1:8765/health`
- MCP endpoint attempted from GoClaw:
  - `http://host.docker.internal:8765/mcp`
  - `http://10.18.231.2:8765/mcp`

Credentials/tokens are intentionally omitted.

## Reproduction Steps

1. Start a Streamable HTTP MCP server on the host machine, listening on port `8765`.

2. Verify health:

   ```text
   GET http://127.0.0.1:8765/health
   => 200 {"ok":true,"server":"windows-host-mcp"}
   ```

3. In GoClaw, test the MCP server with:

   ```json
   {
     "transport": "streamable-http",
     "url": "http://host.docker.internal:8765/mcp",
     "headers": {
       "Authorization": "Bearer [REDACTED]"
     }
   }
   ```

4. Observe that the test succeeds:

   ```json
   {
     "success": true,
     "tool_count": 9
   }
   ```

5. Try to save/create the MCP server using the same transport, URL, and headers.

## Actual Result

Saving fails with SSRF validation errors.

For `host.docker.internal`:

```text
invalid URL: URL validation failed: ssrf: "host.docker.internal" resolved to blocked IP 10.18.231.2
```

For direct private IP:

```text
invalid URL: URL validation failed: ssrf: IP 10.18.231.2 is in a blocked range
```

## Expected Result

Either:

1. The test and save flows should behave consistently. If the URL is not allowed to be saved, the test endpoint should fail with the same validation error.

or preferably for self-hosted/local development:

[...]

PR #1248, opened 2026-06-21
feat(mcp): opt-in allowlist for trusted private MCP hosts

zezaeoh:
## Problem

A remote MCP server (`sse` / `streamable-http`) whose hostname resolves to a **private IP** cannot be registered through the dashboard/API. Create/update validation rejects it via the SSRF guard, and there is **no production-facing config** to permit a specific trusted internal host.

```
invalid URL: URL validation failed: ssrf: "mcp.internal.example.com" resolved to blocked IP 10.x.x.x
```

Key asymmetry: **Test Connection succeeds and lists the server's tools** — the runtime MCP client (`createClient` → `NewStreamableHttpClient` / `NewSSEMCPClient`) connects to the private host with no SSRF guard. Only the **admin-time create/update input validation** (`ValidateServerConfig` → `ValidateURL` → `security.Validate`) blocks it. The only existing bypass is `allowLoopbackForTest`, documented as test-only.

So the destination is already reachable and the runtime connection already works; the rejection is purely at config-validation time and is inconsistent with runtime behavior. Self-hosted MCP servers on a private network are unregisterable.

## Why the block is too coarse here

The SSRF guard exists to stop **agent/attacker-influenced** URLs (`web_fetch`, webhooks, redirect targets) from reaching internal services or the cloud-metadata endpoint. An **MCP server registration is owner/admin configuration** — a deliberate, trusted input. "Self-hosted MCP on our own private network" is a false positive.

## Change

Add an **opt-in, operator-configured allowlist** of trusted MCP hosts exempt from the private-IP block, **scoped to MCP server URL validation only**. Default empty → no behavior change.

- `security.ValidateAllowingHosts(rawURL, allowedHosts)` — like `Validate`, but skips the **private/loopback** block for an allowlisted (case-insensitive, pre-resolution) hostname. Cloud-metadata/link-local (`169.254.0.0/16`, `fe80::/10`), multicast, and unspecified ranges are **never** exempted, even for allowlisted hosts.
- `mcp.SetAllowedHosts` wires the operator allowlist into `ValidateURL` / `ValidateServerConfig` (read at startup from config). `ValidateServerConfig` signature is unchanged, so the five call sites are untouched.
- New gateway config `GOCLAW_MCP_ALLOWED_HOSTS` (comma-separated, default empty).
- `web_fetch` / webhook / redirect SSRF paths are **unchanged** (agent-influenced, higher risk).

### Safety properties

- **Opt-in / default-empty**: zero behavior change unless an operator explicitly allowlists a host.
- **MCP-scoped**: agent-influenced fetch paths are untouched.
- **Metadata stays blocked**: an allowlist entry can never open a path to `169.254.169.254`.
- **Owner-only config**: gateway env, not agent/end-user settable.

> Note on framing: I intentionally did **not** claim "preserve IP pinning" for MCP — the runtime MCP client does not pin the resolved IP today (`ValidateURL` discards the pinned IP and the mcp-go client dials on its own), so there is no MCP-layer pinning to preserve. This PR only aligns config validation with the already-permissive runtime, behind an explicit opt-in. Adding pinning to MCP runtime dials is a separate hardening.

### Alternative considered

A per-server "trusted" boolean (DB + API + UI) instead of a global env. The global env is the least invasive and matches the operator-config nature; happy to switch to per-server if maintainers prefer.

## Tests

- `internal/security`: allowlisted private/loopback IP accepted; non-allowlisted private rejected; metadata endpoint rejected even when allowlisted; nil allowlist == `Validate`; public IP unaffected.
- `internal/mcp`: `SetAllowedHosts` exempts private IP for `ValidateURL`/`ValidateServerConfig`; default still blocks; metadata never exempted; entries normalized (trim+lowercase).
- `go build ./...`, `go vet`, gofmt clean. No migration / schema change.

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -387,6 +387,7 @@ func runGateway() {
 		mcpToolLister = mcpMgr
 	}
 	httpapi.InitGatewayToken(cfg.Gateway.Token)
+	mcpbridge.SetAllowedHosts(cfg.Gateway.MCPAllowedHosts) // operator allowlist: trusted MCP hosts exempt from private-IP SSRF block
 	httpapi.InitGatewayNoAuthFallbackAllowed(config.GatewayNoAuthFallbackAllowed(cfg.Gateway))
 	exportTokenStore := httpapi.InitExportTokenStore()
 	defer exportTokenStore.Stop()
--- a/internal/config/config_channels.go
+++ b/internal/config/config_channels.go
@@ -416,6 +416,7 @@ type GatewayConfig struct {
 	Token                   string              `json:"token,omitempty"`                      // bearer token for WS/HTTP auth
 	OwnerIDs                []string            `json:"owner_ids,omitempty"`                  // sender IDs considered "owner"
 	AllowedOrigins          []string            `json:"allowed_origins,omitempty"`            // WebSocket CORS whitelist (empty = allow all)
+	MCPAllowedHosts         []string            `json:"mcp_allowed_hosts,omitempty"`          // trusted MCP server hostnames exempt from the private-IP SSRF block during config validation (empty = none)
 	MaxMessageChars         int                 `json:"max_message_chars,omitempty"`          // max user message characters (default 32000)
 	RateLimitRPM            int                 `json:"rate_limit_rpm,omitempty"`             // rate limit: requests per minute per user (default 20, 0 = disabled)
 	InjectionAction         string              `json:"injection_action,omitempty"`           // prompt injection action: "log", "warn" (default), "block", "off"
--- a/internal/config/config_load.go
+++ b/internal/config/config_load.go
@@ -311,6 +311,19 @@ func (c *Config) applyEnvOverrides() {
 		c.Gateway.AllowedOrigins = origins
 	}
 
+	// Trusted MCP server hosts from env (comma-separated, whitespace-trimmed).
+	// These hosts are exempt from the private-IP SSRF block when registering MCP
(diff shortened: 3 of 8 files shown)


mrgoonie (approved):
Summary: This PR adds a narrowly scoped, operator-owned MCP host allowlist for private/self-hosted MCP servers and makes the test/save validation paths consistent by routing test connection through ValidateServerConfig too. I checked the changed security and MCP validation paths against the existing SSRF boundary and issue #1070; default behavior remains closed, web_fetch/webhook validation is unchanged, and metadata/link-local targets remain blocked even when listed.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: clear; it directly addresses open issue #1070 and I did not find another merged/open implementation of the MCP allowed-hosts path.
- Project standards: passed; config/env override wiring follows existing GatewayConfig/applyEnvOverrides patterns, and MCP validation stays in internal/mcp with shared SSRF logic in internal/security.
- Strategic necessity: clear value; self-hosted/private MCP servers are a common owner-admin workflow, and this fixes the confusing test-succeeds/save-fails mismatch without weakening agent-influenced fetch paths.
- CI/checks: green; release-versioning, go, and web checks passed.

Findings:
- Critical: none.
- Important: none.
- Suggestion: document GOCLAW_MCP_ALLOWED_HOSTS in operator-facing config docs in a follow-up if not already covered elsewhere, including the warning that it is an admin-only trust decision and not a general SSRF bypass.

Verdict: APPROVE

Issue #1064, opened 2026-04-28
WhatsApp: multiple channel_instances share one whatsmeow_device row


kamushadenes:
## Bug

Each WhatsApp `channel_instance` calls `container.GetFirstDevice(ctx)` on a shared `sqlstore.Container` (one Postgres DB). When two or more WhatsApp instances exist, every channel binds to the same first device row in `whatsmeow_device` and connects as the same WhatsApp account, regardless of the instance name or agent it is wired to.

## Repro

1. Configure one WhatsApp `channel_instance` (e.g. `whatsapp` → agent A). Pair via the QR wizard. Works fine.
2. Add a second WhatsApp `channel_instance` (e.g. `whatsapp-B` → agent B) intending to bind a different phone number.
3. Restart the gateway (or hit "Start QR" on the second instance).

## Observed

```
time=... level=INFO msg="channel instance loaded" name=whatsapp        type=whatsapp ...
time=... level=INFO msg="channel instance loaded" name=whatsapp-B      type=whatsapp ...
time=... level=WARN msg="whatsapp QR: start flow failed" error="whatsapp get QR channel: GetQRChannel must be called before connecting"
time=... level=INFO msg="whatsapp: connected" jid=<accountA-jid> channel=whatsapp
time=... level=INFO msg="whatsapp: connected" jid=<accountA-jid> channel=whatsapp-B   ← same JID!
```

`whatsmeow_device` has only one row. `whatsapp-B` inherits account A's `Store.ID`, so `whatsmeow.Connect()` short-circuits the QR flow and `GetQRChannel` then refuses (the device is already paired).

## Expected

Each `channel_instance` should own its own device row in `whatsmeow_device`, and a fresh instance with no prior pairing should bring up a new QR session bound to that instance.

## Code path

- `internal/channels/whatsapp/whatsapp.go` `Channel.Start` calls `c.container.GetFirstDevice(ctx)`.
- `internal/channels/whatsapp/auth.go` `Channel.StartQRFlow` and `Channel.Reauth` likewise.
- `cmd/gateway.go` registers a single `whatsapp.FactoryWithDBAudio(...)` shared by all WhatsApp instances; the factory has no notion of which instance it is constructing for, so it cannot scope the device lookup.

## Fix proposal

Persist the paired JID on `channel_instances.config` (key `"jid"`) and resolve the device per channel:

1. Read `config.jid` in the factory; pass it + `ChannelInstanceStore` into `whatsapp.New`.
[...]

PR #1254, opened 2026-06-22
Fix/whatsapp connect

haidang1810:
## Type

- [ ] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Checklist
- [ ] `go build ./...` passes
- [ ] `go vet ./...` passes


mrgoonie (changes requested):
Summary: This PR improves WhatsApp QR/deleted-device recovery and adds useful inbound logging/tests, but it does not fix the main WhatsApp multi-instance connection failure and it appears to weaken the existing inbound allowlist safety net.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: overlap found with #1064 / #1065 for the shared WhatsApp device-store root cause.
- Project standards: issue found — the channel inbound safety pattern still needs an allowlist/policy guard equivalent to `BaseChannel.HandleMessage` for DMs.
- Strategic necessity: clear value, but incomplete for the reported connect class.
- CI/checks: green

Findings:
- Important: `internal/channels/whatsapp/auth.go` and `internal/channels/whatsapp/whatsapp.go` still call `c.container.GetFirstDevice(ctx)`. That keeps the root multi-instance WhatsApp bug from #1064: multiple `channel_instances` can reuse the first `whatsmeow_device` row and connect as the same account instead of getting an instance-scoped QR/device. #1065 already documents the needed shape: persist/resolve a per-instance JID or otherwise scope device selection by channel instance. This PR refreshes deleted stores, but it does not remove the shared-first-device behavior.
- Important: `internal/channels/whatsapp/inbound.go` removes the explicit `if !c.IsAllowed(senderID) { ... return }` gate before publishing inbound messages. `BaseChannel.HandleMessage` still enforces `IsAllowed` for direct messages as a safety net, but WhatsApp bypasses that helper and publishes directly to the bus. With the default DM policy now treated as `pairing`, this may be okay for paired senders, but for `DMPolicy=open` it changes allowlist behavior without a product/security note. Please either restore the allowlist safety check for DMs or document/test the intended precedence between `allow_from`, DM policy, and pairing.

Verdict: REQUEST_CHANGES

Next step: scope WhatsApp device lookup per channel instance (or explicitly close the #1064 overlap with evidence) and add/adjust tests proving `allow_from` is still enforced as intended for WhatsApp DMs after this change.

*Posted by /github-maintain at 2026-06-22T04:51:21Z*

mrgoonie (approved):
Summary: Re-reviewed the latest head after the WhatsApp device scoping and inbound policy updates. The previous blockers are addressed: DB-backed WhatsApp instances now resolve/persist a scoped device JID instead of reusing the first global whatsmeow device, and the direct-message allowlist behavior is now explicitly covered for open DM policy.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: overlap with #1064/#1065 is now handled by instance-scoped device_jid persistence and fresh QR device creation when the scoped credential is missing.
- Project standards: passed — config-file WhatsApp keeps legacy first-device fallback, while DB channel instances use credential-scoped device selection.
- Strategic necessity: clear value for multi-instance WhatsApp reliability and deleted-device QR recovery.
- CI/checks: green

Findings:
- Critical: none
- Important: none blocking
- Suggestion: after merge, consider a follow-up docs/admin note explaining that DB-backed WhatsApp instances persist device_jid in encrypted credentials after QR login.

Verdict: APPROVE

*Posted by /github-maintain at 2026-06-22T06:05:00Z*

PR #1192, opened 2026-06-07
fix(agent): check nil td.Function to prevent panic on native tools

thotam:
Fix nil pointer dereference panic when agent processes native tools (such as image_generation) which do not contain function schema wrappers.

Key changes:
- loop_pipeline_callbacks.go & think_stage.go: add checks for td.Function != nil before accessing td.Function.Name.
- loop_tool_filter.go & loop_history_toolnames.go: add safety checks when filtering and extracting tool names.
- anthropic_request.go & codex_build.go: verify t.Function != nil when formatting tools for provider API requests.

## Summary
This PR fixes a nil pointer dereference panic that occurs when the agent processes native tools (such as `image_generation`) which have `Type` other than `"function"` and `Function: nil`. It introduces safety checks across loop pipeline callbacks, think stages, tool filters, history toolnames, Anthropic requests, and Codex builders.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Hotfix (targeting `main`)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
- Target branch: `dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go build -tags sqliteonly ./...` passes (if Go changes)
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` (or standard test suites)
- [ ] Web UI builds: `cd ui/web && pnpm build` (N/A)
- [x] No hardcoded secrets or credentials
- [x] SQL queries use parameterized `$1, $2` (N/A)
- [x] New user-facing strings added to all 3 locales (en/vi/zh) (N/A)
- [ ] Migration version bumped in `internal/upgrade/version.go` (N/A)

## Test Plan
- Verified compiling checks with both standard and `sqliteonly` tags.
- Executed unit tests for `internal/agent/...`, `internal/pipeline/...`, and `internal/providers/...`. All tests passed successfully without regressions.

PR #1256, opened 2026-06-22
fix(agent): nil td.Function panic on native tools — hotfix backport of #1192 to main

aaron-tsar:
## Summary
Backport of #1192 to `main`. The released `v3.14.0` images (`:latest` / `:full`, built from `main`) crash with an **unrecovered `SIGSEGV` nil-pointer dereference** whenever an agent run uses a provider that injects **native tool definitions** (e.g. `image_generation` from the Codex / OpenAI provider). Native tools have `Type != "function"` and therefore a `nil` `ToolDefinition.Function`, but several hot paths dereference `td.Function.Name` without a nil check. The panic is on the chat-dispatch goroutine, so it takes down the **whole gateway process**, not just the run. The fix is already on `dev` (#1192) but had not reached `main`.

Observed in production (`provider=openai-codex`, `model=gpt-5.5`):
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 ...]
internal/agent.(*Loop).pipelineCallbacks...makeBuildFilteredTools.func9
    internal/agent/loop_pipeline_callbacks.go:240
```
(The "trace finalized by safety net" WARN seen alongside it is just the deferred trace finalizer firing during stack unwinding — a symptom, not the cause.)

## Type
- [x] Hotfix (targeting `main`)

## Target Branch
`main` — the bug only remains on `main`; `dev` already contains the fix (#1192). Per the hotfix policy this targets `main` directly; no cherry-pick back to `dev` is needed since the fix originated there.

## What changed
Cherry-pick of #1192 (`389640ae`). Adds `td.Function != nil` / `t.Function == nil` guards before dereferencing in every tool-iteration site that native tools flow through:
- `loop_pipeline_callbacks.go` (mcp-tool counting + `shouldRetryTaskMCP`)
- `think_stage.go`, `loop_tool_filter.go`, `loop_history_toolnames.go`
- `anthropic_request.go`, `codex_build.go`
Plus the regression tests from #1192.

**Backport note:** the only deviation from #1192 is in `image_gen_gate_test.go` — one `buildFilteredTools(...)` call was reverted to the 5-arg signature, because the `userTools` parameter (added by #1235 on `dev`) does not exist on `main`. Production code is identical to #1192.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes (agent/pipeline/providers)
- [x] Tests pass: `go test ./internal/agent/ ./internal/pipeline/ ./internal/providers/`
- [ ] Web UI builds — N/A, backend-only change
- [x] SQL parameterized — N/A, no SQL changes
- [ ] i18n locales — N/A, no user-facing strings
- [ ] Migration version bump — N/A, no migration

Surface parity: web UI / CLI / API contract N/A — this is a backend nil-guard with no schema, protocol, or user-facing surface change.

## Test Plan
1. Added regression tests (from #1192) assert no panic when native (`image_generation`) tool defs are mixed into the tool list across `buildFilteredTools`, `ThinkStage`, and the Anthropic/Codex request builders. All pass.
2. Reproduced the original crash on `main` with a Codex provider + `image_generation` enabled agent (`gpt-5.5`); with this change the run completes instead of segfaulting the gateway.

### DIFF
--- a/internal/agent/image_gen_gate_test.go
+++ b/internal/agent/image_gen_gate_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 
 	"github.com/nextlevelbuilder/goclaw/internal/providers"
+	"github.com/nextlevelbuilder/goclaw/internal/tools"
 )
 
 // imageCapableProvider is a stub provider that also implements CapabilitiesAware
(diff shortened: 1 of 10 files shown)


aaron-tsar:
Closing: the fix (#1192) is already merged on `dev`, and we've moved this deployment to the `dev`/beta build (v3.15.0-beta.21) which includes it. The backport to `main` is no longer needed for our environment. Note for maintainers: the nil-`Function` panic on native tools still exists on `main`/released v3.14.0 images until the next dev→main release — feel free to reopen/cherry-pick if a hotfix to `main` is wanted before then.

Issue #1217, opened 2026-06-12
[Security] Cross-Agent Authorization Bypass in `/v1/tools/invoke` Allows Unauthorized Cron Binding to Foreign Agents


YLChen-007:
### Advisory Details

**Title**: Cross-Agent Authorization Bypass in `/v1/tools/invoke` Allows Unauthorized Cron Binding to Foreign Agents

**Description**:

### Summary

An authenticated caller with only `operator.write` privileges can use the documented HTTP tool invocation endpoint, `POST /v1/tools/invoke`, to create a cron job bound to another user’s private agent. Direct reads of that same foreign agent are correctly blocked with `403`, but the cron creation path accepts an attacker-controlled `agentId` without an access check and persists the unauthorized binding into `cron_jobs.agent_id`. This creates a stored cross-agent authorization bypass that the scheduler later trusts as the execution target.

### Details

The defect is in the HTTP direct-tool entry point. In `internal/http/tools_invoke.go`, the handler accepts top-level request field `agentId`, resolves it with `agentStore.GetByKey`, and injects the resolved UUID into context with `store.WithAgentID`, but never calls `CanAccess` before doing so.

```go
if agentIDStr != "" && h.agentStore != nil {
    ag, err := h.agentStore.GetByKey(ctx, agentIDStr)
    if err == nil {
        ctx = store.WithAgentID(ctx, ag.ID)
    }
}
```

That matters because direct agent reads do enforce authorization. In `internal/http/agents.go`, `GET /v1/agents/{id}` checks `h.agents.CanAccess(...)` and returns `403` when the caller is not allowed to access the target agent.

Once the attacker-controlled agent UUID is present in context, the `cron` tool consumes it in `internal/tools/cron.go` through `resolveAgentIDString(ctx)` and passes it into `handleAdd(...)`. The same function also accepts an explicit `job.agentId` override, which can replace the context agent again before persistence:

```go
// Use agent ID from job object if explicitly provided, otherwise from context
if explicit, _ := jobObj["agentId"].(string); explicit != "" {
    agentID = explicit
}

job, err := t.cronStore.AddJob(ctx, name, schedule, message, deliver, channel, to, agentID, userID)
```

The SQLite cron store then writes that UUID directly into the persisted cron row:

```go
var agentUUID *uuid.UUID
if agentID != "" {
    if aid, err := uuid.Parse(agentID); err == nil {
        agentUUID = &aid
    }
}

[...]

PR #1260, opened 2026-06-22
fix(agent): strip whole text-encoded tool-call blocks, not just tags

zezaeoh:
## Problem

`SanitizeAssistantContent` runs `stripGarbledToolXML` on every provider's output to clean up tool-call XML that some models (DeepSeek/GLM/Minimax) emit as text. The cleanup (`garbledToolXMLPattern`) removes only the *tags*. When a model emits a **complete** Anthropic-style tool-call block as text —

```
<function_calls>
<invoke name="send_message">
<parameter name="channel">argus</parameter>
<parameter name="text">PR #1217 approved</parameter>
</invoke>
</function_calls>
```

— two things go wrong:

1. **Argument values leak.** Stripping only the tags leaves the inner `<parameter>` text (`argus`, `PR #1217 approved`) orphaned in the user-facing reply.
2. **The failure is silent.** A complete block means the model *wrote a tool call as text instead of invoking it natively*, so the tool never ran. The only trace is a generic `stripped garbled tool call XML` warn with a length delta — no indication of which tool was lost.

This is most damaging with the `claude-cli` thin-proxy provider, where tool execution lives inside the CLI and goclaw only sees the final text. A scheduled agent that emits its result-reporting call as text completes "successfully" while silently doing nothing — no error, no delivered action, and the same work re-triggers next cycle because state was never updated.

## Fix

In `stripGarbledToolXML`, handle a complete `<function_calls>...</function_calls>` block before the tag-only strip:

- **Remove the whole block** (tags + inner `<parameter>` text) so argument values no longer leak into the reply.
- **Log the attempted tool name(s)** (parsed from `<invoke name="...">`) at WARN, so the otherwise-invisible no-op is diagnosable.

Partial/malformed artifacts (no closing tag) fall through to the existing tag-level strip unchanged, so DeepSeek/GLM/Minimax handling is preserved. The trailing `stripped garbled tool call XML` warn now fires only when the tag strip actually changed something. No effect on providers that emit proper structured tool calls (their assistant text carries no tool-call XML).

## Test

New `internal/agent/sanitize_toolcall_xml_test.go`:

- whole block removed, surrounding prose preserved
- block-only response collapses to empty
- `SanitizeAssistantContent` no longer leaks `<parameter>` argument values (regression)
- partial stray tag still stripped (backward compat)

```
go vet ./internal/agent/            ✅
go test -race ./internal/agent/     ✅
```

### DIFF
--- a/internal/agent/sanitize.go
+++ b/internal/agent/sanitize.go
@@ -82,6 +82,20 @@ var garbledToolXMLPattern = regexp.MustCompile(
 	`(?s)</?(?:function_calls?|functioninvoke|invoke|invfunction_calls|tool_call|tool_use|parameter|minimax:tool_call)[^>]*>`,
 )
 
+// fullToolCallBlockPattern matches a COMPLETE Anthropic-style tool-call block
+// (`<function_calls>...</function_calls>`) that a model emitted as text instead
+// of invoking it natively. Unlike garbledToolXMLPattern — which removes only the
+// tags — this captures the whole block, tags plus the inner <parameter> text, so
+// a stray block is removed cleanly rather than leaving the argument values
+// orphaned in the user-facing reply.
+var fullToolCallBlockPattern = regexp.MustCompile(
+	`(?is)<function_calls?>.*?</function_calls?>`,
+)
+
+// invokeNamePattern extracts the tool name from an `<invoke name="...">` tag so a
+// dropped text-encoded tool call can be logged with the tool it tried to call.
+var invokeNamePattern = regexp.MustCompile(`(?i)<invoke\s+name="([^"]+)"`)
+
 var garbledToolXMLIndicators = []string{
 	"invfunction_calls",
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Summary: This is a focused agent sanitization fix that removes complete text-encoded tool-call blocks instead of only stripping XML tags, preventing leaked parameter values and adding useful diagnostics for silently dropped text tool calls.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; search found no other issue/PR for this exact sanitizer behavior.
- Project standards: passed; small change in internal/agent with regression tests.
- Strategic necessity: clear value; prevents user-facing argument leakage and makes failed text-encoded tool calls diagnosable.
- CI/checks: green (go, web, release-versioning).

Findings:
- Critical: none.
- Important: none.
- Suggestion: none.

Verdict: APPROVE

PR #1261, opened 2026-06-23
fix(agent): also strip bare <invoke> tool-call blocks (no function_calls wrapper)

zezaeoh:
## Problem

Follow-up to #1260. That PR removes a complete `<function_calls>...</function_calls>` block a model emitted as text. But in production the claude-cli proxy (under a degraded/bloated session) emits the tool call as a **bare** `<invoke name="...">...</invoke>` block with **no `<function_calls>` wrapper**:

```
Format error, retrying.
<invoke name="mcp__goclaw-bridge__exec">
<parameter name="command">( cd ~/path && gh pr view 1218 )</parameter>
</invoke>
```

`fullToolCallBlockPattern` (wrapper-only) misses this, so it falls through to the tag-only strip — which deletes the tags but leaks the inner `<parameter>` command text into the reply, while the tool still never runs (silent no-op).

## Fix

- Add `bareInvokeBlockPattern` and remove complete bare `<invoke>...</invoke>` blocks too, **after** the wrapped form (so wrapper-nested invokes aren't double-handled).
- Add `<invoke name=` to `garbledToolXMLIndicators` so even a bare invoke triggers the cleanup gate.
- Both block types feed the single `dropped text-encoded tool call` WARN with the attempted tool name(s).

Unterminated/partial artifacts still fall through to the existing tag-level strip (DeepSeek/GLM/Minimax behavior unchanged). No effect on providers that emit proper structured tool calls.

## Test

`internal/agent/sanitize_toolcall_xml_test.go` extended:

- bare `<invoke>` block removed, surrounding prose kept
- bare-invoke-only response collapses to empty
- `SanitizeAssistantContent` does not leak the bare-invoke command argument (regression)

```
go vet ./internal/agent/            ✅
go test -race ./internal/agent/     ✅
```

### DIFF
--- a/internal/agent/sanitize.go
+++ b/internal/agent/sanitize.go
@@ -96,9 +96,20 @@ var fullToolCallBlockPattern = regexp.MustCompile(
 // dropped text-encoded tool call can be logged with the tool it tried to call.
 var invokeNamePattern = regexp.MustCompile(`(?i)<invoke\s+name="([^"]+)"`)
 
+// bareInvokeBlockPattern matches a complete `<invoke name="...">...</invoke>`
+// block emitted as text WITHOUT the surrounding <function_calls> wrapper. Some
+// models — and the claude-cli proxy under a degraded session — drop the wrapper
+// and emit just the invoke block, which fullToolCallBlockPattern misses; the
+// tag-only strip would then leak the inner <parameter> text. Applied after the
+// wrapped form so wrapper-nested invokes are already gone.
+var bareInvokeBlockPattern = regexp.MustCompile(
+	`(?is)<invoke\s+name="[^"]*".*?</invoke>`,
+)
+
 var garbledToolXMLIndicators = []string{
 	"invfunction_calls",
 	"functioninvoke",
+	"<invoke name=",
 	"<parameter name=",
 	"</parameter",
 	"<function_call",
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Summary: This is a focused follow-up to the wrapped function_calls sanitizer. It removes complete bare <invoke>...</invoke> text blocks before the fallback tag stripping path, so command parameters do not leak into user-facing replies when a provider emits a tool call as text.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; search only found this PR for the bare-invoke no-wrapper case, while prior work handled the wrapped block form.
- Project standards: passed; the change stays in internal/agent sanitizer code and adds targeted regression coverage.
- Strategic necessity: clear value; this closes a real leakage/no-op failure mode in degraded tool-call output handling.
- CI/checks: green.

Findings:
- Critical: none.
- Important: none.
- Suggestion: none.

Verdict: APPROVE

Reviewed evidence: diff limited to internal/agent/sanitize.go and internal/agent/sanitize_toolcall_xml_test.go; CI release-versioning, go, and web checks are successful; added tests cover bare invoke removal and command-argument non-leakage.

*Posted by /github-maintain at 2026-06-23T00:00:00Z*

PR #1262, opened 2026-06-23
fix(cron): reset session for stateless cron jobs, not stateful ones

zezaeoh:
## Problem

Cron runs reset the session before each run (so a previous run's tool errors can't pollute the next — #294). That reset was gated on `if !job.Stateless`, which **inverts the flag**:

- `stateless = true` (the default, documented as "saves tokens" / fresh each run) → reset **skipped** → session **accumulates unbounded history** across runs.
- `stateless = false` (stateful, wants memory) → reset **runs every time** → session **wiped each run**.

Observed in production: a frequently-firing `stateless=true` cron grew its Claude CLI session `.jsonl` to ~30 MB, replaying the whole history on every run (large token cost) until the model started emitting malformed tool calls.

There's a second layer: the cron reset only cleared the **goclaw session store**, never the **Claude CLI on-disk session**. claude-cli resumes its own `.jsonl` by a deterministic per-key UUID (`--resume`), so even when the reset fired the model still saw the full accumulated history — i.e. `stateless` had no effect on the actual model context for claude-cli agents.

## Fix

- Gate the reset on `if job.Stateless` (correct direction).
- In that branch, also clear the CLI session via `ResetCLISession` (indirected through an overridable package var for testing), so a stateless run truly starts fresh on both layers.
- Guard `sessionMgr != nil` (it is nil in some unit-test setups).

Stateful jobs (`stateless=false`) now correctly keep their session across runs. Non-claude-cli providers are unaffected — `ResetCLISession` is a no-op when no CLI session file exists.

## Test

`cmd/gateway_cron_test.go`: new `TestCronJobHandler_StatelessResetsSession` asserts a stateless run clears both the session store and the CLI session, and a stateful run clears neither (via a fake `SessionStore` + an overridden `cronCLISessionReset`).

```
go vet ./cmd/                        ✅
go test -race ./cmd/                 ✅
```

## Note for operators

Behavior change: any cron currently set `stateless=false` that was relying on the (buggy) per-run wipe will now retain its session. Re-check the flag on existing crons — set `stateless=true` for jobs that should start fresh each run.

### DIFF
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -25,6 +25,11 @@ import (
 // Safe because cron jobs only fire after Start(), well after this is set.
 var cronHeartbeatWakeFn func(agentID string)
 
+// cronCLISessionReset clears the Claude CLI on-disk session (.jsonl + CLAUDE.md)
+// for a session key, mirroring the sessions.reset RPC. Indirected through a var
+// so the stateless-reset behavior can be unit-tested without filesystem effects.
+var cronCLISessionReset = providers.ResetCLISession
+
 func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg *config.Config, channelMgr *channels.Manager, sessionMgr store.SessionStore, agentStore store.AgentStore, providerStore store.ProviderStore, providerReg *providers.Registry) func(job *store.CronJob) (*store.CronJobResult, error) {
 	return func(job *store.CronJob) (*store.CronJobResult, error) {
 		agentID := job.AgentID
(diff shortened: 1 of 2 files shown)


mrgoonie (approved):
Summary: This is a focused cron-session fix. It corrects the stateless gate so stateless cron runs reset both the GoClaw session store and the Claude CLI on-disk session, while preserving stateful cron history.

Risk level: Medium

Mandatory gates:
- Duplicate/prior implementation: overlap found with older open PR #1032, but this PR is the stronger successor because it preserves stateful jobs and also clears the Claude CLI session layer.
- Project standards: passed; the change is localized to cmd/gateway_cron.go with targeted tests in cmd/gateway_cron_test.go.
- Strategic necessity: clear value; prevents stateless cron token/history accumulation and fixes the inverted stateful/stateless behavior.
- CI/checks: green.

Findings:
- Critical: none.
- Important: none.
- Suggestion: after merge, #1032 should be closed as superseded by this implementation.

Verdict: APPROVE

PR #1111, opened 2026-05-06
fix(gateway): re-apply tool rate limiter after system_configs overlay

nguyennguyenit:
## Summary
- `setupToolRegistry` initialises the tool rate limiter from `cfg.Tools.RateLimitPerHour` early in startup (gateway.go line 137), but `system_configs` DB overlay (`cfg.ApplySystemConfigs`) runs ~50 lines later. DB overrides of `tools.rate_limit_per_hour` were silently lost - the limiter object was already built from the JSON5 default.
- Re-apply the limiter after `ApplySystemConfigs` so the DB value wins. Server has not started yet, no in-flight tool calls. `nil` case handled too so DB writes can disable rate limiting without restart.

## Why this matters
Production observation: a Zalo sales-bot session burst past the 150/h default and operators tried to bump it via `system_configs` UPDATE + container restart. No effect. Workaround was to inject `/app/data/config.json` with `{"tools":{"rate_limit_per_hour":800}}`, which contradicts the DB-as-source-of-truth pattern other tunables rely on.

## Tests
- New: `TestRegistry_SetRateLimiter_ReplacesPriorLimiter` covers the replace-with-higher-limit path and the `nil`-disables path.
- All existing rate limiter tests still pass.
- `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./...` clean.

## Out of scope
The same ordering pattern likely affects other config consumed inside `setupToolRegistry` (`tools.scrub_credentials`, MCP server wiring from DB). Those need a broader restructure - kept out of this hotfix.

## Test plan
- [ ] Merge to `dev`, deploy, set `UPDATE system_configs SET value='800' WHERE key='tools.rate_limit_per_hour'`, restart container, verify log line `tool rate limiting reapplied from system_configs per_hour=800`.
- [ ] Confirm the existing `tool rate limiting enabled per_hour=<json5_default>` log still fires earlier - the reapply log is additive.
- [ ] Set DB value to `0`, restart, verify limiter is disabled (no `tool rate limit exceeded` errors regardless of call volume).

PR #1263, opened 2026-06-23
feat(agent): honor per-agent tools.rate_limit_per_hour

zezaeoh:
## Problem

The agent config dashboard (`tools-profile-section.tsx`) and i18n (en/vi/zh) already render a per-agent **"Rate Limit (per hour)"** input that saves `rate_limit_per_hour` into the agent's `tools_config`. But the backend never read it: the tool rate limiter (`ToolRateLimiter`) only used the global `tools.rate_limit_per_hour` (config / `system_configs`), keyed per session. So an operator could set a per-agent limit in the UI and nothing happened — the only knob that worked was the global one (shared by every agent).

## Fix

Wire the existing field through, end to end:

- `config.ToolPolicySpec` gains `RateLimitPerHour int` (`rate_limit_per_hour`), parsed from the per-agent `tools_config` JSONB — no migration (additive).
- The agent loop (`loop_context.go`) threads it into the tool-execution context (`WithToolRateLimitOverride`) when set, mirroring the existing per-agent `wait` config.
- `Registry.ExecuteWithContext` reads the override from context and passes it to the limiter.
- `ToolRateLimiter.AllowWithLimit(key, maxOverride)` uses `maxOverride` (calls/hour) when `> 0`, else the configured global; `Allow(key)` delegates with `0`, so existing callers are unchanged.
- `ui/web/src/types/agent.ts`: add `rate_limit_per_hour?` to `ToolPolicyConfig` (the UI already binds it via `Record<string, any>`).

`0`/unset inherits the global `tools.rate_limit_per_hour`. The limit stays per-session-key, so a per-agent value just gives that agent more headroom without affecting others.

## Test

`internal/tools/rate_limiter_test.go` → `TestToolRateLimiter_AllowWithLimit_Override`: an override of 2 caps a key at 2 regardless of the global 100; `maxOverride <= 0` falls back to the global.

```
go build ./...                                                  ✅
go build -tags sqliteonly ./...                                 ✅
go vet ./internal/tools/ ./internal/agent/ ./internal/config/   ✅
go test -race ./internal/tools/ ./internal/agent/               ✅
```

No migration / schema change. i18n keys already present in en/vi/zh.

### DIFF
--- a/internal/agent/loop_context.go
+++ b/internal/agent/loop_context.go
@@ -113,6 +113,9 @@ func (l *Loop) injectContext(ctx context.Context, req *RunRequest) (contextSetup
 		waitToolCfg = l.agentToolPolicy.Wait
 		ctx = tools.WithWaitToolConfig(ctx, waitToolCfg)
 	}
+	if l.agentToolPolicy != nil && l.agentToolPolicy.RateLimitPerHour > 0 {
+		ctx = tools.WithToolRateLimitOverride(ctx, l.agentToolPolicy.RateLimitPerHour)
+	}
 	if l.sandboxCfg != nil {
 		ctx = tools.WithSandboxConfig(ctx, l.sandboxCfg)
 	}
--- a/internal/config/config_channels.go
+++ b/internal/config/config_channels.go
@@ -537,6 +537,9 @@ type ToolPolicySpec struct {
 	ByProvider     map[string]*ToolPolicySpec `json:"byProvider,omitempty"`
 	Wait           *WaitToolPolicy            `json:"wait,omitempty"`
 	ToolCallPrefix string                     `json:"toolCallPrefix,omitempty"` // prefix to strip from model's tool call names before registry lookup
+	// RateLimitPerHour overrides the global tools.rate_limit_per_hour for this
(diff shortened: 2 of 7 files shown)


mrgoonie (approved):
Summary: This is a focused backend wiring fix for the existing per-agent tools.rate_limit_per_hour setting. The per-agent policy value is parsed from tools_config, threaded into the tool execution context, and applied at the existing per-session limiter with regression coverage for override-vs-global behavior.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; the only prior related work I found is #1111, which re-applies the global limiter after system_configs overlay rather than honoring the per-agent field.
- Project standards: passed; the change follows the existing ToolPolicySpec/agent-loop/context-key/registry patterns already used for wait/tool config propagation.
- Strategic necessity: clear value; it makes an already exposed dashboard setting actually enforceable without a migration or schema break.
- CI/checks: green (release-versioning, go, web).

Findings:
- Critical: none.
- Important: none.
- Suggestion: a future follow-up could add an integration-style registry/agent-loop test proving tools_config JSON flows all the way into ExecuteWithContext, but the current unit coverage is enough for this small additive path.

Verdict: APPROVE

PR #1272, opened 2026-06-24
fix(agent): recover interrupted runs to failed on gateway startup

zezaeoh:
## Summary

Runs killed mid-execution by a gateway stop are never recorded as failed and show as perpetually **running**. A run only emits its terminal `run.status` (completed/failed/cancelled) when the agent loop returns; if the process is killed mid-run (restart, crash, or SIGKILL after a slow graceful shutdown), that terminal event never fires, so the run's last `run.status` item stays `started` forever.

This adds a startup reconciliation — `RunTimelineStore.RecoverInterruptedRuns`, run once on gateway boot — that finds runs with a `started` `run.status` but no terminal sibling and appends a terminal `failed` `run.status` item (marked `interrupted` in metadata). It mirrors the cron scheduler's existing `recomputeStaleJobs`, which already flips stale `running` jobs to `interrupted` on startup. Because it runs only at startup, nothing from before is still executing, so there is **no false-positive risk** — unlike a periodic sweep (cf. the deliberately-disabled `tracing` stale-trace loop, which lacks a way to tell a crash from a legitimately long run).

Observed in production: a gateway restart while an agent run was in flight left the run stuck at `started`; a scan turned up 11 such orphaned runs accumulated across past restarts/crashes.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race` on the affected packages (`internal/store/sqlitestore`, `internal/agent`, `internal/gateway/methods`, `internal/http`)
- [x] SQL queries — static, no string concatenation and no caller/user input (cross-tenant by design; appended rows carry each run's own `tenant_id`)
- [x] No new user-facing strings — `"Run failed"` matches the existing un-localized `run.status` titles (`"Run started"`, etc.) emitted by the recorder
- [x] No new migration — appends to the existing `run_timeline_items` event log only

## Test Plan

- New unit test `TestSQLiteRunTimelineStoreRecoverInterruptedRuns` covers:
  - an interrupted run (`started` + an in-flight `tool.call`, no terminal) gains a terminal `failed` `run.status` at `max(seq)+1`;
  - a completed run is left untouched;
  - the pass is **idempotent** — a second run recovers 0.
- `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./...` all clean.

## Notes for review

- **Tenant isolation:** this is an internal startup job (no request context, no `RoleAdmin`), intentionally cross-tenant like `recomputeStaleJobs`. The detection query reads across tenants, but each appended terminal row is written with the originating run's own `tenant_id`, so timeline data stays correctly tenant-scoped.
- **Status choice:** reuses the existing `failed` status rather than introducing a new `interrupted` enum value, to avoid churn in the UI/status rendering. The `metadata` field (`{"interrupted":true,"reason":"server_restart"}`) keeps it distinguishable from a genuine agent failure. This matches the precedent set by `tracing.RecoverStaleRunningTraces`, which reuses its terminal `error` status.

### DIFF
--- a/cmd/gateway_managed.go
+++ b/cmd/gateway_managed.go
@@ -206,6 +206,17 @@ func wireExtras(
 		slog.Info("agent hooks dispatcher wired", "handlers", "command,http,prompt")
 	}
 	timelineRecorder := agent.NewRunTimelineRecorder(stores.RunTimeline)
+	// Reconcile runs left mid-execution by a previous gateway stop. A run whose
+	// process was killed never emits its terminal run.status, so it would show as
+	// perpetually "running" and never be recorded as failed. Mark such runs failed
+	// on startup, mirroring the cron scheduler's stale-'running' reset.
+	if stores.RunTimeline != nil {
+		if n, err := stores.RunTimeline.RecoverInterruptedRuns(context.Background()); err != nil {
+			slog.Warn("run timeline: failed to recover interrupted runs on startup", "error", err)
+		} else if n > 0 {
+			slog.Info("run timeline: marked interrupted runs as failed on startup", "count", n)
+		}
+	}
 
 	resolver := agent.NewManagedResolver(agent.ResolverDeps{
 		AgentStore:             stores.Agents,
--- a/internal/agent/run_timeline_recorder_test.go
+++ b/internal/agent/run_timeline_recorder_test.go
@@ -28,6 +28,10 @@ func (s *recordingTimelineStore) ListRunTimelineItems(context.Context, store.Run
 	return nil, nil
 }
 
+func (s *recordingTimelineStore) RecoverInterruptedRuns(context.Context) (int64, error) {
+	return 0, nil
+}
+
 func TestRunTimelineItemFromEventScrubsToolArguments(t *testing.T) {
 	tenantID := uuid.Must(uuid.NewV7())
 	item, ok := runTimelineItemFromEvent(AgentEvent{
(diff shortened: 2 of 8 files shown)


mrgoonie (approved):
Summary: This is a focused startup reconciliation for agent runs left with a started run.status after gateway interruption. It adds cross-tenant startup recovery in both PG and SQLite timeline stores and covers the SQLite idempotency path.

Risk level: Low

Mandatory gates:
- Duplicate/prior implementation: clear; no existing PR/issue or git history for RecoverInterruptedRuns / run timeline interrupted recovery found.
- Project standards: passed; implementation follows the existing store interface/tenant context patterns and mirrors the cron stale-running recovery precedent.
- Strategic necessity: clear value; it prevents orphaned runs from staying permanently running after restarts/crashes without adding a periodic false-positive sweeper.
- CI/checks: green on GitHub (go, web, release-versioning). Local verification passed: go test ./internal/store/pg -run RunTimeline; go test -tags sqliteonly ./internal/store/sqlitestore -run RecoverInterruptedRuns; go test ./internal/agent ./internal/gateway/methods ./internal/http.

Findings:
- Critical: none.
- Important: none.
- Suggestion: consider adding a small PG-specific RecoverInterruptedRuns test later for parity with the SQLite coverage, but the SQL shape is close and the local PG RunTimeline tests pass, so this is not a blocker.

Verdict: APPROVE

PR #1274, opened 2026-06-24
fix(gateway): re-apply allowed_paths to filesystem tools after system_configs overlay

zezaeoh:
## Summary
`system_configs['allowed_paths']` never reached the filesystem tools, so agents were denied access to configured shared directories outside their workspace even though the DB value was present. This re-applies the user-configured allowed paths after the `system_configs` overlay — the AllowPaths analogue of the rate-limiter re-apply in #1111.

## Type
- [x] Bug fix

## Root cause
`setupToolRegistry` wires the filesystem tools' `AllowPaths` from the config/JSON5 default **before** `ApplySystemConfigs` overlays `system_configs['allowed_paths']` onto `cfg`. #1111 fixed this exact ordering bug for the tool rate limiter (re-apply after the overlay) but the AllowPaths case was missed, so DB-configured paths never reached `read_file` / `list_files` / `write_file` / `edit` / `send_file`.

Observed in production: setting `system_configs['allowed_paths']` to a path outside the workspace and restarting still produced `read_file: access denied`, and the warn log's `allowedPrefixes` showed only the built-in skill/cli-workspace prefixes — the configured path was absent. `exec` was unaffected (host-native command, no AllowPaths boundary on its working dir target).

## Fix
- Extract the user-allowed-path application from `wireExtraTools` into `applyUserAllowedPaths(toolsReg, allowedPaths)`.
- Re-run it from `runGateway` right after the existing rate-limiter re-apply, using `cfg.Agents.Defaults.AllowedPaths` (now overlaid from the DB).
- The helper is idempotent — `AllowPaths` is additive and the prefix check is membership-based — so the initial wiring call plus the re-apply is safe. Behaviour is unchanged when no allowed paths are configured (empty list → no-op).

Skill/cli-workspace prefixes are unchanged (still applied once during wiring); only the user-configured paths needed re-application because only they are DB-overlaid.

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./cmd/ ./internal/tools/`
- [ ] Web UI builds (no UI changes)
- [x] SQL queries use parameterized `$1, $2` (no SQL changes)
- [ ] New user-facing strings (none — internal startup log only)
- [ ] Migration version bumped (no migration)

## Test Plan
New `cmd/gateway_tools_wiring_test.go`:
- a path outside the workspace is denied before `applyUserAllowedPaths` and readable after;
- an unrelated external path stays denied (grant is scoped);
- repeated application (initial wiring + the gateway re-apply) keeps the grant working and does not error;
- an empty allow list is a no-op.

Also verified end-to-end on a live deployment: after this change a DB-only `system_configs['allowed_paths']` (no other workaround) makes the configured directory readable/writable by `read_file`/`list_files`/`write_file`, with no new `access denied` logs.

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -240,6 +240,17 @@ func runGateway() {
 	} else {
 		toolsReg.SetRateLimiter(nil)
 	}
+
+	// Re-apply user-configured allowed paths for the same reason as the rate
+	// limiter above: setupToolRegistry wired the filesystem tools' AllowPaths
+	// from the JSON5 default before ApplySystemConfigs overlaid
+	// system_configs['allowed_paths'], so DB-driven paths never reached the tools.
+	// Re-run now that cfg reflects the DB value. Safe: server has not started, no
+	// in-flight tool calls.
+	if paths := cfg.Agents.Defaults.AllowedPaths; len(paths) > 0 {
+		applyUserAllowedPaths(toolsReg, paths)
+		slog.Info("filesystem allowed paths reapplied from system_configs", "paths", len(paths))
+	}
 	setupMemoryEmbeddings(pgStores, providerRegistry)
 	usageCapSvc := usagecaps.NewService(pgStores.UsageCaps, pgStores.Providers)
 
--- a/cmd/gateway_tools_wiring.go
+++ b/cmd/gateway_tools_wiring.go
@@ -97,46 +97,30 @@ func wireExtraTools(
 	if pgStores.Skills != nil {
 		skillsAllowPaths = append(skillsAllowPaths, pgStores.Skills.Dirs()...)
 	}
-	// Expand user-configured allowed paths (for cross-drive access on Windows).
-	// These paths are validated per-request in resolvePath for tenant isolation.
-	var userAllowPaths []string
-	for _, p := range agentCfg.AllowedPaths {
-		expanded := config.ExpandHome(p)
-		if expanded != "" {
-			userAllowPaths = append(userAllowPaths, expanded)
-		}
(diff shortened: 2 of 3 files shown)


mrgoonie (approved):
Maintainer review: approved.

Evidence checked:
- Read the full PR metadata and diff: `cmd/gateway.go`, `cmd/gateway_tools_wiring.go`, and new `cmd/gateway_tools_wiring_test.go`.
- Verified the startup ordering issue: `ApplySystemConfigs` overlays DB `system_configs` after initial tool registry setup, and rate limiting is already re-applied after that overlay.
- Verified the fix keeps skill/cli-workspace allow paths in initial wiring and re-applies only user-configured `agents.defaults.allowed_paths` after the DB overlay via `applyUserAllowedPaths`.
- Checked duplicate/prior-work search for `allowed_paths filesystem tools system_configs`; no overlapping issue/PR found besides this PR.
- CI is green (`go`, `web`, `release-versioning`).
- Local spot checks passed: `go test ./cmd` and `go test ./internal/tools`.

Why this is safe:
- The helper is a no-op for empty config.
- Repeated application is harmless because the filesystem tools already treat allowed prefixes as an additive allow-list.
- Scope stays limited to the existing filesystem tools that implement `tools.PathAllowable` (`read_file`, `list_files`, `write_file`, `edit`, `send_file`).
- Added tests cover deny-before-allow, allow-after-grant, unrelated external path denial, repeated apply, and empty allow-list no-op.

No blocking findings.

PR #1275, opened 2026-06-24
fix(webhooks): lease heartbeat to prevent duplicate async processing

thotam:
## Summary
Fixes the async webhook worker double-processing long-running agent calls (e.g. a monthly content-plan run), which created duplicate `content_plans`/schedules via MCP and ended with `context deadline exceeded`. The worker now renews its lease via a heartbeat while the agent runs, so a live run is never reclaimed, while a genuinely dead worker is still recovered within the stale window.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

### Root cause
The worker marked a call `running` then ran the agent with **no heartbeat**. The stale-running sweep treated any `running` row older than `staleRunningWindow` (90s) as a dead worker and reclaimed it → re-ran the agent from scratch → duplicate side-effects. Most visible on heavy/long tasks (>90s).

### Fix — visibility-timeout + heartbeat (lease renewal)
- New `last_heartbeat_at` column (PG migration `000085` + SQLite schema v52).
- `WebhookCallStore.Heartbeat(ctx, callID, lease, now)` — CAS on `lease_token` + `status='running'`; returns `ErrLeaseExpired` if the row was reclaimed.
- `ClaimNext` stamps `last_heartbeat_at` atomically with `started_at`; `ReclaimStale` now keys off `last_heartbeat_at` (or NULL legacy rows) instead of `started_at`.
- Worker runs a heartbeat goroutine every 30s (= window/3) while the agent runs. On lease loss it cancels the run context so the agent stops immediately and writes no further MCP side-effects. Each heartbeat DB call is bounded to 5s.
- `invokeAgent` now honors the (shutdown-detached) run context so heartbeat-triggered cancellation actually propagates.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes for changed packages — note: a pre-existing, unrelated failure (`internal/tools/document_parser_test.go: undefined waitForRecordedPIDs`) already exists on `dev`; not introduced here
- [ ] `go test -race ./...` — could not run `-race` locally (no gcc/CGO on dev machine); ran the suites without `-race` (all green). CI runs `-race` on Linux. Stub counters use `sync/atomic`
- [x] SQL queries use parameterized `$1, $2` / `?` (no string concat)
- [x] New user-facing strings — N/A (no user-facing strings; only `slog`/godoc)
- [x] Migration version bumped — `RequiredSchemaVersion` 84→85; SQLite `SchemaVersion` 51→52

## Test Plan
- **Store tests** (`sqlitestore/webhooks_test.go`): `Heartbeat` renews on valid lease and returns `ErrLeaseExpired` on a wrong/reclaimed lease; `ReclaimStale` does **not** reclaim a row with a fresh heartbeat (even if `started_at` is old).
- **Worker tests** (`webhooks/worker_test.go`): heartbeat loop renews repeatedly without cancelling the run; on `ErrLeaseExpired` it cancels the run context within one interval.
- Verified `go build ./...`, `go build -tags sqliteonly ./...`, and `go test ./internal/webhooks/... ./internal/http/... -run Webhook` + `go test -tags sqlite ./internal/store/sqlitestore/... -run TestWebhook` all pass.
- Smoke-tested on a beta build (`v3.15.0-beta.totaa.5`) — confirmed working.

### Surface parity
Server-only. API contract / Web UI / CLI: N/A — internal async webhook worker behavior, no request/response shape change and `last_heartbeat_at` is not serialized.

### DIFF
--- a/internal/http/webhooks_admin_calls_test.go
+++ b/internal/http/webhooks_admin_calls_test.go
@@ -90,6 +90,7 @@ func (s *adminCallStore) DeleteOlderThan(context.Context, uuid.UUID, time.Time) 
 	return 0, nil
 }
 func (s *adminCallStore) ReclaimStale(context.Context, time.Time) (int64, error) { return 0, nil }
+func (s *adminCallStore) Heartbeat(context.Context, uuid.UUID, string, time.Time) error { return nil }
 
 // ---- stub testers ----
 
--- a/internal/http/webhooks_auth_test.go
+++ b/internal/http/webhooks_auth_test.go
@@ -138,6 +138,9 @@ func (s *stubWebhookCallStore) DeleteOlderThan(_ context.Context, _ uuid.UUID, _
 func (s *stubWebhookCallStore) ReclaimStale(_ context.Context, _ time.Time) (int64, error) {
 	return 0, nil
 }
+func (s *stubWebhookCallStore) Heartbeat(_ context.Context, _ uuid.UUID, _ string, _ time.Time) error {
+	return nil
+}
 
 // ---- helpers ----
 
--- a/internal/http/webhooks_llm_test.go
+++ b/internal/http/webhooks_llm_test.go
@@ -82,6 +82,9 @@ func (s *llmCallStore) DeleteOlderThan(_ context.Context, _ uuid.UUID, _ time.Ti
 func (s *llmCallStore) ReclaimStale(_ context.Context, _ time.Time) (int64, error) {
 	return 0, nil
 }
+func (s *llmCallStore) Heartbeat(_ context.Context, _ uuid.UUID, _ string, _ time.Time) error {
+	return nil
+}
 
 // ---- helpers ----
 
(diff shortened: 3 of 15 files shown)


mrgoonie (approved):
Reviewed for duplicate/prior work, project standards, strategic value, correctness, security, migrations, and tests.nnThis fixes a real async webhook lease/stale-reclaim bug: long-running agent calls could exceed the 90s stale window and be reclaimed while still alive, causing duplicate side effects. The PR adds a dedicated heartbeat timestamp, CAS heartbeat renewal, stale reclaim based on heartbeat freshness, and cancellation on lease loss.nnChecks I verified:n- No duplicate open/merged PR or issue found for the webhook heartbeat/lease double-processing fix.n- CI is green: go, web, release-versioning.n- Migration/version bumps are aligned: PG 000085, SQLite schema 51→52, RequiredSchemaVersion 84→85.n- Store and worker coverage exercises valid/wrong lease heartbeat, fresh-heartbeat reclaim protection, repeated renewal, and cancellation on ErrLeaseExpired.n- The change is scoped to internal async webhook processing; no API/UI contract change.nnOne non-blocking follow-up: WebhookCallData still does not expose LastHeartbeatAt, so list/admin call responses will not show the new heartbeat field. That is fine for this fix because worker logic uses direct SQL updates, but we may want to surface it later for debugging stuck async deliveries.nnApproved.

PR #1276, opened 2026-06-24
feat(discord): follow threads after first mention

Hitesh-Sisara:
## Summary

Once the bot is @mentioned in a Discord thread, it responds to all subsequent messages in that thread without requiring further mentions.

## Motivation

Current behavior requires @mentioning the bot on every single message in groups/threads. This is tedious for ongoing conversations. Users expect to mention once to activate, then the bot follows the entire thread (like Hermes and other Discord bots).

## Changes

- `internal/channels/discord/discord.go`: Added `mentionedThreads` map + mutex to Channel struct
- `internal/channels/discord/handler.go`: Modified mention gating logic to check thread membership

## How it works

1. Bot is @mentioned in a thread → thread channel ID stored in `mentionedThreads`
2. Future messages in that thread bypass mention requirement → bot responds automatically
3. Map capped at 5000 entries (evicts oldest half when full) to prevent unbounded growth
4. Only applies to actual threads (`isThreadChannel` check), not regular channels
5. Non-thread channel messages still require @mention (preserves existing behavior)

## Testing

- Tested on a live Discord server with thread creation + mention + follow-up messages
- Bot correctly responds to all messages after initial mention in thread
- Bot stays silent in threads where it was never mentioned
- Regular channel messages still require @mention

### DIFF
--- a/internal/channels/discord/discord.go
+++ b/internal/channels/discord/discord.go
@@ -34,6 +34,11 @@ type Channel struct {
 	audioMgr        *audio.Manager              // unified STT via audio.Manager (nil = no STT)
 	// pairingService, pairingDebounce, approvedGroups, groupHistory, historyLimit, requireMention
 	// are inherited from channels.BaseChannel.
+
+	// mentionedThreads tracks thread/channel IDs where the bot was @mentioned.
+	// Once mentioned in a thread, respond to ALL subsequent messages without requiring another mention.
+	mentionedThreads   map[string]bool
+	mentionedThreadsMu sync.Mutex
 }
 
 // New creates a new Discord channel from config.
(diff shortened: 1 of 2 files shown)


mrgoonie (changes requested):
Thanks for the contribution. I reviewed this against current dev and the feature direction is reasonable, but this implementation is not safe to merge yet.

Summary: the PR adds in-memory tracking so a Discord thread keeps responding after the first bot mention. The changed files are small (`internal/channels/discord/discord.go`, `handler.go`), and I found no duplicate PR/issue for this exact feature.

Mandatory gates:
- Duplicate / prior implementation: clear. Search only found this PR plus unrelated channel work; git history only found prior thread backfill work.
- Project standards: docs found (`CLAUDE.md`, `AGENTS.md`). This is gateway/channel-only; UI/API/CLI parity is N/A.
- Strategic necessity: clear UX value for Discord thread conversations.

Important blocker:
- The remembered-thread bypass is checked after `checkGroupPolicy(ctx, senderID, channelID, mentioned)`, but `checkGroupPolicy` still receives `mentioned=false` for follow-up thread messages. For `PolicyNeedsPairing`, it immediately returns false when `RequireMention()` is true and `mentioned` is false. That means this feature only works for already-allowed groups and still cannot activate pairing/setup replies or policy flow for remembered threads. Worse, the new bypass and the policy gate now disagree on whether the bot was addressed. The thread-follow decision needs to be computed before policy evaluation and passed consistently into policy gating, or explicitly scoped so remembered threads are only honored after the group is already allowed.

Suggested follow-ups:
- Add regression coverage for: first mention in thread records the thread; second unmentioned message in the same thread reaches the bus when group policy allows it; unrelated regular channel still requires mention; and `PolicyNeedsPairing` behavior is intentionally defined.
- Consider avoiding repeated Discord API lookups by deriving/cache-checking thread status once per message instead of calling `isThreadChannel` twice.

Verdict: Request changes.

PR #1175, opened 2026-05-26
fix(tts): unify voice resolution — dashboard /tts settings now affect the tts tool

vanducng:
## Real-world bug

A user configured Vietnamese voice "HoaiMy" (\`vi-VN-HoaiMyNeural\`) on the dashboard \`/tts\` page. The Test Playground worked. But when the LLM invoked the \`tts\` tool, Edge synthesized English instead.

## Root cause — two TTS storage locations

| Location | Read by | Written by |
|----------|---------|------------|
| \`system_configs[tts.<provider>.voice]\` | Test Playground, auto-apply at construction-time | Dashboard \`/tts\` page |
| \`builtin_tool_tenant_configs[tts].default_voice_id\` | \`tts\` tool (LLM-invoked) | (not exposed in any UI) |

When the LLM calls \`tts\` without an explicit \`voice\` arg, the resolution chain (\`args > agent OtherConfig > tenant builtin\`) finds no override → empty voice → Edge falls back to its built-in default (English).

## Fix

Add \`system_configs\` as a 4th-level fallback so the dashboard is the single source of truth:

\`\`\`
args > agent OtherConfig > tenant builtin > system_configs[tts.<provider>.voice]
\`\`\`

- New \`TtsTool.SetSystemConfigStore(s)\` setter
- \`resolveVoiceAndModel\` now takes \`providerName\` and reads the right key
- Effective provider resolved BEFORE voice/model so the right lookup key is used
- Wired at gateway boot in \`cmd/gateway.go\` after \`pgStores\` is ready

## Tests

4 new unit tests in \`internal/tools/tts_systemconfigs_fallback_test.go\`:

- \`TestResolveVoiceAndModel_SystemConfigsFallback\` — voice + model resolve from system_configs
- \`TestResolveVoiceAndModel_ArgWinsOverSystemConfigs\` — explicit arg precedence preserved
- \`TestResolveVoiceAndModel_NoStoreNoFallback\` — graceful when store unwired
- \`TestResolveVoiceAndModel_EmptyProviderSkipsFallback\` — no key lookup without provider

All existing \`tts\` tests still pass.

## Verification

- \`go build ./...\` (PG) clean
- \`go vet ./...\` clean
- \`go test ./internal/tools/...\` green
- Already running on the fork's prod cluster (v3.23.43) — verified TTS tool now uses dashboard-configured voice.

Cherry-picked from fork's PR dataplanelabs/goclaw#172.

PR #1277, opened 2026-06-24
fix(tts): route agent tool through tenant provider

nguyenha935:
## Summary

- Route the agent built-in `tts` tool through the tenant TTS provider resolver before falling back to globally registered providers.
- Allow explicit `provider=gemini` when Gemini is configured from dashboard / tenant settings but is not present in the boot-time global provider map.
- Keep explicit provider requests strict: if the requested tenant provider does not match, the tool returns `tts provider not found` instead of silently using another provider.
- Update the tool parameter description to include `gemini`.

## Target Branch

`dev`

## Relationship to Existing Work

This is complementary to #1175, not a duplicate.

#1175 improves voice/model fallback from stored settings. This PR fixes the separate routing issue where the agent built-in `tts` tool only looked at global manager providers and could miss a DB-configured tenant provider such as Gemini, causing agent TTS to fall back to Edge even though `/tts` and `/v1/tts/synthesize` use Gemini correctly.

## Surface Parity

- Dashboard / Config TTS: no schema or UI contract change.
- HTTP `/v1/tts/synthesize`: no behavior change; it already uses tenant provider resolution.
- Agent built-in `tts` tool: now matches tenant provider routing used by HTTP and auto-TTS paths.
- Runtime / Docker / migrations: no changes.

## Test Plan

Passed:

```bash
docker run --rm -v /root/dev/goclaw-tts-tool-routing:/src -w /src golang:1.26-bookworm \
  go test ./internal/tools -run 'TestTtsTool_.*TenantProvider|TestTtsTool_ExplicitProviderMismatchStillErrors|TestTtsTool_GlobalFallbackStillWorksWithoutTenantProvider|TestResolvePrimary|TestTtsTool_TTSParams'
```

Result:

```text
ok  	github.com/nextlevelbuilder/goclaw/internal/tools	0.025s
```

Also checked:

```bash
git diff --check
```

Result: clean.

Broader package test was attempted earlier:

```bash
go test ./internal/tools ./internal/audio ./internal/http
```

Known environment-specific failures in the Docker test environment:

- `internal/tools`: process-group timeout/reaping tests fail inside the container.
- `internal/http`: Ollama tests fail because `host.docker.internal` is not resolvable in this Linux Docker environment.
- `internal/audio`: passed.

### DIFF
--- a/internal/tools/tts.go
+++ b/internal/tools/tts.go
@@ -276,6 +276,10 @@ func (t *TtsTool) Execute(ctx context.Context, args map[string]any) *Result {
 				tenantOpts.Params = mergeParams(opts.Params, adapted)
 			}
 			result, err = p.Synthesize(ctx, text, tenantOpts)
+			if err != nil {
+				slog.Warn("tts tenant provider failed, trying fallback", "provider", tenantName, "error", err)
+				result, err = mgr.SynthesizeWithFallbackAdapted(ctx, text, opts, genericAgentParams)
+			}
 		} else {
 			// Resolve primary from tenant settings or default.
 			primary := t.resolvePrimary(ctx, mgr)
--- a/internal/tools/tts.go
+++ b/internal/tools/tts.go
@@ -70,7 +70,7 @@ func (t *TtsTool) Parameters() map[string]any {
 			},
 			"provider": map[string]any{
 				"type":        "string",
-				"description": "TTS provider: openai, elevenlabs, edge, minimax. Optional — uses primary if omitted.",
+				"description": "TTS provider: openai, elevenlabs, edge, minimax, gemini. Optional — uses configured primary if omitted.",
 			},
 		},
 		"required": []string{"text"},
@@ -88,11 +88,11 @@ type ttsOverride struct {
 // agentAudioConfig is the JSON shape read from AgentAudioSnapshot.OtherConfig
 // for per-agent TTS tuning. Keys match the agents.other_config column.
 type agentAudioConfig struct {
-	TTSVoiceID string         `json:"tts_voice_id,omitempty"`
-	TTSModelID string         `json:"tts_model_id,omitempty"`
+	TTSVoiceID string `json:"tts_voice_id,omitempty"`
+	TTSModelID string `json:"tts_model_id,omitempty"`
 	// TTSParams carries the per-agent generic TTS override keys (speed, emotion, style).
 	// Stored as generic keys; AdaptAgentParams converts to provider-specific keys per attempt.
-	TTSParams  map[string]any `json:"tts_params,omitempty"`
+	TTSParams map[string]any `json:"tts_params,omitempty"`
 }
 
 // resolveVoiceAndModel computes the effective voice + model IDs for the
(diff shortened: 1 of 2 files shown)


mrgoonie:
Maintainer triage note: I reviewed this PR metadata/diff and the change is narrowly scoped to the agent built-in TTS tool routing path. The new tests cover omitted-provider tenant routing, explicit tenant provider not registered globally, explicit mismatch strictness, and global fallback.nnI am not approving/merging yet because CI is currently red on the `go` job. The failure is in `internal/agent` race-test logging capture (`TestMaybeSummarize_LogsTriggerDecisionOverThreshold`), while `internal/tools` passed; that looks unrelated to this TTS diff, but the PR still needs a green required check or maintainer override before merge.nnNext action: rerun/fix the failing go check. If it goes green and no new findings appear, this looks mergeable from the TTS-tool-routing review.

mrgoonie (changes requested):
**Summary**: This PR correctly targets the agent `tts` tool routing gap: tenant-resolved providers such as Gemini can now be used even when they are not present in the boot-time global provider map.

**Risk level**: Medium — small diff and good focused tests, but it changes provider selection/fallback behavior on an agent-facing tool.

**Mandatory gates**:
- Duplicate / prior implementation: clear. I checked related TTS routing work; #1175 is related voice/model fallback work, not this provider-routing fix.
- Project standards: docs/current code pattern followed. The diff stays inside `internal/tools` and uses the existing `tts.Manager.ResolveTenantProvider` path used by other TTS surfaces.
- Strategic necessity: clear value. This restores parity between dashboard/HTTP TTS and the agent built-in `tts` tool, especially for tenant-configured Gemini.

**Findings**:
- **Important — preserve global fallback when a tenant provider fails.** In the no-explicit-provider path, the old code tried the primary provider and then called `mgr.SynthesizeWithFallbackAdapted(...)` on error. The new tenant-provider branch calls the tenant provider directly and does not fall back when that tenant provider returns an error. That is a behavior regression for the agent `tts` tool: a configured tenant provider outage now hard-fails instead of trying the manager fallback chain. This matters because the PR’s stated goal is surface parity and routing correctness, not reducing resilience. Please add fallback on tenant provider error, with provider-specific adaptation preserved for the tenant attempt and fallback attempts.
  - Evidence: `internal/tools/tts.go`, new no-provider branch calls `p.Synthesize(ctx, text, tenantOpts)` and then exits the branch; the fallback call only exists under the global-primary branch.
  - Suggested fix: after tenant `Synthesize` fails, log the tenant provider failure and call `mgr.SynthesizeWithFallbackAdapted(ctx, text, opts, genericAgentParams)` unless the design intentionally treats tenant provider selection as strict. If strict is intended, document it and update tests.

**CI**:
- `web` and `release-versioning` passed.
- `go` failed in `internal/agent` race-detection test (`TestMaybeSummarize_LogsTriggerDecisionOverThreshold`) unrelated to this PR’s touched files; `internal/tools` passed in the CI output.

**Verdict**: Request changes — the tenant routing fix is good, but the no-provider tenant branch should not silently drop fallback resilience without an explicit product decision and test coverage.

*Posted by /github-maintain at 2026-06-24T00:00:00Z*

nguyenha935:
Updated the PR to preserve fallback resilience for the omitted-provider tenant TTS path. The tenant provider is tried first; if it fails, the tool now logs the tenant failure and calls SynthesizeWithFallbackAdapted with the original generic agent params so fallback attempts keep provider-specific adaptation. Added TestTtsTool_TenantProviderFailureFallsBackWhenProviderOmitted. Current CI is green: go, web, and release-versioning all pass.

mrgoonie (approved):
Re-reviewed after the fallback update. The previous blocker is addressed: the omitted-provider tenant path now logs tenant provider failure and calls SynthesizeWithFallbackAdapted with the original generic agent params, so fallback attempts keep provider-specific adaptation. Focused regression coverage was added in TestTtsTool_TenantProviderFailureFallsBackWhenProviderOmitted, and current CI is green. Approving.

PR #1279, opened 2026-06-25
feat(cron): deterministic command payloads (run a shell command, no LLM)

zezaeoh:
## Summary
Cron jobs always run an agent turn today, so deterministic work (health probes, backups, syncs) pays model tokens on every fire. This adds a **`command`** cron payload kind that runs a shell command directly in the gateway process with **zero model tokens**, mirroring openclaw's command cron. It is opt-in per gateway via `cron.command_enabled` (default **false**).

## Type
- [x] Feature

## Target Branch
`dev` ✅

## What's included
- **store** — `CronPayload.Command` (`*CronCommandSpec`: `argv` / `cwd` / `env` / `input` / `timeoutSeconds` / `noOutputTimeoutSeconds` / `outputMaxBytes`). Persisted in the **existing `payload` JSON blob**, so there is **no migration** and no schema version bump. (The previously-unused `Command string` field is repurposed into the structured spec; it was never read anywhere.)
- **`internal/cronexec`** — a focused, in-process command runner: wall-clock timeout, no-output watchdog, per-stream output capping, and process-group termination (so a timed-out command's forked children are killed too). Unix/Windows split for the process-group helpers.
- **handler** (`cmd/gateway_cron.go`) — command jobs run in-process and deliver stdout on success (honoring the `NO_REPLY` sentinel). A non-zero exit / timeout returns an error so the run is recorded as `error` and retried per `cron.max_retries`; **failures are not delivered**, mirroring the agent path (only successful output is announced — no channel spam). Token usage is recorded as `0`/`0`. Existing delivery logic is extracted into a shared `deliverCronOutput` helper.
- **surfaces** — `cron.create` RPC, the agent `cron` tool, and a new `goclaw cron create` CLI all accept command payloads.
- **security** — gated by `cron.command_enabled`. Commands run with the gateway process's privileges, so the feature is opt-in per gateway; when disabled the RPC and the agent tool reject command payloads, and the handler refuses to run them (defense in depth).
- **i18n** — `error.command_cron_disabled` added to all three locales (en/vi/zh).
- **docs** — new "Command Payloads" section in `docs/08-scheduling-cron.md`.

## Example
```bash
# operator CLI (requires cron.command_enabled=true)
goclaw cron create --name disk-probe --cron '*/15 * * * *' \
  --command 'df -h /' --deliver --channel telegram --to '-100123'

goclaw cron create --name nightly-backup --at 2026-07-01T18:00:00Z \
  --argv '["/opt/backup.sh","--full"]' --timeout 5m
```
```jsonc
// config
{ "cron": { "command_enabled": true, "command_timeout": "5m" } }
```

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes (only pre-existing, unrelated gofmt drift in `keys.go`/`gateway.go` OAuth blocks left untouched to keep the diff clean)
- [x] Tests pass: `go test -race ./...` for `internal/cronexec`, `cmd`, `internal/store`, `internal/tools`, `internal/gateway/methods`, `internal/store/pg`, `internal/config`, `internal/i18n` (full repo suite needs DB services)
- [ ] Web UI builds — N/A (no UI changes)
- [x] SQL queries use parameterized `$1, $2` — no new SQL; the command spec persists via the existing `payload` JSON-blob write path
- [x] Migration version bumped — **N/A**, no migration (payload stored in the existing `JSONB` blob)

## Test Plan
- `internal/cronexec/runner_test.go` — stdout capture, non-zero exit → error (stderr surfaced), wall-clock timeout, no-output timeout (and not tripping while output flows), stdin, env, output cap, empty argv, stdout-preferred-over-stderr summary. Race-clean.
- `cmd/gateway_cron_command_test.go` — handler refuses command jobs when `command_enabled=false`; success path returns content with zero tokens and delivers to the channel; non-zero exit returns an error and does **not** deliver; empty argv rejected.
- Manual: `goclaw cron create` against a local gateway with `cron.command_enabled=true`.

## Notes / open questions for reviewers
- Command failures retry via the shared cron retry path (`cron.max_retries`). For side-effecting commands operators can set `max_retries: 0`. Happy to special-case command jobs to single-shot if you'd prefer.
- The agent `cron` tool can create command jobs when the gateway enables the feature (parity with the RPC/CLI). If you'd rather restrict command-cron creation to operators only, that's a small change to gate the tool path off.

### DIFF
--- a/cmd/cron_cmd.go
+++ b/cmd/cron_cmd.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"text/tabwriter"
 	"time"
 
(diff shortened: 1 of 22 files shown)


mrgoonie (changes requested):
Thanks for the detailed implementation and test coverage. I found one merge-blocking security/contract gap before this can land.

**Important: command cron can be enabled via update even when `cron.command_enabled=false`.**

`cron.create` and the agent tool add path correctly gate command payloads, and the runtime refuses to execute them when disabled. But both update paths accept `CronJobPatch.Command` without the same gate/validation:

- `internal/gateway/methods/cron.go` `handleUpdate` unmarshals `patch.command` and calls `m.service.UpdateJob(...)` directly.
- `internal/tools/cron.go` `handleUpdate` unmarshals the patch and calls `t.cronStore.UpdateJob(...)` directly.
- `internal/store/pg/cron_update.go` and `internal/store/sqlitestore/cron_crud.go` switch the payload to `CronPayloadKindCommand` whenever `patch.Command != nil`.

So a caller can create a normal message cron while the gateway has command cron disabled, then update it with a command payload. The handler will refuse execution later, but the disabled-gateway contract is that command payloads cannot be created/mutated into existence through RPC/tool surfaces. This also lets invalid command specs be persisted via update (for example empty `argv`) because update bypasses `ValidateCronCommandSpec`.

Please add the same checks on both update surfaces before `UpdateJob`: if `patch.Command != nil`, require the relevant command-enabled flag and run `store.ValidateCronCommandSpec`. Add regression coverage for RPC update and agent tool update with command disabled + invalid argv.

After that, this looks mergeable: CI is green, the feature is opt-in by default, and duplicate/prior-work search did not show an existing implementation.

zezaeoh:
Thanks @mrgoonie — that's a real gap, you're right. I gated create but left both update surfaces open, so a job could be created as a normal message cron and then mutated into a command payload via update on a gateway with `cron.command_enabled=false`, and invalid specs (e.g. empty `argv`) could slip in because update bypassed `ValidateCronCommandSpec`.

Fixed in `3b91046c`:

- **`internal/gateway/methods/cron.go` `handleUpdate`** — if `params.Patch.Command != nil`, require `cron.command_enabled` and run `store.ValidateCronCommandSpec` before `UpdateJob`, matching `handleCreate`.
- **`internal/tools/cron.go` `handleUpdate`** — parses the command via the same `parseCronCommandSpec` path as `add` and removes the raw `command`/`commandArgv` keys before the generic patch unmarshal (so a shell-string `command` can't break the unmarshal, and no command payload can reach `UpdateJob` without passing the `commandEnabled` + validate gate).
- Left the stores as pure persisters per your "checks on both update surfaces" guidance.

Regression coverage added:
- RPC update: command-disabled → blocked, empty `argv` → rejected, enabled + valid → updates.
- Agent tool update: command-disabled → blocked, invalid `argv` → rejected.

`go build ./...` / `go build -tags sqliteonly ./...` / `go vet` / `go test -race` (tools + methods) all green. Ready for another look when you have a moment.

mrgoonie (approved):
Summary: Re-reviewed latest head 3b91046c after the update-surface fix. The previous blocker is resolved: RPC cron.update now gates patch.command on cron.command_enabled and validates the command spec before UpdateJob, and the agent cron tool update path parses command/commandArgv through the same command-enabled + ValidateCronCommandSpec checks as add before persisting. Regression tests now cover disabled update and invalid argv for both RPC and agent tool paths.

Verification run locally against this PR head:
- go test ./internal/gateway/methods ./internal/tools ./internal/cronexec ./cmd

CI note: web and release-versioning are green; the go check was still pending when reviewed, so merge should wait for that required check to finish green.

Verdict: Approved once the pending go CI check is green.

PR #1293, opened 2026-06-28
feat(bitrix24): inject <media:*> tags so the LLM sees attachments

tech-synity:
## Problem

The Bitrix24 channel handler downloads attachments (PDF, audio, image) and forwards them as `bus.MediaFile` to the agent loop, but it never prepends `<media:*>` tags into the user message body the LLM sees. The agent loop's `enrichInputMedia` REPLACES tags it finds — it does NOT insert new ones — so a Bitrix-borne attachment arrives at the LLM as bare text.

Observed in production:
- User sends `HDDT.pdf` + "đọc giúp" → agent calls `web_fetch` against the Bitrix CDN URL (which 401s for non-intranet callers), or — worse — hallucinates a `crm.documentgenerator.document.getPdf` call against a document id parsed out of the filename.
- User sends an mp3 → agent replies "no file attached" (`tool_calls_count=0, iterations=0`).

Telegram / Slack / Discord already prepend `<media:*>` tags via the shared `media.BuildMediaTags` helper. Bitrix24 was the lone holdout.

## Fix

- `internal/channels/bitrix24/media_tags.go` *(new)*
  - `mediaFilesToInfos(files []bus.MediaFile) []media.MediaInfo` adapter.
  - `classifyMediaType(mime string) string` MIME-prefix → `media.Type*` mapping. Unknown MIMEs fall back to `TypeDocument` (safest: read_document handles generic binaries via local parser + vision chain). Bitrix ships voice notes as `audio/*` in the MIME — classified as `TypeAudio` since `read_audio` handles both audio and voice identically.
- `internal/channels/bitrix24/handle.go`
  - After `downloadEventFiles` and before `HandleMessageMedia`, prepend `media.BuildMediaTags(mediaFilesToInfos(mediaFiles))` to `text`. No-op when there are no media files.
- `internal/channels/bitrix24/media_tags_test.go` *(new)*
  - Pins the MIME map, the conversion, and the end-to-end tag shape through `media.BuildMediaTags`.

## Verification

```
go build ./...                    # PG
go build -tags sqliteonly ./...   # Desktop
go test ./internal/channels/bitrix24/...   → ok 2.952s
```

Live-tested against `tamgiac.bitrix24.com` after deploying a local image build:
- PDF upload (323KB) + "đọc giúp anh" → agent calls `read_document` with the resolved workspace path, file loads, gemini-native completes the analysis. (Audio path exercised similarly via `read_audio`.)

## Surface parity

Backend-only. No API contract / web UI / CLI change. Tag format already matches Telegram / Slack / Discord output via the shared `media.BuildMediaTags` helper, so downstream consumers (enrichInputMedia, read_*) need no change.

[B24:2794]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,14 @@ All notable changes to GoClaw are documented here. For full documentation, see [
   pre-write discovery via `vault_search`, `memory_search`, and
   `knowledge_graph_search` to surface related files before writing and
   avoid duplicates; documents Vault scope mirroring and id-routing rules.
+- **Bitrix24 channel 2-way media (file) transfer** — Inbound media downloads via
+  `imbot.v2.File.download` (one-time authenticated URL) with MIME preservation for
+  images, PDFs, audio, and video. Outbound uploads via `imbot.v2.File.upload` (base64).
+  Shared `media_max_mb` config knob (default 20 MB) caps both directions. Requires
+  `imbot` OAuth scope (no `disk` scope needed). Inbound handled by new
+  `internal/channels/bitrix24/download.go`; outbound by `send_media.go`. New
+  `BaseChannel.HandleMessageMedia()` method centralizes media-aware message handling.
+  See `docs/05-channels-messaging.md` § 16 (Bitrix24) for configuration.
 
 - **Skill agent manage grants** — Adds per-agent skill edit/delete grants with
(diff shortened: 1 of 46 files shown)


tech-synity:
Replacing with a single-commit PR — this branch carries unrelated [B24:2794] history because the base ref didn't have our internal Synity PRs yet.

PR #1300, opened 2026-06-28
fix: filter streamed thinking tags

nguyenha935:
## Summary
- keep parsing streamed <think>/<thinking>/<thought>/<antThinking> tags after visible answer chunks have started
- hold partial tag prefixes across chunks so split tags do not leak into user-facing chat text
- route extracted tagged thinking into reasoning bubbles/metadata and mirror filtering in the web chat stream adapter

## Relation to #415
This covers the streaming/channel path that #415 does not handle. #415 promotes already-complete thinking tags from final assistant text; this PR handles realtime chunked tags, late tags, and split tag prefixes before they can render in chat.

## Surface parity
- Gateway/channel: updated stream handling and reasoning bubble extraction
- Pipeline: promotes tagged thinking from observe/final content into thinking metadata
- Web UI: filters streamed tagged thinking before updating visible chat content
- CLI/runtime package: N/A, no CLI contract changes

## Tests
- docker run --rm -v /root/.config/superpowers/worktrees/Goclaw/pr-streaming-think-tag-filter:/src -w /src golang:1.26-bookworm go test ./internal/channels ./internal/pipeline -run "Think|Thinking|Reasoning|Observe" -count=1
- docker run --rm -v /root/.config/superpowers/worktrees/Goclaw/pr-streaming-think-tag-filter:/src -w /src golang:1.26-bookworm go test ./internal/channels ./internal/pipeline -count=1
- cd ui/web && pnpm test think-tag-stream chat-message.adapter --run
- cd ui/web && pnpm build

### DIFF
--- a/internal/channels/events.go
+++ b/internal/channels/events.go
@@ -98,10 +98,10 @@ func (m *Manager) HandleAgentEvent(eventType, runID string, payload any) {
 			currentStream := rc.stream
 			rc.stream = nil
 			rc.inToolPhase = true
-			rc.thinkingDone = false    // allow new thinking in next iteration
-			rc.thinkingBuffer = ""     // reset thinking buffer for new iteration
-			rc.hasThinking = false     // new iteration starts fresh
-			rc.tagParseSkipped = false // re-enable tag parsing for next iteration
+			rc.thinkingDone = false // allow new thinking in next iteration
+			rc.thinkingBuffer = ""  // reset thinking buffer for new iteration
+			rc.hasThinking = false  // new iteration starts fresh
+			rc.tagParsePending = "" // discard any incomplete tag from the previous iteration
 			rc.mu.Unlock()
 			if currentStream != nil {
 				if err := currentStream.Stop(ctx); err != nil {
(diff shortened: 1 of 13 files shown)


clark-cant (approved):
## Review: Approved ✅

Solid fix — addresses a real UX issue where raw `<think>` tags leak into the visible answer stream for providers that embed thinking in content (DeepSeek-via-OpenRouter, Qwen, Ollama).

**Backend (Go):**
- Replacing `tagParseSkipped` boolean with `tagParsePending` string is the right call. The old approach would skip re-parsing after the first chunk had no tags, missing late-arriving think tags or tags split across stream chunks.
- `splitTrailingThinkTagPrefix` + `isPossibleThinkOpenPrefix` properly handle partial tag names at chunk boundaries (e.g., `<think` at end of one chunk, `ing>...` in the next).
- Pipeline-level `splitTaggedThinkingContent` in observe_stage ensures final responses also get cleaned.

**Frontend (TypeScript):**
- `think-tag-stream.ts` mirrors the backend logic for the web UI streaming path. Clean state machine with proper pending buffer handling.
- History adapter applies the same filter to loaded messages — good for backward compat with sessions that already have tagged content in DB.

**Tests:** Good coverage across all layers — split tags, late tags, streaming filter, history transform. Tests verify the exact failure mode (partial tag name leaking before completion).

**CI green, mergeable. Ready to merge.**

PR #1305, opened 2026-06-29
fix(mcp): Update button silently doing nothing due to hidden Zod vali…

bclermont:
fix(mcp): Update button silently doing nothing when editing MCP servers

  Problem

  Clicking Update after adding headers (e.g. Authorization / Bearer token) to an MCP server did nothing — no spinner, no error, no dialog close. Clicking Test worked fine.

  Root cause: The Update button calls rhfHandleSubmit(callback) from react-hook-form. Before invoking the callback, RHF runs full Zod validation on all fields. If any field fails, RHF calls
  the (implicit) no-op error handler — completely silently. The Test button bypasses RHF entirely via a direct handleTest() call, which is why it worked.

  Three silent failure paths were identified:

  1. timeout field — the input's onChange called setValue("timeout", Number(e.target.value)). Number("") === 0, which fails z.number().min(1). If the user cleared the field even briefly, the
  form was permanently invalid until the dialog was closed and reopened.
  2. headers / env fields — setValue("headers", v) was called without { shouldDirty: true, shouldValidate: true }, meaning RHF did not re-validate after headers changed. On certain form
  states this left the form in a stale invalid state.
  3. No error surface — formState.errors from RHF were never displayed to the user. Only the manual error state (set inside the callback) was shown, but the callback was never reached.

#  Changes

 - mcp-form-dialog.tsx: Display RHF formState.errors inline so Zod failures are visible; add type="button" to Update/Create button
 - mcp-connection-fields.tsx: Add { shouldDirty: true, shouldValidate: true } to setValue("headers", v)
 - mcp-settings-fields.tsx: Same for setValue("env", v); guard timeout with parseInt(..., 10) || 60 to prevent 0/NaN entering RHF
 - mcp.schema.ts: No schema change (kept z.number().min(1); input-layer guard is the right fix

### DIFF
--- a/ui/web/src/pages/mcp/mcp-connection-fields.tsx
+++ b/ui/web/src/pages/mcp/mcp-connection-fields.tsx
@@ -90,7 +90,7 @@ export function McpConnectionFields({ form }: McpConnectionFieldsProps) {
             <Label>{t("form.headers")}</Label>
             <KeyValueEditor
               value={headers}
-              onChange={(v) => setValue("headers", v)}
+              onChange={(v) => setValue("headers", v, { shouldDirty: true, shouldValidate: true })}
               keyPlaceholder={t("form.headerKeyPlaceholder")}
               valuePlaceholder={t("form.headerValuePlaceholder")}
               addLabel={t("form.addHeader")}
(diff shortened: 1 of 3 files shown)


clark-cant (approved):
Reviewed and approved. Clean fix for the silent Zod validation failure in MCP form dialog.

**Summary**: Fixes three silent failure paths where RHF/Zod validation errors were swallowed, causing the Update button to do nothing.

**Changes look good**:
- `shouldDirty: true, shouldValidate: true` on headers/env setValue calls — correct
- `parseInt(..., 10) || 60` guard for timeout — prevents 0/NaN entering RHF
- Inline formState.errors display — good UX improvement
- `type="button"` on submit button — prevents accidental double-submit

One minor note: the `eslint-disable` for `@typescript-eslint/no-explicit-any` on the resolver line — consider typing the resolver explicitly in a follow-up, but not blocking.

CI green, merge state clean. Merging now.

PR #1306, opened 2026-06-29
fix(telegram): honor per-group/topic overrides for DB instances

aaron-tsar:
## Summary
DB-instance Telegram channels (created via the dashboard) ignored per-group/per-topic overrides, so the bot replied in every forum topic regardless of the configured per-topic gating. This maps the `groups` payload through the factory so per-topic config takes effect.

## Type
- [x] Bug fix

## Target Branch
`dev`

## Root Cause
The dashboard (`channel-instance-form-step.tsx`) persists per-group/per-topic overrides (`enabled`, `require_mention`, `skills`, `tools`, `system_prompt`, ...) into the channel instance config JSONB. But `telegramInstanceConfig` in `internal/channels/telegram/factory.go` had no `groups` field, so the payload was silently dropped on unmarshal. `config.TelegramConfig.Groups` stayed `nil`, `resolveTopicConfig` hit its `cfg.Groups == nil` early return and fell back to global defaults (`isEnabled() == true`), so the bot responded in all topics.

The static file-config path (`telegram.New(cfg.Channels.Telegram, ...)`) was unaffected because it passes a fully-populated `config.TelegramConfig`.

## Fix
- Add `Groups map[string]*config.TelegramGroupConfig` to `telegramInstanceConfig` (mirrors `config.TelegramConfig.Groups`, same `json:"groups,omitempty"` key the dashboard writes).
- Propagate `ic.Groups` into the constructed `tgCfg` in `buildChannel`.

No new config surface, no migration, no API/contract change — DB instances now honor the same per-topic gating the file-config path already supported.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./internal/channels/telegram/`
- [ ] Web UI builds (no UI changes)
- [x] SQL queries use parameterized `$1, $2` (no SQL changes)
- [ ] New user-facing strings in all 3 locales (no new strings)
- [ ] Migration version bumped (no migration)

## Test Plan
Added `factory_test.go`:
- `TestFactory_PropagatesGroupsAndTopics` — group disabled + topic `13` re-enabled (the "reply in one thread only" setup): topic 13 resolves enabled, while topic 99 and the General topic inherit the disabled group and stay silent.
- `TestFactory_NilGroupsWhenAbsent` — omitting `groups` leaves `Groups` nil and defaults to enabled, with no panic.

Manual: a dashboard-created Telegram forum channel with group `enabled:false` + one topic `enabled:true` now replies only in that topic.

### DIFF
--- a/internal/channels/telegram/factory.go
+++ b/internal/channels/telegram/factory.go
@@ -20,26 +20,27 @@ type telegramCreds struct {
 
 // telegramInstanceConfig maps the non-secret config JSONB from the channel_instances table.
 type telegramInstanceConfig struct {
-	APIServer         string                     `json:"api_server,omitempty"`
-	Proxy             string                     `json:"proxy,omitempty"`
-	DMPolicy          string                     `json:"dm_policy,omitempty"`
-	GroupPolicy       string                     `json:"group_policy,omitempty"`
-	RequireMention    *bool                      `json:"require_mention,omitempty"`
-	MentionMode       string                     `json:"mention_mode,omitempty"`
-	HistoryLimit      int                        `json:"history_limit,omitempty"`
-	DMStream          *bool                      `json:"dm_stream,omitempty"`
-	GroupStream       *bool                      `json:"group_stream,omitempty"`
-	DraftTransport    *bool                      `json:"draft_transport,omitempty"` // sendMessageDraft for DM streaming (default true)
-	ReasoningDelivery string                     `json:"reasoning_delivery,omitempty"`
-	ReasoningStream   *bool                      `json:"reasoning_stream,omitempty"` // show reasoning as separate message (default true)
-	ReactionLevel     string                     `json:"reaction_level,omitempty"`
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
## Review: Approve ✅

**Summary**: Fixes DB-instance Telegram channels silently dropping per-group/per-topic overrides by adding the missing `Groups` field to `telegramInstanceConfig` and propagating it to the resolved channel config.

**Risk level**: Low — narrow scope, mirrors existing file-config path, no migration/API changes.

**Mandatory gates**:
- Duplicate / prior implementation: clear — no overlapping PRs or issues found
- Project standards: clear — follows existing struct patterns and JSON key conventions
- Strategic necessity: clear value — dashboard-created instances now honor the same per-topic gating as file-config instances

**Findings**: None. Clean implementation with meaningful test coverage.

*Posted by github-maintain cron at 2026-06-29T16:53Z*

PR #1307, opened 2026-06-29
fix(ollama): always send num_ctx to prevent context-window errors

bclermont:
Fix two-layer bug preventing num_ctx from being sent to Ollama:

1. resolveOllamaNumCtx was returning nil when /api/show returned the default (131072)
2. OllamaProvider.buildRequest had no fallback — omitted num_ctx when nil

Result: Ollama defaulted to 4096 context, causing 'exceeds context size' errors.

Fix: Both layers now always send num_ctx (131072 default or user-configured value).

- Implement native OllamaProvider using github.com/ollama/ollama/api
- Fix config-registered providers to fetch model context
- Always inject num_ctx into request, never rely on Ollama defaults

Tested: Ollama successfully processes 11K+ token prompts without context overflow.


clark-cant:
🦸‍♂️ **CI check failed** — the `go` test job has 2 failing subtests in `TestRegisterInMemoryOllamaURLNormalization`:

```
providers_ollama_url_test.go:265: registered provider does not implement APIBase(), type = *providers.OllamaProvider
```

**Root cause:** The new native `OllamaProvider` struct in `internal/providers/ollama.go` does not implement the `APIBase()` method that the URL normalization tests expect. The test at `providers_ollama_url_test.go:265` type-asserts the registered provider to an interface requiring `APIBase()`.

**Fix:** Add an `APIBase() string` method to `OllamaProvider` returning the resolved API base URL (e.g. the normalized `http://host:11434/v1` form).

**CI run:** https://github.com/nextlevelbuilder/goclaw/actions/runs/28396617349

*Posted by github-maintain at 2026-06-29T19:25:00Z*

clark-cant (approved):
✅ CI all green, merge state CLEAN. LGTM — approving and merging.

PR #1311, opened 2026-06-30
feat(usage): preserve cache tokens in usage events

zezaeoh:
## Summary
- add cache/thinking token fields to usage_events and hourly rollups
- propagate provider usage metadata into event records so event-level audit keeps cache reads, cache creates, and thinking tokens
- expose the new fields in usage analytics/detail DTOs and web usage event types

## Context
This complements #1309 and #1310. Trace/snapshot accounting already preserves Codex cache reads via span metadata and usage_snapshots, but usage_events only kept input/output/total tokens, so detailed event audit could lose cache/thinking token detail.

## Tests
- `go test ./internal/agent ./internal/store/pg ./internal/http`
- `go test ./internal/store ./internal/providers ./internal/tracing`
- `go test -tags sqliteonly ./internal/store/sqlitestore -run 'TestSQLiteUsageEventStoreRefreshRollupHourIsIdempotent|TestSQLiteUsageEventStore'`
- `go build ./...`
- `go build -tags sqliteonly ./...`

## Notes
- Full `go test -tags sqliteonly ./internal/store/sqlitestore` still hits the pre-existing migration fixture issue around duplicate `provider_id` in v49; unrelated to this change.
- `pnpm --dir ui/web build` was started but timed out locally before completion without a concrete failure output.

PR #1312, opened 2026-06-30
fix(sqlite): make column migrations idempotent

zezaeoh:
## Summary
- make SQLite v49 cron provider override migration idempotent when columns already exist
- make SQLite v52 usage token migration idempotent when cache/thinking token columns already exist

## Verification
- go test -tags sqliteonly ./internal/store/sqlitestore
- go test ./internal/agent ./internal/store/pg ./internal/http
- go test ./internal/store ./internal/providers ./internal/tracing
- go build ./...
- go build -tags sqliteonly ./...
- pnpm --dir ui/web build

### DIFF
--- a/internal/store/sqlitestore/schema.go
+++ b/internal/store/sqlitestore/schema.go
@@ -1411,6 +1411,18 @@ func EnsureSchema(db *sql.DB) error {
 					patch = `SELECT 1;`
 				}
 			}
+			if v == 49 {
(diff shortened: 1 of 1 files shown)


clark-cant (approved):
LGTM — clean idempotent migration fix for SQLite v49 (cron_jobs provider columns) and v52 (usage_events token columns). Correct approach: check column existence before ALTER TABLE, SELECT 1 no-op when all columns present. Low risk, well-scoped follow-up to #1311.

PR #1313, opened 2026-07-01
fix(channels): honor paired DM policy before publish

itsddvn:
## Summary
- Add explicit policy-checked channel publish paths so paired DM senders are not dropped by the shared static allowlist safety gate after policy approval.
- Route Zalo OA, Zalo Personal, Bitrix24, and Slack through the authorized publish path after their policy gates pass.
- Keep the default direct-message publish safety net for adapters without explicit policy gates, and add regressions for paired DM plus guard cases.

## Channel audit
- Zalo OA / Zalo Personal: policy gate runs before inbound publish; now uses authorized publish for DM.
- Bitrix24: DM/group policy gates run before media publish; now uses authorized media publish.
- Slack: DM/group policy gates run before publish; removed duplicate DM allowlist drop after policy and carries authorization through debounce.
- WhatsApp: already scopes allowlist fallback to open DM policy, pairing remains pair-based.
- Discord / Feishu: publish directly after policy gates, no shared BaseChannel allowlist recheck.
- Telegram: custom handler path, not affected by this BaseChannel gate.
- Pancake / Facebook: no DM policy gate; still rely on default allowlist safety net.

## Verification
- `go test ./internal/channels/... -count=1`
- `go test ./cmd -run 'Pairing|Channel|Consumer' -count=1`
- `go build ./...`
- `go vet ./...`

### DIFF
--- a/docs/project-changelog.md
+++ b/docs/project-changelog.md
@@ -4,6 +4,25 @@ Significant changes, features, and fixes in reverse chronological order.
 
 ---
 
+## 2026-07-01
+
+### Paired DM routing after policy checks
+
+**Fixes**
+
+- Policy-checked direct messages now pass the shared channel safety gate for
+  Zalo OA, Zalo Personal, Bitrix24, and Slack when the sender is paired but not
+  listed in static `allow_from`.
+- Direct-message callers without an explicit policy gate still use the default
+  `allow_from` safety net.
+
+**Tests**
+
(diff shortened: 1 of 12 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. The authorized publish path correctly preserves the DM safety net for channels without explicit policy gates while allowing policy-checked paired senders through. Good regression test coverage for Zalo and Slack paired DMs. Low risk.

PR #1314, opened 2026-07-01
Fix channel paired DM routing and allowlist config clears

itsddvn:
## Summary
- Honor paired DM policy before channel publish routing.
- Fix channel Advanced Settings config merge so clearing Allowed Users removes stale allow_from instead of preserving old config.
- Add regression coverage for clearing allow_from, preserving essential/group config, empty tag arrays, and nested chat_behavior config.

## Testing
- pnpm --dir ui/web test -- src/pages/channels/channel-detail/channel-advanced-dialog.test.ts
- pnpm --dir ui/web build
- make build-full
- curl -fsS http://127.0.0.1:18790/health
- curl -fsSI http://127.0.0.1:18790/

## Notes
- Rebuilt and restarted local embedded UI on port 18790.
- Unrelated dirty working tree files were left untouched.

### DIFF
--- a/docs/project-changelog.md
+++ b/docs/project-changelog.md
@@ -4,6 +4,25 @@ Significant changes, features, and fixes in reverse chronological order.
 
 ---
 
+## 2026-07-01
+
+### Paired DM routing after policy checks
+
+**Fixes**
+
+- Policy-checked direct messages now pass the shared channel safety gate for
+  Zalo OA, Zalo Personal, Bitrix24, and Slack when the sender is paired but not
+  listed in static `allow_from`.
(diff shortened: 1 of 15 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. The authorized publish path correctly preserves the DM safety net for channels without explicit policy gates while allowing policy-checked paired senders through. Slack debounce threading of the authorized flag is a nice touch. UI config merge fix properly handles clearing allow_from via flatten/unflatten. Good regression test coverage for Zalo, Slack, and Bitrix24 paired DMs. Low risk.

PR #1317, opened 2026-07-01
fix: require usage event cache token schema

itsddvn:
## Summary
- Bump the required PostgreSQL schema version to include usage event cache/thinking token columns.
- Add a regression test that keeps RequiredSchemaVersion aligned with the latest up migration.
- Document the Resource Event Analytics schema gate fix in the project changelog.

## Root cause
The binary allowed startup against schema version 87 even though Resource Event Analytics now reads columns introduced by migration 000088_usage_events_cache_tokens.up.sql.

## Test plan
- [x] go test ./internal/upgrade
- [x] go build ./...
- [x] go build -tags sqliteonly ./...
- [x] go vet ./...
- [x] go test ./...

Related issues: none found in GitHub issue search.

### DIFF
--- a/docs/project-changelog.md
+++ b/docs/project-changelog.md
@@ -6,6 +6,19 @@ Significant changes, features, and fixes in reverse chronological order.
 
 ## 2026-07-01
 
+### Usage event analytics schema gate
+
+**Fixes**
+
+- Bumped required PostgreSQL schema version to include the usage event cache and
+  thinking token migration required by Resource Event Analytics.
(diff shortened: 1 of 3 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. Bumping RequiredSchemaVersion from 87→88 to match migration 000088 (usage_events cache/thinking tokens) prevents startup against an incomplete schema. The regression test (glob migrations, assert parity with RequiredSchemaVersion) is excellent — prevents future drift. Changelog entry is clear. Low risk, pure additive safety net.

PR #1320, opened 2026-07-01
fix: infer text-encoded tool calls by schema

nguyenha935:
## Summary
- Root cause: some OpenAI-compatible providers return tool-call markers inside assistant content while native tool_calls is missing or null.
- Normalize text-encoded tool-call markers into ToolCall records when the target tool can be identified deterministically.
- Preserve native tool calls, handle multiple text markers independently, and fail closed when schema matching is ambiguous.

## Test Plan
- docker run --rm -v /root/.config/superpowers/worktrees/Goclaw/source-runtime:/src -w /src golang:1.26-bookworm go test ./internal/providers -run TestOpenAIProviderChat_
- docker run --rm -v /root/.config/superpowers/worktrees/Goclaw/source-runtime:/src -w /src golang:1.26-bookworm go test ./internal/agent -run Test
- Runtime check: Kimchi returned marker content with provider_tool_calls_type=null; GoClaw resolved tool_calls_count=1 and invoked tts.

## Safety
- Ambiguous command-like markers are stripped and logged instead of being executed.
- Native tool calls remain unchanged.
- Existing final-output sanitizers remain in place as defense in depth.

## Cross-surface parity
- Gateway/provider runtime: covered by OpenAI provider tests.
- Web UI: N/A.
- DB/migration: N/A.
- CLI/runtime command: N/A.
- API contract: N/A.

### DIFF
--- a/internal/providers/adapter_openai.go
+++ b/internal/providers/adapter_openai.go
@@ -112,8 +112,10 @@ func (a *OpenAIAdapter) FromStreamChunk(data []byte) (*StreamChunk, error) {
 
 	// Text content
 	if delta.Content != "" {
-		sc.Content = delta.Content
-		hasContent = true
+		normalized := normalizeControlOutput(delta.Content, "", nil)
+		sc.Content = normalized.Content
+		sc.Thinking = strings.TrimSpace(strings.Join([]string{sc.Thinking, normalized.Thinking}, "\n"))
+		hasContent = sc.Content != "" || sc.Thinking != ""
 	}
 
 	if !hasContent {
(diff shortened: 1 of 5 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix for providers that embed tool-call markers in text content (Kimi/Moonshot pattern). Schema-based resolution is solid: exact header match → single-tool fallback → unique schema match by required properties. Fails closed on ambiguity (strips + logs instead of guessing). Streaming normalizer correctly handles split chunks via safeControlCut pending buffer. Excellent test coverage (8 cases: Kimi markers, multi-tool schema inference, ambiguity rejection, mixed native+text, multiple text calls, thinking tags, streaming split). Low risk, pure additive change.

PR #1322, opened 2026-07-01
fix: resolve duplicated tenant-path segment in team shared-workspace resolution

bclermont:
## Problem
  
  When an agent queried its team's shared workspace path (e.g. via a system/tool-facing prompt), the reported path contained a duplicated tenant segment:

  /app/workspace/tenants/<teamName>/tenants/<teamName>/teams/<teamID>

  instead of the correct, single-tenant-segment path actually used on disk (confirmed via the web UI's file-upload path, which lands correctly at):

  /app/workspace/tenants/<teamName>/teams/<teamID>

  ## Root cause

  `internal/agent/loop_context.go` was re-applying tenant scoping (`TenantLayer` / `TenantID`) on top of `l.dataDir`, which was **already** tenant-scoped upstream by `internal/agent/resolver.go` (via `config.TenantDataDir`). This double-application of
  tenant scoping only affected the agent-facing workspace resolution path (`tools.ResolveWorkspace` and `workspace.Resolve` call sites in `loop_context.go`) — the web-upload path in `internal/http/files.go` was already correct, since it applies tenant
  scoping exactly once from a raw base dir.

  Investigation surfaced that tenant-path joining logic was duplicated across four independent implementations with no single source of truth for "is this base dir already tenant-scoped":
  - `config.TenantDataDir`
  - `config.TenantWorkspace`
  - `tools.TenantLayer` / `tools.ResolveWorkspace`
  - `tenantPath()` in `internal/workspace/resolver_impl.go`

  ## Fix

  1. **Bug fix:** Removed the redundant tenant-scoping calls in `internal/agent/loop_context.go` (4 call sites), since `l.dataDir` is already tenant-scoped. Verified via grep that other callers of the same helpers operate on a *raw* (non-tenant-scoped)
  base dir and were correctly left untouched.
  2. **Consolidation refactor:** Introduced a single canonical function, `config.TenantScopedDir(base, tenantID, tenantSlug string)` in `internal/config/tenant_paths.go`, containing the sole path-join + master-tenant no-op + empty-slug fallback +
  path-traversal-defense logic. All previously-duplicated implementations (`TenantDataDir`, `TenantWorkspace`, `tools.TenantLayer`, `workspace.tenantPath`) now delegate to it instead of re-implementing the join logic, eliminating this entire bug class
  going forward.

  ## Files changed

  - `internal/agent/loop_context.go` — removed redundant tenant-scoping at 4 call sites
  - `internal/config/tenant_paths.go` — new canonical `TenantScopedDir`; `TenantDataDir`/`TenantWorkspace` now delegate to it
  - `internal/config/tenant_paths_test.go` — new tests for `TenantScopedDir` and delegation equivalence
  - `internal/workspace/resolver_impl.go` — `tenantPath()` now delegates to `config.TenantScopedDir`
  - `internal/tools/workspace_resolver_test.go` — regression test pinning no double-join via `tools.ResolveWorkspace`
  - `internal/workspace/resolver_impl_test.go` — regression test pinning no double-join via `workspace.Resolve`

  ## Testing

  - `go build ./...` — pass
  - `go build -tags sqliteonly ./...` — pass
  - `go vet ./...` — clean
  - `go test ./...` — all packages pass, including new/updated regression tests
  - `go test -race ./tests/integration/` — skipped locally (no Postgres reachable at `localhost:5433`)
  - Full local CI (`make ci`: Go build + test + vet + web build) — verified green

  ## Surface parity

  - **Gateway server:** fixed (workspace resolution).
  - **API contract:** N/A — no request/response shape changes, this only affects internal path construction.
  - **Web UI:** N/A — web-upload path was already correct and unaffected by this change.
  - **CLI/runtime package:** N/A — no CLI-facing behavior change.

### DIFF
--- a/internal/agent/loop_context.go
+++ b/internal/agent/loop_context.go
@@ -212,9 +212,11 @@ func (l *Loop) injectContext(ctx context.Context, req *RunRequest) (contextSetup
 		ctx = tools.WithToolTeamID(ctx, req.TeamID)
 		// Team root for dispatched tasks: resolve the UserChatLayer-stripped root
 		// so the dispatched agent can still read peer-scoped files in the same team.
+		// l.dataDir is already tenant-scoped (see resolver.go: config.TenantDataDir),
+		// so TenantLayer must NOT be reapplied here — doing so double-joins the
+		// tenant segment (tenants/<slug>/tenants/<slug>/teams/<id>).
 		if teamUUID, err := uuid.Parse(req.TeamID); err == nil && l.dataDir != "" {
 			teamRoot := tools.ResolveWorkspace(l.dataDir,
-				tools.TenantLayer(store.TenantIDFromContext(ctx), store.TenantSlugFromContext(ctx)),
 				tools.TeamLayer(teamUUID),
 			)
 			ctx = tools.WithToolTeamRoot(ctx, teamRoot)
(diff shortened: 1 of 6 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. The consolidation into a single canonical `TenantScopedDir()` eliminates an entire bug class. Good regression tests pinning the no-double-join invariant at both the tools and workspace layers. Path traversal defense preserved. Low risk.

PR #1323, opened 2026-07-02
fix(agents): pre-aggregate usage snapshots before deletion

bclermont:
## Summary

- When an agent is deleted, `usage_snapshots` rows have `agent_id` set to `NULL` via FK `ON DELETE SET NULL`.
- If aggregate (NULL-agent) rows already exist for the same `(bucket_hour, provider, model, channel, tenant_id)`, the unique index `idx_usage_snapshots_unique` causes a constraint violation.
- This PR adds `aggregateAndDeleteAgentUsage()` which upserts the agent's usage rows into the existing NULL-agent aggregate rows (summing token counts) before deletion, then deletes the original agent rows — eliminating the collision entirely.

## Test plan

- [ ] Delete an agent that has usage_snapshots; verify no unique constraint error is raised.
- [ ] Verify the NULL-agent aggregate row reflects the summed token counts from the deleted agent.
- [ ] Delete an agent with no usage_snapshots; verify deletion succeeds without error.
- [ ] Delete multiple agents with overlapping bucket/provider/model/channel/tenant; verify aggregate accumulates correctly.

### DIFF
--- a/internal/store/pg/agents.go
+++ b/internal/store/pg/agents.go
@@ -391,7 +391,37 @@ func isEmptyOrNullJSONUpdate(v any) bool {
 	}
 }
 
+func (s *PGAgentStore) aggregateAndDeleteAgentUsage(ctx context.Context, agentID uuid.UUID) error {
+	// Upsert this agent's usage_snapshots into NULL-agent aggregate rows, then delete the agent rows.
+	// This is necessary because the unique index uses COALESCE(agent_id, '00000000...') so ON DELETE SET NULL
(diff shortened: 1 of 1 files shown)


clark-cant (approved):
LGTM — clean, well-scoped bugfix. The SQL correctly pre-aggregates agent-specific usage into NULL-agent aggregate rows via INSERT...SELECT...ON CONFLICT DO UPDATE (SUM tokens), then deletes the original rows. Eliminates the unique constraint violation that occurs when FK ON DELETE SET NULL collides with existing aggregate rows. Error wrapping and context propagation are correct. Low risk, pure additive change.

PR #1324, opened 2026-07-02
fix(providers): preserve streamed content whitespace

nguyenha935:
## Summary
- Preserve raw OpenAI-compatible stream content deltas instead of normalizing each delta independently.
- Keep regular assistant content whitespace intact while still extracting text-encoded tool calls and thinking tags.
- Add regression tests for leading whitespace in stream chunks and whitespace across streamed content deltas.

## Test Plan
- `docker run --rm -v "$PWD":/src -w /src golang:1.26-bookworm go test ./internal/providers -count=1`

### DIFF
--- a/internal/providers/adapter_openai.go
+++ b/internal/providers/adapter_openai.go
@@ -112,10 +112,8 @@ func (a *OpenAIAdapter) FromStreamChunk(data []byte) (*StreamChunk, error) {
 
 	// Text content
 	if delta.Content != "" {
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean fix for streaming whitespace loss. Correct insight: per-delta TrimSpace eats inter-chunk spaces (especially visible in Vietnamese/CJK tokenization). Moving normalization to aggregate level while keeping thinking-tag and tool-call extraction intact is the right approach. Good regression coverage. Low risk.

PR #1326, opened 2026-07-02
fix(zalo): preserve nested reply context

itsddvn:
## Summary
- preserve Zalo Personal quote payloads so agents can read nested reply chains
- add bounded channel-agnostic reply context cache with depth, TTL, size, and cycle guards
- parse Zalo quote/protocol payloads more defensively, including string/numeric IDs and attachment placeholders
- run Zalo Personal policy checks before media extraction/download/cache writes

Trace: 019f22c1-3342-7864-9b97-a59fa739cdae

## Test plan
- [x] go test ./internal/channels/replycontext ./internal/channels/zalo/...
- [x] go build ./...

PR #1331, opened 2026-07-02
fix(zalo): preserve root quote in reply chains

itsddvn:
## Summary
- preserve the original quoted message when Zalo Personal replies go 2-6 levels deep
- store the immediate parent quote snapshot in the channel-agnostic reply-context cache
- render the saved parent quote as root fallback when the parent ID is not cached, such as outbound/original Zalo messages
- add generic and Zalo regression tests for deep reply chains with an uncached original root

Trace evidence:
- first reply trace: 019f23bb-4119-7b23-b987-2abc19ade192
- deep reply trace: 019f23bc-86e2-70a7-94ef-76e83e5db3c4

## Test plan
- [x] go test ./internal/channels/replycontext ./internal/channels/zalo/...
- [x] go test ./internal/channels/replycontext ./internal/channels/zalo/... -race
- [x] go vet ./internal/channels/replycontext ./internal/channels/zalo/...
- [x] go build ./...
- [x] go build -tags sqliteonly ./...
- [x] go test ./...

### DIFF
--- a/internal/channels/replycontext/reply-context-cache.go
+++ b/internal/channels/replycontext/reply-context-cache.go
@@ -23,11 +23,12 @@ type Scope struct {
 
 // Message is a platform-normalized accepted message stored for future replies.
 type Message struct {
-	IDs       []string
-	ParentIDs []string
-	Sender    string
-	Body      string
-	CreatedAt time.Time
+	IDs         []string
+	ParentIDs   []string
+	ParentQuote Quote
+	Sender      string
+	Body        string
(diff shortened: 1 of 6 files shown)


clark-cant (approved):
LGTM — clean follow-up to #1326. Storing the immediate parent quote snapshot in the cache node and using it as fallback when the root is evicted/uncached is the right approach. Good test coverage for 6-level deep chains with uncached root. No breaking changes.

PR #1337, opened 2026-07-03
feat(hooks): add debug logging, capture console.log output, fail clos…

bclermont:
## fix(hooks): add debug logging, capture console.log output, fail closed on matched-then-errored hooks

  ### Problem

  `prefilter()` had zero logging — matcher regex failures, tool-name mismatches, CEL compile
  errors, and CEL evaluating to false were all completely silent. A misconfigured hook was
  indistinguishable from a correctly-configured one that simply chose not to fire: there was no
  way to tell whether a hook was ever considered, whether its matcher matched, whether its
  condition evaluated, or why a decision was made.

  This was confirmed live: a `PreToolUse` script hook intended to restrict file writes silently
  failed to block an unauthorized file creation, with no log line anywhere indicating what
  happened.

  ### Changes

  - **Debug logging throughout hook dispatch** — new greppable log lines:
    `hooks.prefilter.considered`, `hooks.prefilter.matcher_checked`,
    `hooks.prefilter.condition_evaluated`, `hooks.prefilter.error`, `hooks.dispatch.decision`.
    Covers matcher match/miss, CEL compile/eval result, and the final decision with reasoning —
    both for the blocking (`runSync`) and non-blocking (`runAsync`) dispatch paths.

  - **`console.log`/`console.error` output captured** — script hooks already had a `console`
    binding into the goja sandbox, but its output was discarded after execution. It's now
    surfaced via `slog.Debug` immediately, and persisted into the hook's audit record
    (`HookExecution.ConsoleOutput` + `Metadata["console_output"]`, using the existing `metadata`
    JSON column — no migration needed) so it's visible in the hook execution history.

  - **Fail closed, narrowly scoped** — when a hook's matcher successfully matches a tool call but
    CEL condition or script execution then errors, the dispatcher now blocks that specific tool
    call instead of silently treating the failure as a non-match and letting the call proceed.
    This does **not** turn one broken hook into a global outage: hooks that legitimately don't
    match remain unaffected, and other unrelated tool calls are never touched. It only blocks the
    one gated action that couldn't be confidently evaluated.

  ### Tests

  - Matched-then-CEL-errors → blocks the call, handler never invoked.
  - Matcher-doesn't-match → proceeds normally, unaffected (confirms the fail-closed scope is
    narrow, not global).
  - `console.log` output is captured and present in the audit record.

  ### Verified live

  Deployed and tested against a real hook configured to restrict `docker-engineer`'s file writes
  to `Dockerfile`/`entrypoint.sh`/`.dockerignore`. The new logging immediately revealed the actual
  root cause of the hook not firing (matcher only checked lowercase tool names, but the agent was
  calling the capitalized `Write` alias) — something that was previously undiagnosable.

### DIFF
--- a/internal/hooks/dispatcher.go
+++ b/internal/hooks/dispatcher.go
@@ -202,11 +202,30 @@ func (d *stdDispatcher) runSync(ctx context.Context, ev Event, chain []HookConfi
 		if !cfg.Enabled {
 			continue
 		}
+		slog.Debug("hooks.prefilter.considered",
+			"hook_id", cfg.ID, "tool_name", evMut.ToolName, "agent_id", evMut.AgentID, "hook_event", evMut.HookEvent)
 		if d.cb.isTripped(cfg.ID, d.now()) {
-			d.writeExec(ctx, cfg, evMut, DecisionBlock, 0, "circuit breaker open")
+			slog.Warn("hooks.dispatch.decision",
+				"hook_id", cfg.ID, "decision", "block", "reason", "circuit breaker open")
+			d.writeExec(ctx, cfg, evMut, DecisionBlock, 0, "circuit breaker open", "")
 			return FireResult{Decision: DecisionBlock}, nil
 		}
-		if !d.prefilter(cfg, evMut) {
+		pf := d.prefilter(cfg, evMut)
+		if pf.errored {
+			// Fail-closed (narrow scope): the matcher matched this call, but
+			// the CEL condition then failed to compile/evaluate. We cannot
+			// know whether the hook intended to block or allow, so we treat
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean, well-scoped implementation. Debug logging throughout hook dispatch with greppable log lines. console.log output capture for script hooks (persisted in audit record via metadata JSON, no migration needed). Fail-closed behavior narrowly scoped to only block the specific tool call that couldn't be evaluated, not global outage. Good test coverage for matched-then-error, matcher-miss, and console output capture paths. Verified live against real hook configuration. Ready to merge.

PR #1344, opened 2026-07-04
fix(docker): add /app to PATH so agents can use the bundled goclaw self-management skill

bclermont:
The gateway binary (/app/goclaw) has embedded operator/admin subcommands (agent list, doctor, skills list, sessions list, providers list, channels list, cron list, traces list, etc) that the bundled goclaw skill teaches agents to use for self-administering the gateway. But /app was never added to PATH -- confirmed live via docker exec: `which goclaw` empty, while `/app/goclaw --help` works fine directly.

Non-sandboxed agent exec runs in the same container as the gateway itself, so /app/goclaw is a real, reachable file for those agents -- the only blocker was PATH resolution for the bare `goclaw` command.

Prepend /app to PATH in docker-entrypoint.sh, ahead of the existing runtime-installed package directories, so bare `goclaw` commands resolve correctly for both operators (docker exec) and agents (exec tool) alike.


clark-cant (approved):
LGTM — clean one-liner. /app/goclaw exists in the container, agents need bare `goclaw` on PATH for the self-management skill. Correctly placed ahead of runtime dirs. Ship it.

PR #1345, opened 2026-07-04
feat(webhooks): split webhook usage tokens by cache vs non-cache

thotam:
## Summary
Webhook `usage` responses now split token usage into **cache vs non-cache** by surfacing `cache_read_input_tokens`, `cache_creation_input_tokens`, and `prompt_tokens_include_cached_segments` — for both the sync LLM webhook response and the async callback payload. Previously the envelopes reported only `prompt_tokens`/`completion_tokens`/`total_tokens`, so cached input (much cheaper) was indistinguishable from non-cached input.

The data already flowed end-to-end via `providers.Usage` and `agent.RunResult.Usage`; the two webhook envelope structs simply dropped the cache fields. This change copies them through, mirroring `providers.Usage` JSON tags exactly. The async mapping is extracted into a pure `newCallbackUsage` helper for unit testing.

Non-cached input is derived by the caller as `prompt_tokens - cache_read_input_tokens` when `prompt_tokens_include_cached_segments` is `true`.

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev` — feature.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes — for the changed packages (`internal/http`, `internal/webhooks`); repo-wide `go vet ./...` has a pre-existing failure in `internal/tools/document_parser_test.go` unrelated to this change
- [ ] Tests pass: `go test -race ./...` — `-race` not runnable locally (no CGO/gcc on the Windows dev box); CI runs it on Linux. Webhook suites (`go test ./internal/webhooks/ ./internal/http/ -run Webhook`) pass.
- [ ] Web UI builds — N/A, no UI changes (webhook response envelope is an external-consumer API contract; no web/desktop/CLI surface consumes it)
- [x] SQL queries use parameterized `$1, $2` — N/A, no SQL changes
- [ ] New user-facing strings added to all 3 locales — N/A, no user-facing UI strings
- [ ] Migration version bumped — N/A, no schema change (cache data already exists in `providers.Usage`; `response` column stores free-form JSON)

**Additive / non-breaking:** all new fields use `omitempty`, so responses without prompt caching are byte-identical to before.

**Surface parity:** server-only. Web UI / Desktop / CLI N/A — no surface consumes the webhook usage envelope struct. Migration N/A — no schema change. Message webhook (`webhooks_message.go`) intentionally returns no usage, unchanged.

## Test Plan
- **Sync path:** extended `TestWebhookLLMHandler_SyncHappyPath` — stub `providers.Usage` carries `CacheReadTokens`/`CacheCreationTokens`/`PromptTokensIncludeCachedSegments`; asserts they propagate through the full HTTP handler round-trip onto the JSON response body.
- **Async path:** added `TestNewCallbackUsageMapsCacheTokens` — unit-tests the `newCallbackUsage` mapper for the nil case and full field mapping (base + cache tokens + include-cached flag). The `callbackUsage` payload is serialized into the `response` column and returned on both the callback POST and status polls, so cache tokens appear on both.
- `go build ./...` + `go build -tags sqliteonly ./...` + webhook test suites all green.

### DIFF
--- a/docs/webhooks.md
+++ b/docs/webhooks.md
@@ -266,11 +266,19 @@ Triggers an agent with an input prompt. Available in all editions.
   "usage": {
     "prompt_tokens": 150,
     "completion_tokens": 200,
-    "total_tokens": 350
+    "total_tokens": 350,
+    "cache_read_input_tokens": 120,
+    "cache_creation_input_tokens": 30,
+    "prompt_tokens_include_cached_segments": true
   },
   "finish_reason": "stop"
 }
 ```
+
+> `cache_read_input_tokens` and `cache_creation_input_tokens` are present only when
+> prompt caching was active (omitted otherwise). When
+> `prompt_tokens_include_cached_segments` is `true`, `prompt_tokens` already counts
+> the cached segments, so non-cached input = `prompt_tokens - cache_read_input_tokens`.
 
 Sync mode times out after the configured deadline (default **600s**). On timeout: `504 Gateway Timeout` with `webhook.llm_timeout`.
 
@@ -428,7 +436,10 @@ User-Agent: goclaw-webhook/1
   "usage": {
     "prompt_tokens": 150,
     "completion_tokens": 200,
-    "total_tokens": 350
+    "total_tokens": 350,
+    "cache_read_input_tokens": 120,
+    "cache_creation_input_tokens": 30,
+    "prompt_tokens_include_cached_segments": true
   },
   "metadata": {},
   "error": ""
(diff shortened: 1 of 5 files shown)


clark-cant (approved):
LGTM — clean, well-scoped change. Cache token fields (cache_read_input_tokens, cache_creation_input_tokens, prompt_tokens_include_cached_segments) already flow end-to-end via providers.Usage; this correctly surfaces them in both the sync LLM webhook response and async callback payload. Good extraction of newCallbackUsage helper for testability. omitempty preserves backward compatibility. Docs updated. Tests cover both paths.

PR #1346, opened 2026-07-04
fix(web): catch unhandled promise rejections on config save buttons

bclermont:
## Problem

Saving a config change in the web UI produced an uncaught console error:
```
Uncaught (in promise) ApiError: failed to save config: open /app/data/config.json: permission denied
```

`useConfig()`'s `patch()`/`applyRaw()` already correctly catch `ApiError`, show a toast, and
re-throw (the established pattern used by ~19 other data hooks in this app) -- but nothing
caught that re-thrown promise at the config page's save-button call sites, producing console
noise on every failed save even though the user did see a toast.

## Fix

Two-layer fix:

1. **Local fix** at each affected call site (shell-security, server, channels, providers,
   telemetry, behavior, quota, bindings, cron, tools-exec, tools-browser, tools-profile config
   sections), matching the pre-existing correct convention already used by
   `ai-defaults-section.tsx`.

2. **Global safety net** in `main.tsx`: a `window` `unhandledrejection` listener that suppresses
   console noise specifically for `ApiError` instances (already toasted by the originating
   hook), while leaving non-`ApiError` rejections untouched so real bugs still surface. This
   covers all existing call sites plus any future ones that forget to catch, without requiring
   every future save button to remember this boilerplate.

Confirmed no React Query mutation/global `onError` mechanism exists in this app -- the
hand-rolled toast+rethrow convention in each data hook is already the de-facto global
error-display mechanism; the actual gap was purely the missing safety net for re-thrown
promises.

Note: the underlying `open /app/data/config.json: permission denied` error itself is a separate
Docker/deployment file-ownership issue, not addressed by this PR.

## Verified

- `tsc --noEmit` clean
- `npm run build` succeeds
- `npx vitest run` -- 313 tests pass

### DIFF
--- a/tests/integration/v3_test_helper.go
+++ b/tests/integration/v3_test_helper.go
@@ -106,6 +106,20 @@ func seedTenantAgent(t *testing.T, db *sql.DB) (tenantID, agentID uuid.UUID) {
 
 	// Cleanup after test — delete in FK order (children first, parents last).
 	t.Cleanup(func() {
+		// Knowledge stores — must run BEFORE agent_teams is deleted. vault_documents.team_id
+		// is ON DELETE SET NULL with a BEFORE UPDATE trigger that force-sets scope='personal',
+		// which violates vault_documents_scope_consistency for team-scoped docs whose agent_id
+		// is NULL. Deleting the rows outright avoids the trigger entirely.
+		// vault_links has no tenant_id column (it links vault_documents by from/to_doc_id) and
+		// cascades automatically via vault_links_{from,to}_doc_id_fkey ON DELETE CASCADE, so no
+		// explicit delete is needed here.
+		db.Exec("DELETE FROM vault_documents WHERE tenant_id = $1", tenantID)
+		db.Exec("DELETE FROM kg_dedup_candidates WHERE tenant_id = $1", tenantID)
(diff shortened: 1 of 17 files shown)


bclermont:
build failed because of #1347

bclermont:
build pass

clark-cant (approved):
LGTM — clean, well-scoped fix. Two-layer approach is correct: local .catch() at call sites matches existing convention (ai-defaults-section.tsx), and the global unhandledrejection safety net in main.tsx is well-designed — only suppresses ApiError (already toasted by hooks), leaves genuine bugs untouched.

Test cleanup reorder is also correct: vault_documents.team_id ON DELETE SET NULL trigger was firing before agent_teams deletion, causing FK violations. Moving knowledge-store deletes before team-related deletes avoids the trigger entirely. Good catch on the agent_user_profiles → user_agent_profiles schema correction too.

Minor note: shell-security-section.test.tsx is trivial (just 'is defined') — harmless but adds no real coverage. Not blocking.

Mandatory gates:
- Duplicate/prior implementation: clear
- Project standards: passed
- Strategic necessity: clear value — eliminates console noise without hiding real bugs

Verdict: APPROVE

Issue #1000, opened 2026-04-22
feat: native image generation for Codex + OpenAI-compat providers


kaitranntt:
## Summary

Add native `image_generation` support to the providers layer so agents routed through ChatGPT (Codex OAuth) accounts can produce inline images in chat. Parallel forward-compat support for the OpenAI-compatible `/v1/chat/completions` track.

## Scope

**Track A — Codex native (primary):**
- Request: declare `image_generation` as a first-class tool object in the `tools[]` of `POST /codex/responses` (fields: `type`, `action`, `model`, optional `output_format`, `size`, `partial_images`).
- Response: handle native SSE events `response.image_generation_call.partial_image` and `response.output_item.done` with `item.type == "image_generation_call"`, plus non-stream completion via `response.output[]` walk.
- Decode final base64 PNG/JPEG/WEBP into `Message.Images`.

**Track B — OpenAI-compat (parallel track, forward-compat):**
- Serialize `ToolDefinition{Type:"image_generation"}` as `{"type":"image_generation"}` in `/v1/chat/completions` request body.
- Parse `message.images[]` / `delta.images[]` (data URLs) from response into `Message.Images`.

**Cross-cutting:**
- Provider-agnostic media persistence: write base64 to workspace, replace inline bytes with `MediaRef`, idempotent on sha256.
- `ProviderCapabilities.ImageGeneration` flag; `CodexProvider` sets `true`.
- Agent config toggle (`AllowImageGeneration`, default on for Codex-routed agents) + per-request opt-out header.
- Web UI: composer toggle chip, streaming placeholder, `MediaGallery` render verification.

## Out of scope

- DALL-E API-key mode
- Video generation
- Separate `/v1/images/generations` endpoint

## Wire format evidence

Implementation is evidence-backed against the native ChatGPT Responses API event shape (pre-translation), not the compat shim shape. Internal research notes kept in `plans/reports/`.

## Status

In progress on branch `kai/feat/chatgpt-image-gen`. Draft PR to follow.

PR #1349, opened 2026-07-04
fix(providers): unbounded SSE line reader to stop image_generation "token too long" crash

### LINKED ISSUE #1348: image_generation crashes with "bufio.Scanner: token too long" (SSE line exceeds fixed 1MB cap)
## Summary

Any agent using a Codex provider with native `image_generation` fails the turn right after the model finishes generating an image:

```
iter 0 think: llm call: openai-codex: stream read error: bufio.Scanner: token too long
```

The image is actually generated (visible server-side / in provider logs), but the turn errors out before the response reaches the agent — the user gets nothing.

## Environment

- Provider: `openai-codex` (ChatGPT OAuth or native), model with native `image_generation` support (e.g. `gpt-5.4`/`gpt-5.5`)
- GoClaw `dev`
- Any request that produces an image roughly ≥ 1MB as PNG/JPEG (i.e. almost any real image, not a tiny thumbnail)

## Root cause

`internal/providers/sse_reader.go`'s `NewSSEScanner` reads the SSE stream with `bufio.Scanner`, bounded by a fixed max token size:

```go
// internal/providers/defaults.go
SSEScanBufMax = 1024 * 1024 // 1MB max line for large tool call / thinking chunks
```

Codex's native `image_generation` tool streams the **entire base64-encoded image in a single SSE `data:` line** — both the intermediate `response.image_generation_call.partial_image` events (`partial_image_b64`) and the final `image_generation_call` item (`Result`). A 1024×1024 PNG/JPEG encoded as base64 is typically 1.3–4MB, comfortably over the 1MB cap, so `bufio.Scanner.Scan()` fails with `bufio.Scanner: token too long` on essentially every non-trivial generated image.

`SSEScanBufMax` (1MB) was set by #72 to fix an earlier "token too long" issue for large tool-call/thinking chunks — before native `image_generation` (#1000, #1002) existed. The buffer was never revisited when image streaming was added, so a full image line now regularly exceeds it.

## Suggested fix

Replace `bufio.Scanner` with `bufio.Reader.ReadString('\n')` in `NewSSEScanner`, which grows to fit a line of any length instead of erroring past a fixed cap — matching how large-message stdio readers already work in this codebase (`StdioScanBufMax` = 10MB for Claude CLI/ACP), except here there's no reasonable fixed upper bound to pick since generated image size is unbounded.

PR to follow.

Buns0705:
Fixes #1348.

## Summary

Native `image_generation` streams the entire base64-encoded image in a **single SSE data line** (`response.image_generation_call.partial_image`'s `partial_image_b64`, and the final `image_generation_call` item's `Result`). `internal/providers/sse_reader.go` read the stream with `bufio.Scanner`, capped at a fixed `SSEScanBufMax` of 1MB (`internal/providers/defaults.go`). Any generated image whose base64 line exceeds 1MB — essentially any real image, not a tiny thumbnail — makes `Scanner.Scan()` fail with:

```
stream read error: bufio.Scanner: token too long
```

killing the turn after the model already generated the image.

## Root cause

`SSEScanBufMax` was introduced by #72 to fix an earlier "token too long" case for large tool-call/thinking chunks, before native `image_generation` (#1000, #1002) existed. The buffer was never revisited when image streaming was added.

## Fix

Replace `bufio.Scanner` with `bufio.Reader.ReadString('\n')` in `NewSSEScanner`, which grows to fit a line of any length instead of erroring past a fixed cap. There's no reasonable fixed upper bound to pick here since generated image size is unbounded (unlike the 10MB cap used for stdio/JSONRPC readers, which are bounded protocol messages).

- `SSEScanBufMax` is now unused and removed from `internal/providers/defaults.go`.
- `SSEScanBufInit` is kept and reused as the reader's initial buffer size (still avoids reallocation churn for typical small lines).
- `[DONE]` termination and `event:`/`data:` parsing behavior are unchanged.
- Handles a final line delivered together with `io.EOF` and no trailing newline (previously relied on `bufio.Scanner`'s built-in handling of this case).

## Tests

Added `internal/providers/sse_reader_bigline_test.go`:
- `TestSSEScannerHandlesMultiMBLine` — a 3MB single data line (well over the old 1MB cap) reads correctly end-to-end.
- `TestSSEScannerFinalLineNoNewline` — a final line with no trailing `\n` (EOF-terminated) is still delivered.

`go build ./internal/providers/`, `go vet ./internal/providers/`, and `go test ./internal/providers/` all pass (go 1.26, verified against a clean `dev` checkout).

## Notes for reviewers

Scoped to `NewSSEScanner`/`SSEScanner.Next()` only — `event:`/`data:` field parsing, `[DONE]` termination, and callers (OpenAI, Anthropic, Codex) are unchanged.

### DIFF
--- a/internal/providers/defaults.go
+++ b/internal/providers/defaults.go
@@ -11,9 +11,10 @@ const (
 	// ctx cancellation from unblocking bufio.Scanner. Use NewDefaultHTTPClient() instead.
 	DefaultHTTPTimeout = 300 * time.Second
 
-	// SSE stream scanner buffer sizes (OpenAI-compat, Anthropic, Codex).
-	SSEScanBufInit = 64 * 1024   // 64KB initial buffer
-	SSEScanBufMax  = 1024 * 1024 // 1MB max line for large tool call / thinking chunks
+	// SSE stream reader initial buffer size (OpenAI-compat, Anthropic, Codex).
+	// No max: bufio.Reader.ReadString grows to fit a line of any length — a full
+	// base64 image in a single image-generation SSE data line can exceed several MB.
+	SSEScanBufInit = 64 * 1024 // 64KB initial buffer
 
 	// Stdio/JSONRPC scanner buffer sizes (Claude CLI, ACP).
 	StdioScanBufInit = 256 * 1024       // 256KB initial buffer
--- a/internal/providers/sse_reader.go
+++ b/internal/providers/sse_reader.go
@@ -11,55 +11,76 @@ import (
 // SSEScanner reads an SSE (Server-Sent Events) stream line by line,
 // extracting data payloads. Shared by OpenAI, Anthropic, and Codex providers.
 type SSEScanner struct {
-	scanner   *bufio.Scanner
+	reader    *bufio.Reader
 	data      string
 	eventType string
 	err       error
+	done      bool
 }
 
-// NewSSEScanner creates an SSE scanner with appropriate buffer sizes.
+// NewSSEScanner creates an SSE scanner.
+//
+// It reads lines with bufio.Reader.ReadString, which grows to fit a line of
+// any length, instead of bufio.Scanner, whose fixed max-token size (formerly
+// SSEScanBufMax) overflowed with "token too long" on a single data line
+// carrying a full base64 image — image generation streams multi-MB frames in
+// a single SSE line. SSEScanBufInit is reused as the reader's initial size.
 func NewSSEScanner(r io.Reader) *SSEScanner {
-	sc := bufio.NewScanner(r)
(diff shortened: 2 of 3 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. bufio.Scanner's fixed 1MB cap was a ticking time bomb once image_generation started streaming multi-MB base64 in a single SSE line. bufio.Reader.ReadString is the right replacement: grows to fit, no arbitrary ceiling needed, and SSEScanBufInit is preserved for the initial allocation. [DONE] termination and event/data parsing are unchanged. Tests cover both the 3MB case and the EOF-without-newline edge case. Ship it.

PR #1353, opened 2026-07-04
fix(zalo): correctly forward inbound reference images to the agent

### LINKED ISSUE #1352: Zalo Personal: photo sent as reference image is not usable by the agent (wrong classification, dropped caption, quoted photo never downloaded)
## Summary

On the `zalo_personal` channel, sending a photo to the agent as a reference (e.g. "make a poster from this jewelry photo") produces an incorrect result — the agent's output ignores the actual product in the photo. This happens in **both** common patterns:

1. Sending the photo with a caption in the same message.
2. Sending the photo first, then replying to it with the instruction (quote).

Telegram works correctly for the same prompts and the same underlying model — this is specific to the `zalo_personal` channel's inbound media handling.

## Root cause — three separate bugs

**1. Downloaded images are classified as documents, not images**

`internal/channels/zalo/personal/content-media.go`'s `downloadFile` saves the attachment to a temp file using the extension derived from the CDN URL. Zalo CDN URLs frequently carry no usable file extension, so the file lands as `.bin`.

Downstream, `internal/agent/media.go`'s `persistMedia` infers the MIME type **purely from the file extension** (`mimeFromExt`). A `.bin` file is classified `application/octet-stream` → kind `"document"` → the image is silently **not** attached as vision/image-generation input.

This is misleading in traces: `extractContentAndMedia` still emits a `<media:image>` tag (it force-detects the kind via `att.IsImage()`, independent of the actual downloaded file), so the trace input *looks* correct while the pipeline actually treats the file as a document.

**2. The photo's caption is dropped**

`extractContentAndMedia` returns only the `<media:image>` tag; the caption (Zalo puts it in the attachment's `title`) is discarded entirely. The model sees an image with no instruction attached to it.

**3. A reply to a photo never downloads the photo**

`protocol.TQuote.Attach` (`QuoteAttachment`) carries the full quoted attachment payload including its URL, but `content-media.go`/`handlers.go` only ever rendered it as placeholder text (`[Quoted image]` / `[Quoted non-text message]` via `AttachmentText()`) — the quoted image was never downloaded or attached as media. So "reply to this photo with a new instruction" had **no image at all**, not even a misclassified one.

## Suggested fix

1. Sniff the first bytes of the downloaded file (`http.DetectContentType`) and rename to the correct image extension when it disagrees with the URL-derived one.
2. Append the caption (attachment title/description) after the `<media:image>` tag instead of discarding it.
3. Add `QuoteAttachment.ParseAttachment()` (decode the raw payload into an `Attachment`) and download the quoted image when replying, wiring it into both `handleDM` and `handleGroupMessage`.

PR to follow.

Buns0705:
Fixes #1352.

## Summary

Three separate bugs combined to make `zalo_personal` unable to use a photo as a vision/image-generation reference — even in traces where the `<media:image>` tag appears present and correct.

## Root cause & fix

**1. Wrong classification (`.bin` → treated as document, not image)**

Zalo CDN URLs frequently have no usable file extension, so `downloadFile` in `internal/channels/zalo/personal/content-media.go` saved images as `.bin`. `internal/agent/media.go`'s `persistMedia` infers MIME purely from the file extension, so a `.bin` image was silently classified as a document and never attached as vision input — while `extractContentAndMedia`'s own `<media:image>` tag (which force-detects the kind independently via `att.IsImage()`) still claimed it was an image, making the trace look correct when it wasn't.

→ Added `fixDownloadedExt`: sniffs the first 512 bytes with `http.DetectContentType` and renames the file to the correct extension (`.jpg`/`.png`/`.gif`/`.webp`) when it disagrees with the URL-derived one. No-ops for non-images and for files whose extension is already correct.

**2. Caption dropped**

`extractContentAndMedia` returned only the `<media:image>` tag; the caption (Zalo attachment `title`, falling back to `description`) was discarded, so the model saw an image with no attached instruction.

→ Append the caption after the media tag for image attachments.

**3. Quoted photo never downloaded**

Replying to a photo with a new instruction only rendered `[Quoted image]`/`[Quoted non-text message]` placeholder text. `TQuote.Attach` (`QuoteAttachment`) carries the full quoted attachment payload including its URL, but nothing ever downloaded it.

→ Added `QuoteAttachment.ParseAttachment()` and `extractQuoteMedia()`, wired into `handleDM`, `handleGroupMessage`, and `extractContentAndMediaWithQuote`.

## Tests

- `internal/channels/zalo/personal/content_media_ext_test.go`: `fixDownloadedExt` renames a PNG-sniffed `.bin` to `.png`, leaves non-image content untouched, and leaves already-correct extensions untouched.
- `internal/channels/zalo/personal/protocol/message_quote_parse_test.go`: `QuoteAttachment.ParseAttachment()` decodes an image quote's URL, and returns `nil` for empty/non-object payloads.

`go build`, `go vet`, and the full `internal/channels/zalo/...` test suite pass (go 1.26, verified against a clean `dev` checkout).

## Notes for reviewers

- Live-verified end-to-end against production Zalo traffic: direct image+caption, quoted image + instruction, and multi-turn image edits (referencing a just-generated image) all correctly carry the reference photo after this fix.
- `downloadFile`'s network call and the caption-append branch aren't covered by a network-free unit test (would need a local SSRF-exempt test server); the two extracted, network-free helper functions (`fixDownloadedExt`, `ParseAttachment`) that do the actual classification/decoding work are fully covered instead.

### DIFF
--- a/internal/channels/zalo/personal/content-media.go
+++ b/internal/channels/zalo/personal/content-media.go
@@ -47,6 +47,17 @@ func extractContentAndMedia(content protocol.Content) (string, []string) {
 		ContentType: mimeType,
 		FileName:    att.Title,
 	}})
+	// Zalo photo messages carry the user's caption in the attachment title.
+	// Keep it next to the media tag — dropping it loses the actual request text.
+	if att.IsImage() {
+		caption := strings.TrimSpace(att.Title)
+		if caption == "" {
+			caption = strings.TrimSpace(att.Description)
+		}
+		if caption != "" {
+			tag = tag + "\n" + caption
+		}
+	}
 	return tag, []string{filePath}
 }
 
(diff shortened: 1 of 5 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix for 3 compounding Zalo personal image-reference bugs.

1. **fixDownloadedExt**: correct approach — http.DetectContentType on first 512 bytes, rename when extension disagrees with sniffed content. No-ops for non-images and already-correct extensions. Good edge case handling.
2. **Caption append**: simple and correct — preserves the user's actual instruction text alongside the media tag.
3. **extractQuoteMedia + ParseAttachment**: clean extraction of quoted photo download. Proper nil/empty guards throughout.

Tests cover the two network-free helper functions that do the actual classification/decoding work. Live verification against production Zalo traffic noted in PR body.

No security concerns, no breaking changes, well-documented.

PR #1355, opened 2026-07-04
feat(providers): add provider-level thinking/reasoning toggle for Ollama

bclermont:
Ollama has two separate request-building code paths: OllamaProvider (native /api/chat, used for num_ctx control) and OpenAIProvider (OpenAI-compat /v1/chat/completions). An earlier fix disabled thinking mode by hardcoding think=false, but only in the OpenAI-compat path -- OllamaProvider.buildRequest() never set the think field at all, so reasoning-capable models (qwq, deepseek-r1) defaulted to visible chain-of-thought reasoning regardless of that fix. Confirmed live via a docker-engineer agent streaming full reasoning traces despite the existing disable.

Replaced the hardcoded always-off behavior with a provider-level tri-state setting (llm_providers.settings.thinking_enabled: unset = default off, explicit true/false overrides), configurable via the provider's Advanced settings dialog. Both OllamaProvider.buildRequest() and OpenAIProvider.buildRequestBody() now read and respect this same setting, so the toggle works regardless of which Ollama code path a given deployment routes through.

Added tests for setting parsing (unset/true/false/malformed) and both provider request-builders' handling of the override.

### DIFF
--- a/cmd/gateway_providers.go
+++ b/cmd/gateway_providers.go
@@ -317,7 +317,8 @@ func registerProvidersFromDB(registry *providers.Registry, provStore store.Provi
 				host = "http://localhost:11434"
 			}
 			numCtx := resolveOllamaNumCtx(&p, config.DockerLocalhost(host), "")
-			prov := providers.NewOllamaProvider(p.Name, config.DockerLocalhost(host), "llama3.3", numCtx, nil)
+			prov := providers.NewOllamaProvider(p.Name, config.DockerLocalhost(host), "llama3.3", numCtx, nil).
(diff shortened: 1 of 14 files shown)


clark-cant (approved):
LGTM — clean, well-scoped feature. Provider-level `thinking_enabled` for Ollama is the right abstraction: nil preserves existing default-off behavior, explicit true/false overrides it. Builder pattern mirrors `WithOllamaNumCtx` consistently. Precedence is correct (provider-level > OptThinkingLevel > default). Both native Ollama and OpenAI-compat paths covered. Tests cover default, override, precedence, and accessor round-trip. UI properly scoped to Ollama provider type with complete i18n for all 4 locales. No breaking changes.

PR #1358, opened 2026-07-05
fix(migrate): use iofs source so migrations load on Windows

thotam:
## Summary

`goclaw migrate` (up/down/version/…) is unusable on Windows: every subcommand fails before touching the DB with

```
create migrator: failed to open source, "file:///D:/Work/goclaw/migrations":
open .: The filename, directory name, or volume label syntax is incorrect.
```

golang-migrate's `file` source driver mis-parses absolute drive-letter `file://` URLs. This PR switches the migration source from a `file://` URL to an `iofs` source over `os.DirFS`, which uses native OS path handling and works identically on every platform.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev`

## Root cause & fix

- `newMigrator` built a `file://` URL via `migrationsSourceURL`/`absoluteToFileURI` and passed it to `migrate.New`. On Windows the resulting `file:///D:/…` URL is rejected by golang-migrate's file driver (`open .` error), so migrations never load. The Linux CI path (relative/POSIX) hid this.
- Fix: `newMigrationSource()` now returns an `iofs` source built from `os.DirFS(resolveMigrationsDir())`, and `newMigrator` uses `migrate.NewWithSourceInstance("iofs", …)`. The fragile drive-letter URL helpers (`absoluteToFileURI`, `migrationsSourceURL`) and their URL-shape unit tests are removed.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes (changed package clean)
- [x] Tests pass — new DB-free regression test; one unrelated `cmd` test (`TestShellDenyGroupsConfigReload_…`) fails on this Windows box on clean `dev` too, because it needs a `claude-test` binary on PATH — not touched here. ⚠️ `-race` not run locally (Windows env has CGO disabled); CI runs it on Linux.
- [x] Web UI builds — N/A, no UI changes
- [x] SQL queries use parameterized placeholders — N/A, no SQL
- [x] New user-facing strings — N/A
- [x] Migration version bumped — N/A, no schema change

## Surface parity
- CLI/runtime only (`cmd/migrate.go`). `newMigrator` signature is unchanged, so the `onboard` and `upgrade` callers are unaffected.
- Gateway / API / Web UI: **N/A**.

## Test Plan

- **Unit** (`cmd/migrate_test.go`): `TestNewMigrationSource_LoadsMigrations` opens the real `migrations/` directory via `newMigrationSource()` and asserts the first version loads — DB-free and cross-platform. A probe confirmed the old `file://` source fails (`open .`) while the `iofs` source loads `first=1` on this machine.
- **End-to-end on Windows**: `goclaw migrate version` → `version: 89, dirty: false`; `goclaw migrate up` → `migration complete version=89` plus data hooks. Both previously failed at source creation.

### DIFF
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -9,11 +9,11 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	"github.com/golang-migrate/migrate/v4"
 	_ "github.com/golang-migrate/migrate/v4/database/postgres"
-	_ "github.com/golang-migrate/migrate/v4/source/file"
+	"github.com/golang-migrate/migrate/v4/source"
+	"github.com/golang-migrate/migrate/v4/source/iofs"
 	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — clean, well-scoped bug fix. Root cause: golang-migrate's file source driver mis-parses Windows absolute drive-letter file:// URLs (file:///D:/... fails with "open ." error). Fix: switch to iofs source over os.DirFS, which uses native OS path handling and works identically on every platform. Removal of absoluteToFileURI and migrationsSourceURL helpers is correct — they were shape-driven workarounds for a broken driver. New newMigrationSource() is simple and testable. migrate.NewWithSourceInstance("iofs", src, dsn) is the right API. newMigrator signature unchanged, so onboard/upgrade callers unaffected. Test is DB-free, cross-platform, and guards against the exact regression. CI all green.

PR #1363, opened 2026-07-05
fix(skills): allow editing SKILL.md content for database-managed skills

bclermont:
### What
  
  Adds the ability to edit a skill's actual SKILL.md content directly from the web UI,
  restricted to database/tenant-managed skills only — bundled/system skills remain
  read-only, both in the UI (disabled button + tooltip) and enforced server-side (403).
  
  ### Backend

  - New `PUT /v1/skills/{id}/files/{path...}` endpoint (`handleWriteFile` in
    `internal/http/skills_versions.go`), gated to non-system skills only.
  - Follows the same versioning convention as the `skill_manage` tool's `patch` action:
    saving creates a new immutable version directory rather than overwriting the current
    one in place. The skill's DB row (`version`, `file_path`, `file_size`, `file_hash`)
    is updated to point at the new version; the previous version's files are left
    untouched.
  - Same path-traversal/symlink/system-artifact guards and ownership checks as the
    existing read endpoint.

  ### Frontend

  - Skill detail dialog's Content tab gets an "Edit content" button, disabled with a
    tooltip for system skills.
  - Editing fetches the **raw** file content (frontmatter intact) rather than the
    stripped preview normally shown, so saving doesn't silently destroy the skill's
    YAML frontmatter.
  - Save/cancel follow the established catch-and-toast error pattern (no uncaught
    promise rejections, matching the earlier config-save-button fix).
  - Fixed two follow-up bugs found during manual testing:
    - **Version display staleness**: the dialog cached version metadata on first load
      and never invalidated it, so the header kept showing the pre-save version number
      until a full page reload. Now invalidates and refetches when the skill's version
      changes.
    - **Scroll regression**: the read-only content preview had `overflow-hidden` on a
      flex child with no explicit height — per the flexbox spec this forces the item's
      minimum size to zero, letting it silently clip ~7000px of content into ~600px of
      visible space instead of overflowing so the scroll container could do its job.
      Removed the offending `overflow-hidden`, confirmed scrollable via measured
      `scrollHeight`/`clientHeight` in a real build.
  
  i18n keys added to all 4 locales (en, ko, vi, zh).

  ### Verified

  - `go test ./internal/http/... -run TestHandleWriteFile` — pass
  - `npx tsc --noEmit`, `npm run build` — clean
  - `npx vitest run` — 313/313 tests pass
  - Manually tested live: content edit + version bump + scroll all confirmed working
    after fixes.

  ### Commits

  - `c745e950` — initial implementation
  - `9190b7fc` — fix version bump behavior + scroll (first pass)
  - `129b68d4` — fix version display refresh + scroll (root-caused correctly this time)

  Want me to push the branch and open this as an actual PR?

### DIFF
--- a/internal/http/skills.go
+++ b/internal/http/skills.go
@@ -94,6 +94,7 @@ func (h *SkillsHandler) RegisterRoutes(mux *http.ServeMux) {
 	mux.HandleFunc("GET /v1/agents/{agentID}/skills", h.authMiddleware(h.handleListAgentSkills))
 	mux.HandleFunc("GET /v1/skills/{id}/versions", h.authMiddleware(h.handleListVersions))
 	mux.HandleFunc("GET /v1/skills/{id}/files/{path...}", h.authMiddleware(h.handleReadFile))
+	mux.HandleFunc("PUT /v1/skills/{id}/files/{path...}", h.adminMiddleware(h.handleWriteFile))
 	mux.HandleFunc("GET /v1/skills/{id}/files", h.authMiddleware(h.handleListFiles))
 	mux.HandleFunc("GET /v1/skills/{id}/evolution", h.authMiddleware(h.handleGetEvolution))
 	mux.HandleFunc("GET /v1/skills/{id}/metrics", h.authMiddleware(h.handleGetSkillMetrics))
--- a/internal/http/skills_versions.go
+++ b/internal/http/skills_versions.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/nextlevelbuilder/goclaw/internal/skills"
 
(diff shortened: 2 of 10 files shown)


clark-cant (approved):
LGTM — clean, well-scoped feature. Backend correctly gates writes to non-system skills only (403 for bundled), with proper path-traversal/symlink/system-artifact guards and ownership checks. Version bump via immutable new directory matches skill_manage patch convention. Tests cover the critical paths: happy path with version bump + original immutability, system skill rejection, path traversal rejection. Frontend fetches raw content (preserving YAML frontmatter), disabled button + tooltip for system skills, and the scroll fix (removing overflow-hidden on unbounded flex child) is correct per flexbox spec. i18n for all 4 locales. No security concerns — adminMiddleware required, ownership enforced for non-admins, no secrets exposed.

PR #1364, opened 2026-07-05
fix(channel-memory): retry truncated passive extraction JSON

namnhfreelancer:
## Summary

Passive channel memory extraction parsed the background provider response as strict JSON immediately after the LLM call. In production this could fail intermittently with:

```text
parse extraction JSON: unexpected end of JSON input
```

This PR makes the extraction path resilient to incomplete provider output by retrying once with a smaller input window and a larger output budget before surfacing an error.

## Root Cause

The failure is caused by incomplete or empty JSON returned by the configured background provider, not by the channel message cap itself.

The previous implementation:

- sent up to ~12k chars of redacted channel history
- limited extraction output to `1200` tokens
- parsed `resp.Content` directly with `json.Unmarshal`
- did not inspect `FinishReason == "length"`
- did not retry when the content was empty or ended mid-JSON

Because of that, different message caps (`100`, `50`, `10`) could still fail. A lower cap reduces input size but does not guarantee the model's JSON array finishes before the output limit or provider interruption.

## Changes

- Introduce explicit extraction limits:
  - initial input cap: `12000` chars
  - retry input cap: `6000` chars
  - output budget: `4096` tokens
- Split provider invocation into `callExtractionProvider()` so the first call and retry share request construction and usage-cap metadata.
- Retry once when:
  - the provider returns `FinishReason == "length"`
  - the response content is empty
  - JSON parsing fails with `unexpected end of JSON input`
- Return a clearer error if the retry is also truncated:

```text
extraction response truncated after retry
```

- Add prompt guidance to return at most 20 highest-confidence items, which bounds output size while preserving durable facts.
- Add unit coverage for:
  - truncated provider output with `FinishReason == "length"`
  - empty provider response
  - incomplete JSON without a length finish reason
  - fenced JSON array parsing

## Behavior Notes

This keeps the public API and review-queue schema unchanged. The fix is limited to the passive channel memory extraction LLM call path.

If the retry succeeds, the run proceeds normally and creates review items as before. If the retry also truncates or returns invalid JSON, the run still fails, but with a clearer root-cause error instead of leaking a low-level JSON EOF as the primary diagnosis.

## Surface Parity

- Gateway/server: affected; passive extraction now retries incomplete LLM JSON output.
- API contract: N/A; no request/response shape changes.
- Web UI: N/A; no UI behavior or copy changes.
- CLI/runtime package: N/A; no command or package contract changes.
- DB/migrations: N/A; no schema changes.

## Verification

- `gofmt -w internal/channelmemory/extractor.go internal/channelmemory/extractor_test.go`
- `go test ./internal/channelmemory`
- `git diff --check`
- GitHub Actions CI: `release-versioning`, `go`, and `web` passed
- Docker Compose rebuild completed successfully and the `goclaw` service started with the rebuilt image

## Related Work

No duplicate GitHub issue or PR was found with a quick search for passive memory JSON EOF failures.

### DIFF
--- a/internal/channelmemory/extractor.go
+++ b/internal/channelmemory/extractor.go
@@ -12,6 +12,13 @@ import (
 	usagecaps "github.com/nextlevelbuilder/goclaw/internal/usage/caps"
 )
 
+const (
+	extractionMaxInputChars        = 12000
+	extractionRetryMaxInputChars   = 6000
+	extractionMaxOutputTokens      = 4096
+	extractionRetryMaxOutputTokens = 4096
+)
+
 type ExtractedItem struct {
 	Type       string   `json:"type"`
 	Summary    string   `json:"summary"`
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — clean resilience fix. Root cause (max_tokens=1200 truncating JSON) correctly addressed with configurable limits, retry on length/empty/EOF, and bounded prompt guidance. Tests cover all retry paths. No concerns.

PR #1366, opened 2026-07-05
fix(tracing): dedupe tool_call span rows in hourly snapshot aggregation

bclermont:
Bug

  Production logs showed intermittent hourly backfill failures:

  level=WARN msg="backfill: aggregate hour failed" hour=2026-07-04T22:00:00-04:00 error="ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time (SQLSTATE 21000)"

  Root cause

  In internal/tracing/snapshot_worker.go:

  - querySpanAggregates joined traces to spans with s.span_type IN ('llm_call', 'tool_call'), grouped by t.agent_id, t.channel, s.provider, s.model.
  - tool_call spans always have provider/model = NULL. All such spans for a given (agent_id, channel) in the hour collapsed via GROUP BY into a single row with provider="", model="" (post-COALESCE).
  - mergeTraceAndSpanRows step 1 already builds a "totals" row per (agent_id, channel) from queryTraceAggregates, also keyed at Provider="", Model="" — the identical key.
  - The upsert conflict target (internal/store/pg/snapshot.go, mirrored in internal/store/sqlitestore/snapshots.go) is (bucket_hour, agent_id, provider, model, channel, tenant_id).
  - Result: a single aggregateHour batch could produce two store.UsageSnapshot rows mapping to the same conflict-target key → Postgres rejects the whole statement with 21000. This only surfaces on hours where an agent/channel has tool_call spans, which
  explains the intermittent nature.

  Fix

  1. querySpanAggregates: narrowed the JOIN to s.span_type = 'llm_call' only (dropped 'tool_call'). Tool-call counts are already captured per-trace via traces.tool_call_count in queryTraceAggregates — tool_call spans never carry meaningful provider/model
  to aggregate separately, so including them in this query only produced a colliding duplicate row.
  2. mergeTraceAndSpanRows: added defense-in-depth — if a span row still resolves to Provider=="" && Model=="" for any other reason, its metrics (LLMCallCount, tokens, cost, cache/thinking tokens) are now merged into the existing totals row for that
  (agent_id, channel) key via a new findTotalsSnapshotIndex helper, instead of being appended as a duplicate row.
  3. Confirmed DB-agnostic: SnapshotWorker operates on *sql.DB directly and is shared by both PostgreSQL and SQLite backends — no parallel SQLite-specific aggregation query exists, so no second fix location needed.

  Test

  Added TestMergeTraceAndSpanRowsDedupesToolCallProviderModelRow in internal/tracing/snapshot_worker_test.go: builds trace rows for one agent/channel plus span rows containing a provider="" model="" row (simulating tool_call-only spans) alongside a
  normal provider="anthropic" model="claude-..." row, calls mergeTraceAndSpanRows, and asserts:
  - exactly one row has Provider=="" && Model=="" for that agent/channel (no duplicate conflict-target keys),
  - its LLMCallCount/tokens/cost equal the sum of both contributing rows.

  Verification

  - go build ./... — pass
  - go vet ./... — pass
  - go test ./internal/tracing/... — pass

  Files changed

  - internal/tracing/snapshot_worker.go
  - internal/tracing/snapshot_worker_test.go (new test)

  Scope note

  Unrelated to the in-progress feat/mcp-crud-server branch — this is a standalone bugfix off dev.

### DIFF
--- a/internal/tracing/snapshot_worker.go
+++ b/internal/tracing/snapshot_worker.go
@@ -340,7 +340,7 @@ func querySpanAggregates(ctx context.Context, db *sql.DB, from, to time.Time) ([
 			COALESCE(t.channel, '') as channel,
 			COALESCE(s.provider, '') as provider,
 			COALESCE(s.model, '') as model,
-			COUNT(*) FILTER (WHERE s.span_type = 'llm_call') as llm_call_count,
+			COUNT(*) as llm_call_count,
 			COALESCE(SUM(s.input_tokens), 0) as span_input_tokens,
 			COALESCE(SUM(s.output_tokens), 0) as span_output_tokens,
 			COALESCE(SUM(s.total_cost), 0) as span_cost,
@@ -348,7 +348,7 @@ func querySpanAggregates(ctx context.Context, db *sql.DB, from, to time.Time) ([
 			COALESCE(SUM(CAST(s.metadata->>'cache_creation_tokens' AS INTEGER)), 0) as cache_create_tokens,
 			COALESCE(SUM(CAST(s.metadata->>'thinking_tokens' AS INTEGER)), 0) as thinking_tokens
 		FROM traces t
-		JOIN spans s ON s.trace_id = t.id AND s.span_type IN ('llm_call', 'tool_call')
+		JOIN spans s ON s.trace_id = t.id AND s.span_type = 'llm_call'
 		WHERE t.start_time >= $1 AND t.start_time < $2
 		  AND t.parent_trace_id IS NULL
 		GROUP BY t.agent_id, t.channel, s.provider, s.model`, from, to)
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — clean, well-scoped bug fix.

**Root cause**: `querySpanAggregates` joined `spans` with `span_type IN ('llm_call', 'tool_call')`. Tool_call spans always have NULL provider/model → COALESCE to empty strings → collides with the totals row from `queryTraceAggregates` keyed at (agent_id, channel, provider='', model=''). Single upsert batch produces two rows mapping to the same conflict-target key → SQLSTATE 21000.

**Fix A**: Narrowed `querySpanAggregates` JOIN to `span_type = 'llm_call'` only. Correct — tool_call spans never carry meaningful provider/model; tool call counts already captured via `traces.tool_call_count` in `queryTraceAggregates`.

**Fix B**: Defense-in-depth in `mergeTraceAndSpanRows` — if a span row still resolves to Provider=="" && Model=="" for any other reason, merge metrics into existing totals row via `findTotalsSnapshotIndex` helper instead of appending duplicate row.

**Test**: `TestMergeTraceAndSpanRowsDedupesToolCallProviderModelRow` covers the exact regression — empty provider/model row merged into totals, distinct provider/model detail row preserved.

DB-agnostic (shared `*sql.DB` path for PG and SQLite). No security concerns.

PR #1368, opened 2026-07-05
fix(skills): allow archived skills to be found by slug/name lookup, matching UUID lookup and ListSkills behavior

bclermont:
Bug

  GetSkill(ctx, name)'s slug and name lookup branches filtered status = 'active' only. Any archived-but-enabled skill (e.g. in this environment: docx, pdf, pptx, skill-creator, xlsx) always 404'd when looked up by name/slug — even with correct tenant
  context — while lookup by UUID worked fine, and ListSkills() already included archived skills.

  Root cause

  Three places disagreed on which skills a caller should be able to see:

  - GetSkill UUID branch: status IN ('active', 'archived')
  - GetSkill slug/name branches: status = 'active' only ← the bug
  - ListSkills(): status IN ('active', 'archived'), with an existing code comment noting archived skills are "shown dimmed in the UI so admins can see missing deps and re-activate after installing them"

  Per docs/15-core-skills-system.md, "archived" means a system skill is enabled but has missing dependencies — a dependency-check outcome, not a user-intent "hide this" state. Visibility is already gated separately by the Enabled flag (FilterSkills), not
  by status. So filtering out archived skills specifically on the name/slug path was an inconsistency, not intentional behavior — confirmed by the docs, the ListSkills comment, and the UUID branch's own inclusive filter.

  There was even a pre-existing test asserting the buggy behavior (TestSQLiteSkillStore_GetSkill_UUIDCanReadArchivedSlugStaysActiveOnly), which is replaced by this PR.

  Fix

  Changed the slug and name WHERE clauses in both store backends to status IN ('active', 'archived'), matching the UUID branch and ListSkills():
  - internal/store/pg/skills_content.go
  - internal/store/sqlitestore/skills_content.go

  (Dual-DB fix, per repo convention — PostgreSQL and SQLite kept in sync.)

  Test

  Replaced the old (bug-asserting) test with TestSQLiteSkillStore_GetSkill_ArchivedFoundByUUIDAndSlugAndName in internal/store/sqlitestore/skills_test.go: creates an archived skill and asserts it's found via UUID, slug, and name lookup, all returning
  Status == "archived".

  Verification

  - go build ./... — pass
  - go vet ./... — pass
  - go test ./internal/skills/... — pass
  - go test -tags sqlite ./internal/store/sqlitestore/... -run GetSkill -v — pass
  - PG-side change is a direct mirror of the verified SQLite fix; no live PG integration DB was available in this environment to additionally exercise it there.

  Files changed

  - internal/store/pg/skills_content.go
  - internal/store/sqlitestore/skills_content.go
  - internal/store/sqlitestore/skills_test.go

### DIFF
--- a/internal/store/pg/skills_content.go
+++ b/internal/store/pg/skills_content.go
@@ -135,13 +135,17 @@ func (s *PGSkillStore) GetSkill(ctx context.Context, name string) (*store.SkillI
 		return &enriched[0], true
 	}
 
+	// "archived" means a system skill has missing dependencies (see seeder.go); it does not
+	// mean the skill is hidden from lookup. ListSkills() and GetSkillByID() both include
+	// active + archived skills, so name/slug lookup must match to avoid 404s on archived
+	// skills that are otherwise still enabled.
 	if id, err := uuid.Parse(name); err == nil {
 		return scan(baseSelect+"id = $"+fmt.Sprint(len(args)+1)+" AND status IN ('active', 'archived')"+scope, append(args, id)...)
 	}
-	if info, ok := scan(baseSelect+"slug = $"+fmt.Sprint(len(args)+1)+" AND status = 'active'"+scope, append(args, name)...); ok {
+	if info, ok := scan(baseSelect+"slug = $"+fmt.Sprint(len(args)+1)+" AND status IN ('active', 'archived')"+scope, append(args, name)...); ok {
 		return info, true
 	}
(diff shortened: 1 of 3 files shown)


clark-cant (approved):
LGTM — clean, well-scoped consistency fix. Three lookup paths disagreed on archived skill visibility: UUID branch included 'archived', slug/name branches didn't. Fix aligns all three to status IN ('active', 'archived'), matching ListSkills() behavior and docs/15-core-skills-system.md semantics (archived = missing deps, not hidden). Dual-DB fix keeps PG and SQLite in sync. Test correctly replaces the old bug-asserting test with proper regression coverage for all three lookup paths. No security concerns.

PR #1369, opened 2026-07-05
fix(mcp): resolve tool descriptions/schemas consistently across all MCP tool-listing paths

bclermont:
Bug

  MCP tool descriptions and input parameter schemas were showing up empty (description: '', parameters: {"type": "object"}) in multiple places that build the agent's function-calling prompt — both in goclaw_prompt <agent> output and the web UI's
  system-prompt preview — while an external MCP inspector confirmed the raw tools/list response always had the correct, full schema. This affected goclaw's own newly-added MCP CRUD server disproportionately, but the underlying bugs were general, not
  specific to that server.

  Three distinct, compounding bugs were found and fixed:

  1. Admin "browse server tools" endpoint returned no descriptions once connected

  internal/http/mcp_tools.go's handleListServerTools had two code paths depending on connection state: a not-yet-connected server used DiscoverTools (bare names + real descriptions, correct); an already-connected server used ServerToolNames (prefixed
  names, no description at all — the old code even commented "names only, no descriptions available"). Whichever shape was showing when an admin configured an agent's tool grants got persisted verbatim into tool_allow.

  Fix: added Manager.ServerToolInfos, which resolves registered (prefixed) names back to the live *BridgeTool and returns the bare name + real description — same shape as DiscoverTools, both paths now consistent.

  2. Runtime prompt-building ignored legacy-persisted grant data

  ListToolsForAgent's explicit-tool_allow branch looked up toolCache[toolName] using the raw stored value, with no registry fallback. Any grant saved while its server was already connected (bug #1) had the tool_allow persisted with the prefixed name —
  but toolCache is keyed by bare names, so every lookup silently missed, producing empty description/parameters for all granted tools, regardless of which MCP server.

  Fix: added bareMCPToolName() (normalizes legacy-prefixed and current bare stored names to the same key) and resolveMCPToolInfo() (resolves from the live registry first, falls back to the cached settings snapshot). Both the explicit-allow and
  empty-allow branches now share this one resolution path. Existing already-saved agent grants resolve correctly immediately — no manual re-save required.

  3. No live data yet for freshly-added/rebuilt servers

  ListToolsForAgent is "connection-free by design" — it only reads from the in-memory registry (populated by real chat turns) or a persisted cache (populated after the first live connect via a real chat turn). For a server that's never had an actual chat
  turn run against it yet (e.g. right after adding/rebuilding), both are empty, so bugs #1/#2's fixes have nothing to normalize or find — it correctly fell through to an empty placeholder. A clean rebuild alone doesn't retroactively populate this data.

  Fix: when both the registry and cache are empty, ListToolsForAgent now performs one bounded (5s), best-effort on-demand discovery — reusing the existing credential-resolution path (resolveServerCredentials) and existing
  discoverRawTools/buildCachedToolInfo helpers — uses the result immediately for that response, and persists it asynchronously (same fire-and-forget pattern already used elsewhere) so subsequent calls hit the cache. Failure (unreachable server, no
  credentials yet) falls through to the pre-existing placeholder — no regression for legitimately-unreachable servers.

  Tests

  - internal/mcp/self_connect_schema_test.go → replaced with internal/mcp/server_tool_infos_test.go: builds a minimal generic MCP server, self-connects, verifies description/schema survive through buildCachedToolInfo/ListToolsForAgent, independent of any
  specific server implementation.
  - internal/mcp/list_tools_for_agent_test.go → TestListToolsForAgent_LegacyPrefixedToolAllowResolvesLiveDescriptions: reproduces an agent grant with old prefixed-format tool_allow entries and confirms full descriptions/schemas resolve.
  - On-demand discovery fallback: existing tests (TestListToolsForAgent_EmptyToolAllowUsesCache etc.) continue to pass unchanged — servers with empty transport config fail fast via ValidateServerConfig and fall through exactly as before.

  Verification

  - go build ./... — pass
  - go vet ./... — pass
  - go test ./internal/mcp/... ./internal/http/... ./internal/agent/... — pass

  Files changed

  - internal/http/mcp_tools.go, internal/http/mcp.go
  - internal/mcp/manager.go, internal/mcp/manager_tools.go, internal/mcp/manager_connect.go
  - internal/mcp/server_tool_infos_test.go (new), internal/mcp/list_tools_for_agent_test.go (new)

  Note

  This is a pre-existing regression, not something introduced by the new MCP CRUD server — it was just disproportionately exposed by it, since a self-connecting server is far more likely to hit the "already connected" / "never had a chat turn yet" edge
  cases than a rarely-touched external server.

### DIFF
--- a/internal/http/mcp.go
+++ b/internal/http/mcp.go
@@ -18,9 +18,14 @@ import (
 	"github.com/nextlevelbuilder/goclaw/pkg/protocol"
 )
 
-// MCPToolLister returns discovered tool names for a specific MCP server.
+// MCPToolLister returns discovered tool names/info for a specific MCP server.
 type MCPToolLister interface {
 	ServerToolNames(serverName string) []string
+	// ServerToolInfos returns the original (bare, unprefixed) tool name and
+	// real description for each tool of an already-connected server. Bare
+	// names match the shape returned by mcp.DiscoverTools, so callers don't
+	// have to special-case "already connected" vs. "on-demand discovery".
+	ServerToolInfos(serverName string) []mcp.ToolInfo
 }
 
 // MCPPoolEvictor evicts pooled connections for a tenant+server (called on credential rotation).
--- a/internal/http/mcp_tools.go
+++ b/internal/http/mcp_tools.go
@@ -89,14 +89,19 @@ func (h *MCPHandler) handleListServerTools(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	// Try runtime Manager first — returns names only (no descriptions available).
+	// Try runtime Manager first — returns the original (bare, unprefixed) MCP
+	// tool names with their real descriptions, sourced from the live
+	// *BridgeTool for each registered tool. This must match the bare-name
+	// shape returned by the DiscoverTools fallback below: tool_allow grants
+	// saved from whichever shape this endpoint happens to return are matched
+	// against bare names by ListToolsForAgent's tool_cache lookups
+	// (internal/mcp/manager.go), so returning the registered/prefixed name
+	// here (as before) silently broke prompt-preview schemas for any server
+	// that was already live-connected when its grants were configured.
 	var tools []mcpbridge.ToolInfo
 	if h.mgr != nil {
-		if names := h.mgr.ServerToolNames(srv.Name); len(names) > 0 {
-			tools = make([]mcpbridge.ToolInfo, len(names))
-			for i, n := range names {
-				tools[i] = mcpbridge.ToolInfo{Name: n}
-			}
+		if infos := h.mgr.ServerToolInfos(srv.Name); len(infos) > 0 {
+			tools = infos
 		}
 	}
 
--- a/internal/mcp/list_tools_for_agent_test.go
+++ b/internal/mcp/list_tools_for_agent_test.go
@@ -3,9 +3,12 @@ package mcp
 import (
 	"encoding/json"
 	"sort"
+	"sync/atomic"
 	"testing"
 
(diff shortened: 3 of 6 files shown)


clark-cant (approved):
LGTM — clean, well-scoped bug fix addressing 3 compounding MCP tool description/schema resolution bugs.

**Bug 1** (admin browse endpoint): handleListServerTools used ServerToolNames for already-connected servers → prefixed names with no descriptions. Fix: new ServerToolInfos method returns bare names + real descriptions from live BridgeTool instances. Both code paths now consistent with DiscoverTools shape.

**Bug 2** (runtime prompt): ListToolsForAgent looked up toolCache by raw stored name, but legacy grants had prefixed names → silent misses → empty description/parameters for ALL granted tools. Fix: bareMCPToolName() normalizes legacy prefixed entries to bare names before matching. resolveMCPToolInfo() prefers live registry over stale cache. Both explicit-allow and empty-allow branches share one resolution path. Existing saved grants resolve correctly without manual re-save.

**Bug 3** (no data for fresh servers): ListToolsForAgent was connection-free by design, but servers that never had a chat turn had empty registry+cache → permanently empty preview. Fix: bounded (5s) on-demand discovery when both are empty, reusing existing credential/discovery helpers, async persist so subsequent calls hit cache. Failure falls through to existing placeholder — no regression.

Tests cover: self-connect schema propagation (server_tool_infos_test.go), legacy prefixed tool_allow resolution (list_tools_for_agent_test.go). Both well-constructed regression tests.

No security concerns — purely metadata resolution. All CI passing. Co-authored with Claude Sonnet 5 but code quality is high.

PR #1375, opened 2026-07-06
fix(mcp): refresh per-user credentials without restart

daotri:
## Summary
- add user-scoped MCP pool eviction when per-user credentials are updated or deleted
- include per-user MCP tool definitions in the provider tool list so granted user tools are visible to the model
- return an empty credential status instead of panicking when a user has no MCP credential row
- keep the credential dialog open after save/delete and refresh the selected user's status only

## Tests
- `go test ./internal/http ./internal/store/pg ./internal/mcp ./internal/agent`
- `pnpm --dir ui/web build`

### DIFF
--- a/cmd/gateway_http_wiring.go
+++ b/cmd/gateway_http_wiring.go
@@ -64,6 +64,9 @@ func (d *gatewayDeps) wireHTTPHandlersOnServer(
 		d.server.SetMCPHandler(h.mcp)
 	}
 	if h.mcpUserCreds != nil {
+		if mcpPool != nil {
+			h.mcpUserCreds.SetPoolEvictor(mcpPool)
+		}
 		d.server.SetMCPUserCredentialsHandler(h.mcpUserCreds)
 	}
 	if h.channelInstances != nil {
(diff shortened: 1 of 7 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix for per-user MCP credential refresh without restart.

**Review summary:**
1. Pool eviction: `EvictUser()` properly closes user-scoped connections on credential rotation, mirrors existing `Evict()` pattern. Mutex-held cleanup with semaphore release is correct.
2. Tool visibility: `makeBuildFilteredTools` now includes per-user MCP tool definitions with proper dedup. The `state.Iteration != maxIter` guard avoids injecting tools on the final iteration (correct).
3. Empty credential status: returns structured response with `has_credentials: false` instead of panicking on nil creds — good defensive fix.
4. UI: keeps dialog open after save/delete and refreshes status — better UX than closing on success.
5. Tests: comprehensive coverage of get-missing, set-eviction, and delete-eviction paths.
6. Wiring: `SetPoolEvictor` pattern avoids circular imports, nil-safe.

No security concerns — credential rotation with pool eviction is the correct behavior.

Issue #1374, opened 2026-07-06
claude_cli: TodoRead/NotebookRead in cliNativeToolsAlwaysBlocked no longer exist in current CLI (stderr warnings on every run)


taminhtoan2601:
## Summary
`cliNativeToolsAlwaysBlocked` (`internal/providers/claude_cli_session.go:35`) still lists `TodoRead` and `NotebookRead`, which recent Claude CLI releases (`@anthropic-ai/claude-code` ≥ ~2.x) no longer register. Every CLI invocation now emits:

```
Permission deny rule "NotebookRead" matches no known tool — check for typos.
Permission deny rule "TodoRead" matches no known tool — check for typos.
```

## Where it hurts
Observed on goclaw dev @ af80890e with CLI 2.1.x during background episodic summarization — the stderr warnings show up wrapped in repeated eventbus errors:

```
WARN eventbus: handler error type=session.completed attempt=1
  err="episodic: summarize: claude-cli: signal: killed (stderr: Permission deny rule \"NotebookRead\" matches no known tool — check for typos.\nPermission deny rule \"TodoRead\" matches no known tool — check for typos.\n)"
WARN eventbus: handler error type=session.completed attempt=2 ... (same)
```

To be transparent: we have NOT proven the `signal: killed` itself is caused by the deny-rule warnings (could be a summarize timeout on a slow box — 4 vCPU / 7.6 GB VPS). But the warnings make every such failure noisier and harder to triage, and they fire on every normal run too.

## Suggested fix
Drop `TodoRead`/`NotebookRead` from `cliNativeToolsAlwaysBlocked` (their write-side counterparts `TodoWrite`/`NotebookEdit` still exist and stay blocked), or version-gate the list. Happy to send the one-line PR if you confirm the direction.

## Open question for maintainers
Is episodic summarize expected to spawn a full claude CLI session? On small hosts the repeated `signal: killed` + retry loop silently drops episodic memory — worth a separate look.

clark-cant:
## Maintainer Triage — github-maintain

Valid bug report, @taminhtoan2601. Confirmed by code inspection of `internal/providers/claude_cli_session.go`.

**Root cause confirmed:** `cliNativeToolsAlwaysBlocked` (line ~35) lists `TodoRead` and `NotebookRead`:
```go
var cliNativeToolsAlwaysBlocked = []string{
    "Globs", "Grep", "TodoRead", "TodoWrite", "NotebookRead", "NotebookEdit",
[...]

PR #1376, opened 2026-07-06
fix(claude-cli): drop TodoRead/NotebookRead from always-blocked deny rules

taminhtoan2601:
## Summary
Implements the fix direction confirmed in #1374: remove `TodoRead`/`NotebookRead` from `cliNativeToolsAlwaysBlocked` — current Claude CLI (≥ 2.x) no longer registers them, so the deny rules trigger `Permission deny rule "..." matches no known tool` warnings on every CLI invocation. Write-side counterparts `TodoWrite`/`NotebookEdit` remain blocked.

## Changes
- `internal/providers/claude_cli_session.go`: drop the two stale names + comment explaining why (so the next CLI rename gets the same treatment).
- `internal/providers/claude_cli_session_test.go`: new `TestDisallowedCLITools_NoStaleToolNames` — asserts stale names absent AND `TodoWrite`/`NotebookEdit`/`Glob`/`Grep` still blocked (fail-closed path, `allowedToolNames == nil`).

## Testing
- `go build ./...` and `go build -tags sqliteonly ./...` — clean
- `go vet ./internal/providers/` — clean
- `go test ./internal/providers/` — all pass (`-race` skipped locally: no cgo toolchain on this box; CI will cover it)

## Note
The `signal: killed` retry loop on episodic summarize observed in #1374 is likely a separate resource/timeout concern on small hosts — not addressed here, happy to keep digging in the issue.

### DIFF
--- a/internal/providers/claude_cli_session.go
+++ b/internal/providers/claude_cli_session.go
@@ -31,8 +31,14 @@ var cliNativeToolGoclawEquivalent = map[string]string{
 // cliNativeToolsAlwaysBlocked lists Claude CLI native tools with no GoClaw
 // policy equivalent. They are always disallowed so agent tool policy cannot
 // be bypassed through them.
+//
+// TodoRead and NotebookRead were removed here: current Claude CLI releases
+// (>= 2.x) no longer register them, and a deny rule naming an unknown tool
+// makes the CLI print "Permission deny rule ... matches no known tool" on
+// every invocation. Their write-side counterparts TodoWrite/NotebookEdit
+// still exist in the CLI and remain blocked.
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — clean, well-scoped fix. Removes stale TodoRead/NotebookRead from cliNativeToolsAlwaysBlocked (Claude CLI ≥2.x no longer registers them, causing 'matches no known tool' warnings on every invocation). Write-side counterparts TodoWrite/NotebookEdit correctly remain blocked. Regression test covers both directions: stale names absent AND required names still blocked under fail-closed (nil allowedToolNames). No security concerns — purely removing dead deny rules.

PR #1377, opened 2026-07-06
fix(mcp): derive claude_cli bridge tool surface from agent policy (fixes #1373)

### LINKED ISSUE #1373: claude_cli MCP bridge omits use_skill/delegate while the system prompt requires them (BridgeToolNames drift)
## Summary
The system prompt's Skill loading protocol instructs agents to call `use_skill(...)`, but the Claude CLI MCP bridge never exposes `use_skill` — `BridgeToolNames` (`internal/mcp/bridge_server.go:22`) is a static subset that omits `use_skill`, `delegate`, `datetime`, and `knowledge_graph_search` even when the agent's tool policy allows them. Agents on the `claude_cli` provider follow the protocol, hit a nonexistent tool, and can derail (in our case: fabricate results).

## Environment
- goclaw dev @ af80890e (also verified present on current dev HEAD), Docker `ghcr.io/nextlevelbuilder/goclaw:latest`
- Provider: `claude_cli` (`@anthropic-ai/claude-code`), model sonnet
- Agent: predefined, tools_config allow includes `use_skill`, `skill_search`, `exec`, `delegate`, `datetime`, `knowledge_graph_search`

## Repro
1. Create a predefined agent on the `claude_cli` provider with an uploaded (managed) skill that is NOT pinned, tools allowlist including `use_skill`.
2. Ask something the skill covers.
3. Observed log sequence:
```
skill_search executed  query="task triage assign chores family" results=2 hybrid=true
skill_search executed  query="task assign household chore delegate member" results=2
debug.llm.first_response has_error=false tool_calls_count=0 tools_provided=17
```
The CLI found the skill via `skill_search` (bridge exposes it) but could not `use_skill` (not on the bridge). Our agent then replied "task created ✅" while the database had 0 rows — the tool-contract mismatch pushed it into fabrication. Repro'd twice before we worked around it with `other_config.pinned_skills` (inline mode needs no `use_skill`).

Related noise: every probe of a non-bridged tool logs `WARN security.mcp_bridge_denied tool=team_tasks agent_key=…` per run.

## Why this is a contract bug rather than a policy choice
`makeToolHandler` already enforces the per-agent policy at call time (`policyEngine.WouldAllow`, bridge_server.go:120) — the static list is only an outer bound, so widening it does not bypass any agent policy. Meanwhile `skillLoadingProtocolLines()` is provider-agnostic and hard-requires `use_skill` for the non-pinned path.

## Suggested directions (happy to PR whichever you prefer)
1. **Minimal:** add `use_skill` (and arguably `datetime`, `knowledge_graph_search`, `delegate`) to `BridgeToolNames`; per-call policy gate keeps enforcement unchanged.
2. **Structural:** drop the static list and derive the bridge's tool set from the agent's policy-filtered allowlist when building the MCP config, eliminating this class of drift.
3. Either way, consider downgrading `security.mcp_bridge_denied` to Info or log-once-per-session — the CLI probes denied tools every run.

## Impact
Any claude_cli agent relying on non-pinned managed skills or on `delegate` (agent_links delegation is effectively unavailable on this provider).

taminhtoan2601:
## Summary
Implements the structural fix (Option 2) recommended in the #1373 triage: the claude_cli MCP bridge now derives its tool surface from the calling agent's policy-filtered allowlist instead of the static `BridgeToolNames` set, eliminating the drift where the system prompt requires `use_skill` but the bridge never exposed it.

## Design
One shared predicate `bridgeToolAllowed` gates **both** `tools/list` (new — via mcp-go `WithToolFilter`, ctx-aware) and `tools/call`, so visibility and callability can never diverge:

| Caller | Surface |
|---|---|
| Agent with verified tool policy in ctx (X-Agent-ID + HMAC) | exactly `policyEngine.WouldAllow(...)` — the same check the call path always enforced |
| No agent policy (anonymous, or agent without `tools_config`) / nil engine | legacy conservative `BridgeToolNames` set — **no exposure widening** |
| `spawn`, `create_forum_topic` | hard-excluded always (can't operate over the bridge) |

Registration covers the full registry minus hard exclusions; `BridgeToolNames` is kept as the documented legacy fallback. The per-call denial log drops Warn→Info per triage note — list filtering makes probes rare and the call gate is the intended enforcement point.

## Correction to the issue report
One claim in my original #1373 report was wrong: `delegate` **is** already in `BridgeToolNames` (my grep window was truncated — apologies). The genuinely missing names were `use_skill`, `datetime`, `knowledge_graph_search`, `skill_manage`. With this PR the question disappears entirely: whatever the agent's policy allows is what the bridge serves. I've also commented on #1373 with this correction.

## Testing
- New `internal/mcp/bridge_server_test.go` (6 tests): legacy fallback for policy-less callers · policy-derived surface (use_skill allowed, non-allowed rejected) · hard exclusions win over policy · nil-engine legacy behavior · list filter matches callable surface · registration = registry minus exclusions.
- `go build ./...` + `go build -tags sqliteonly ./...` clean; `go vet` clean.
- `go test ./internal/mcp ./internal/tools ./internal/providers ./internal/gateway/...` all pass (`-race` skipped locally — no cgo toolchain; CI covers it).

## Real-world validation
Field context from the pilot that surfaced this: a claude_cli agent with non-pinned managed skills called `skill_search` successfully, could not `use_skill`, and fabricated "task created" with 0 DB rows. With policy-derived surface, `use_skill` is available whenever the agent's policy allows it.

### DIFF
--- a/internal/mcp/bridge_server.go
+++ b/internal/mcp/bridge_server.go
@@ -15,8 +15,13 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/tools"
 )
 
-// BridgeToolNames is the subset of GoClaw tools exposed via the MCP bridge.
-// Excluded: spawn (agent loop), create_forum_topic (channels).
+// BridgeToolNames is the LEGACY conservative tool set, kept as the fallback
+// surface for callers that carry no verified agent tool policy in context
+// (no/invalid X-Agent-ID headers, or an agent without a tools_config). For
+// callers WITH an agent policy, the bridge surface is derived from that
+// policy instead — see bridgeToolAllowed. This removes the drift where the
+// system prompt requires tools (e.g. use_skill) that a static list forgot to
+// expose (#1373), without widening exposure for unauthenticated callers.
 // delegate is included: it self-gates via CanDelegate/agent_links and resolves
 // its source agent from the X-Agent-ID header context, same as team_tasks.
 var BridgeToolNames = map[string]bool{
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — clean, well-scoped structural fix implementing the Option 2 recommendation from #1373 triage.

**Key design strengths:**
1. Single predicate `bridgeToolAllowed` gates both `tools/list` and `tools/call` — visibility and callability can never diverge
2. Backward compatible: anonymous/policy-less callers get the same legacy `BridgeToolNames` set (no exposure widening)
3. Hard exclusions (`spawn`, `create_forum_topic`) win over any policy — correct safety invariant
4. Nil engine fallback preserves legacy constructor path behavior
5. `WithToolFilter` integration via mcp-go is ctx-aware — per-caller filtering at list time
6. Denial log Warn→Info is correct: list filtering makes probes rare, call gate is the enforcement point

**Tests:** 6 comprehensive tests covering legacy fallback, policy-derived surface, hard exclusions, nil engine, list/call parity, and registration scope. Well-constructed.

**No security concerns** — this strictly tightens the feedback loop between policy and exposure without widening any attack surface.

PR #1378, opened 2026-07-06
fix(cron): scope cron context to tenant slug so managed skills resolve

taminhtoan2601:
## Summary
Cron job execution scoped its context to the tenant **ID** but not the tenant **slug**. Tenant-scoped filesystem paths (`skills-store`, `workspace`, `media` via `config.TenantScopedDir`) key off the slug and fall back to an **id-based** path when it is absent — a *different* directory than where HTTP/WS skill upload materializes files (that path sets the slug). Consequently a cron agent turn in a non-master tenant saw **none** of its tenant's managed skills.

## Symptom (field repro on a live tenant)
A daily "agenda" cron whose agent-turn message asks the agent to run an uploaded skill that queries the DB:
```
skill_search executed  query="daily agenda calendar"  results=0 hybrid=true   (×6, all 0)
v3.run.completed  agent=family-calendar  iterations=0  total_tokens=459
→ agent replied "hôm nay nhà mình rảnh" (no events today) — but by GUESS, never ran the skill
```
With a real event inserted for that day, the same cron still said "rảnh" (wrong). The agent's skill loader was resolving `…/tenants/<tenantUUID>/skills-store` while the skills live under `…/tenants/<slug>/skills-store`.

## Root cause
`cmd/gateway_cron.go` built every job context with `store.WithTenantID(...)` only. `config.TenantScopedDir` returns `base/tenants/<slug>` when a slug is present but `base/tenants/<tenantID>` when it is not — and skill upload (HTTP/WS) always runs with the slug set (`router.go` / `auth.go` call `store.WithTenantSlug`). So cron and upload disagreed on the directory.

## Fix
Add `cronTenantContext(ctx, tenantStore, tenantID)` that sets **both** tenant ID and slug (resolved via `TenantStore.GetTenant`). Master tenant, nil store, and lookup failure fall back to id-only (prior behavior, logged). Threaded `TenantStore` through `makeCronJobHandler` and `runCommandCronJob`; the single wiring call in `gateway_heartbeat.go` now passes `pgStores.Tenants`.

## Testing
- New unit tests (`cmd/gateway_cron_test.go`): slug injected for non-master tenant · master tenant skips the store lookup · nil store and lookup error both fall back to id-only.
- `go build ./...` + `go build -tags sqliteonly ./...` + `go vet ./cmd/` clean; `go test ./cmd/` passes (`-race` skipped locally — no cgo; CI covers it).
- End-to-end on the live tenant: after the fix the daily-agenda cron reads the real event from the DB and reports it correctly.

## Note for maintainers
Other background executors that construct a context from a tenant ID (heartbeat, and possibly consolidation/eventbus workers) likely share this gap — worth a quick audit. I scoped this PR to cron, where I have a concrete repro.

### DIFF
--- a/cmd/gateway_cron.go
+++ b/cmd/gateway_cron.go
@@ -32,12 +32,34 @@ var cronHeartbeatWakeFn func(agentID string)
 // so the stateless-reset behavior can be unit-tested without filesystem effects.
 var cronCLISessionReset = providers.ResetCLISession
 
-func makeCronJobHandler(sched *scheduler.Scheduler, msgBus *bus.MessageBus, cfg *config.Config, channelMgr *channels.Manager, sessionMgr store.SessionStore, agentStore store.AgentStore, providerStore store.ProviderStore, providerReg *providers.Registry) func(job *store.CronJob) (*store.CronJobResult, error) {
+// cronTenantContext scopes a context to the job's tenant. It sets BOTH the
+// tenant ID and the tenant SLUG: tenant-scoped filesystem paths
+// (skills-store, workspace, media via config.TenantScopedDir) key off the
+// slug, and resolve to an id-based path when the slug is absent — a different
+// directory than where HTTP/WS upload materialized the files. Without the
+// slug, a cron agent turn sees NONE of its tenant's managed skills. tenantStore
+// may be nil (older wiring) — then only the tenant ID is set, preserving prior
+// behavior. Master tenant needs no slug (TenantScopedDir returns the base).
+func cronTenantContext(ctx context.Context, tenantStore store.TenantStore, tenantID uuid.UUID) context.Context {
+	ctx = store.WithTenantID(ctx, tenantID)
+	if tenantStore == nil || tenantID == uuid.Nil || tenantID == store.MasterTenantID {
+		return ctx
+	}
+	tenant, err := tenantStore.GetTenant(ctx, tenantID)
+	if err != nil || tenant == nil || tenant.Slug == "" {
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean, well-scoped bug fix.

Root cause confirmed: cronTenantContext correctly identifies that store.WithTenantID alone produces a different config.TenantScopedDir path than store.WithTenantID + store.WithTenantSlug (which is what HTTP/WS upload uses). This divergence means cron agent turns in non-master tenants cannot see managed skills.

Fix review:
1. cronTenantContext helper sets BOTH tenant ID and slug — correct symmetry with the upload path
2. Master tenant short-circuits (no store lookup needed, slug irrelevant for base dir) — good
3. Nil store and lookup error both fall back to id-only with slog.Warn — correct degradation, preserves prior behavior for older wiring
4. tenantStore threaded through makeCronJobHandler and runCommandCronJob — single wiring point in gateway_heartbeat.go passes pgStores.Tenants
5. All existing test call sites updated with extra nil argument — no stale signatures
6. Tests cover: slug injected for non-master, master skips store lookup (with sentinel error asserting non-call), nil store → id-only, lookup error → id-only with id still scoped

Maintainer note in PR body about heartbeat/consolidation/eventbus sharing this gap is valuable — good that this is scoped to cron with a concrete repro.

Cannot merge yet — UNSTABLE (no CI checks reported ~15min since PR creation). Will wait for next run.

bclermont:
there is so many issues left with tennants, this had never been tested. I fixed so many of them in the last week

PR #1380, opened 2026-07-07
fix(discord): preserve channel agent routing

### LINKED ISSUE #1264: Discord channel connection not working
I have a naive set-up, OpenAI Codex and a single Discord channel pairing (a bot with full permissions, this bot previously worked fine in OpenClaw). When I type to the bot in a Discord channel, it almost immediately replies “Thinking…”, then after a few seconds, I see “<Bot> is typing…”, this stays for 10+ seconds, then disappears, I never get a response and the “Thinking…” message remains. 

If I give it a secret word and then ask it about it in the GoClaw Dashboard chat, it won’t be aware. If I ask it in chat what Discord-related stuff it sees, it says it can’t see anything, only the dashboard chat. Asking it to send me a message, it won’t know where. The only thing that works is explicitly giving it the channel id, it will then be able to text me but it won’t see the channel, other messages, or replies.

In OpenClaw, hooking up the Discord bot was enough to make it all work, for reference.

namnhfreelancer:
## Summary

Fixes #1264.

This restores Discord channel-instance agent routing that was accidentally bypassed by [`bdc9abd5`](https://github.com/nextlevelbuilder/goclaw/commit/bdc9abd5768bdbcea55f437375fe3fa748fa0fee) (`feat(discord): enable binding-based agent routing`, [#1243](https://github.com/nextlevelbuilder/goclaw/pull/1243)). That commit changed Discord inbound handling from using the channel instance's configured `AgentID` to always publishing an empty `AgentID`, forcing the gateway consumer to resolve via top-level `cfg.Bindings`.

That broke dashboard-configured Discord channels because Admin UI channel setup stores the selected agent in `channel_instances.agent_id`, and `InstanceLoader` resolves that UUID to `BaseChannel.SetAgentID(agent_key)`. If no separate config-file binding matched the Discord instance, the consumer fell back to the literal agent key `default`. Deployments without an agent keyed `default` then logged `inbound: agent not found agent=default`, never entered `loop.Run`, never created traces, and left the Discord `Thinking...` placeholder in place.

## Root cause

Before [`bdc9abd5`](https://github.com/nextlevelbuilder/goclaw/commit/bdc9abd5768bdbcea55f437375fe3fa748fa0fee), Discord followed the same model as Telegram/Feishu/WhatsApp:

```go
targetAgentID := c.AgentID()
```

[`bdc9abd5`](https://github.com/nextlevelbuilder/goclaw/commit/bdc9abd5768bdbcea55f437375fe3fa748fa0fee) replaced that with:

```go
targetAgentID := ""
```

That made binding-based routing possible, but it silently discarded the explicit agent configured on the Discord channel instance. The fallback chain then became:

1. ignore channel instance `AgentID`
2. try `cfg.Bindings`
3. fall back to config default literal `default`

For DB-managed channels, step 1 is the normal and expected route, so this regressed existing Discord setups.

## Changes

- Restore Discord's default target to `c.AgentID()` via `targetAgentID`.
- Preserve `voice_agent_id` behavior: audio/voice inbound still overrides the channel agent.
- Keep binding routing for cases where a channel has no explicit agent.
- Add DB-default fallback for inbound routing before falling back to literal `default`.
- Ensure pre-runtime agent lookup failures publish an outbound cleanup message, so external channels can remove placeholders instead of hanging.
- Add regression tests for:
  - Discord preserving the configured channel agent.
  - Discord voice-agent override only applying to audio/voice.
  - Inbound route fallback using DB default agent.
  - Binding rules taking precedence over DB default.
  - External-channel cleanup on agent lookup failure.

## Effective routing order after this PR

1. If the channel instance has `AgentID`, use it.
2. If the message is voice/audio and `voice_agent_id` is configured, override to that agent.
3. If no channel agent is present, use `cfg.Bindings`.
4. If no binding matches, use the tenant-scoped DB default agent.
5. Last resort: fall back to literal `default` for legacy config-only installs.

## Verification

- `go test ./cmd ./internal/channels/discord`
- `git diff --check -- cmd/gateway_consumer.go cmd/gateway_consumer_debounce.go cmd/gateway_consumer_handlers.go cmd/gateway_consumer_helpers.go cmd/gateway_consumer_normal.go cmd/gateway_consumer_normal_test.go internal/channels/discord/handler.go internal/channels/discord/handler_test.go`

## Risk

Low. This restores the pre-[`bdc9abd5`](https://github.com/nextlevelbuilder/goclaw/commit/bdc9abd5768bdbcea55f437375fe3fa748fa0fee) Discord channel-agent behavior and matches the routing pattern already used by Telegram, Feishu, and WhatsApp. Binding-based routing still works when no explicit channel agent is set, and tests pin the precedence.

### DIFF
--- a/cmd/gateway_consumer.go
+++ b/cmd/gateway_consumer.go
@@ -147,7 +147,7 @@ func consumeInboundMessages(ctx context.Context, msgBus *bus.MessageBus, agents 
 		}
 
 		// --- Normal messages: route through debouncer ---
-		prepareInboundDebounceMessage(&msg, deps)
+		prepareInboundDebounceMessage(ctx, &msg, deps)
 		debouncer.Push(msg)
 	}
 }
--- a/cmd/gateway_consumer_debounce.go
+++ b/cmd/gateway_consumer_debounce.go
@@ -18,11 +18,12 @@ import (
 // debouncing (issue #63). See plans/260528-1351-multi-attachment-debounce/.
 const mediaDebounceFloorMs = 1000
 
-func prepareInboundDebounceMessage(msg *bus.InboundMessage, deps *ConsumerDeps) {
+func prepareInboundDebounceMessage(ctx context.Context, msg *bus.InboundMessage, deps *ConsumerDeps) {
 	if msg == nil || deps == nil || deps.Cfg == nil || msg.AgentID != "" {
 		return
 	}
-	msg.AgentID = resolveAgentRoute(deps.Cfg, msg.Channel, msg.ChatID, msg.PeerKind)
+	routeCtx := inboundMessageTenantContext(ctx, *msg)
+	msg.AgentID = resolveAgentRouteForInbound(routeCtx, deps.Cfg, deps.AgentStore, msg.Channel, msg.ChatID, msg.PeerKind)
 }
 
 func resolveInboundDebounceDelay(ctx context.Context, msg bus.InboundMessage, deps *ConsumerDeps) time.Duration {
--- a/cmd/gateway_consumer_handlers.go
+++ b/cmd/gateway_consumer_handlers.go
@@ -401,7 +401,8 @@ func handleResetCommand(
 
 	agentID := msg.AgentID
 	if agentID == "" {
-		agentID = resolveAgentRoute(deps.Cfg, msg.Channel, msg.ChatID, msg.PeerKind)
+		ctx := inboundMessageTenantContext(context.Background(), msg)
+		agentID = resolveAgentRouteForInbound(ctx, deps.Cfg, deps.AgentStore, msg.Channel, msg.ChatID, msg.PeerKind)
 	}
 	peerKind := msg.PeerKind
 	if peerKind == "" {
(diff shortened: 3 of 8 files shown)


clark-cant (approved):
LGTM — clean regression fix for #1264.

**Summary**: Restores Discord channel-instance agent routing that was accidentally bypassed by #1243. The fix correctly:
1. Restores `c.AgentID()` as the default target in Discord handler
2. Preserves voice-agent override for audio/voice inbound
3. Adds DB default agent fallback in consumer before literal `default`
4. Publishes outbound cleanup on agent lookup failure (prevents hanging Discord placeholders)
5. Extracts `bindingMatchesInbound()` — clean refactor, no behavior change for existing paths

**Routing order is correct**: channel agent → voice override → bindings → DB default → literal default

**Tests**: 4 new tests cover DB fallback, binding precedence, external cleanup, and voice override. Good coverage.

**Risk**: Low — matches Telegram/Feishu/WhatsApp pattern. No anti-slop patterns detected.

Approved and merging.

PR #1383, opened 2026-07-07
Telegram: trigger-words, channel posts, message edit & topic posting (+ image_gen crash fix)

skensell201:
# Telegram: agent trigger-words, channel posts, message edit & topic posting (+ image_generation crash fix)

## Summary

Extends the Telegram channel so an agent can wake on name aliases, react in
broadcast channels, edit existing messages in place, and post into a named forum
topic — plus fixes a crash that killed every codex-provider agent on v3.14.0.

## Changes

**Fix — `image_generation` nil-pointer crash (all codex agents).**
`imageGenToolDef` was a `Type`-only `ToolDefinition` with a nil `Function`. Several
pipeline/provider sites read `td.Function.Name` on the tool set that includes it
(`makeBuildFilteredTools` mcp counter, `think_stage` allowlist, `shouldRetryTaskMCP`,
history tool names), so every message SIGSEGV'd. The sentinel now carries a
name-only `Function`, removing the whole nil-deref class at once; `codex_build`
still branches on `Type` for the native tool. Codex agents can keep image
generation enabled.

**Feature — agent-declared trigger words (group wake by alias).**
An agent declares aliases in its `IDENTITY.md` context file:
`Trigger words: Alice, Hey, Boss`. In group chats the bot then wakes without an
`@mention` when a message names it by one of these (whole-word, case-insensitive,
Unicode/Cyrillic-aware; matches text and media caption). Parsed by
`bootstrap.ParseTriggerWords`, matched by a source-agnostic matcher, cached
per-agent on the channel for 60s so `IDENTITY.md` edits apply without a restart.
Trigger-word loading is tenant-scoped (`GetAgentContextFiles` requires tenant in
ctx). Agent-owned so the words travel across every channel the agent serves.

**Feature — `channel_post` support.**
The poller now subscribes to `channel_post`; posts (which have no `From`) get a
synthetic sender via `resolveMessageSender` and flow through the same group-style
gate, so trigger words fire in broadcast channels too. A defensive `recover()` in
the update handler goroutine prevents any single malformed update from crashing
the gateway.

**Feature — `message` tool `action:"edit"`.**
The agent can edit an existing message in place (e.g. flip a status marker)
instead of only sending new ones, via a `ChannelEditor` →
`Manager.EditChannelMessage` → Telegram `editMessageText`, with a fallback to
`editMessageCaption` for media messages (captions). The replied-to message id is
surfaced to the agent as `reply_to_message_id` in the reply context so it can
target the edit. Channel/chat are taken from context (models otherwise pass the
platform name). Note: Telegram only lets a bot edit others' messages in a channel
where it is admin, never in a group.

**Feature — post into a forum topic by name (`message` `topic:"..."`).**
Learns forum topics (name → `message_thread_id`) from `forum_topic_created`
service messages into `channel_contacts` (`contact_type=topic`); a `TopicResolver`
resolves the name for the current group and the post is routed with the right
`message_thread_id`. Lets a user in one topic ask the bot to post into another.
Bots cannot enumerate topics, so only topics created while the bot is present are
addressable by name.

## Tests

New unit tests: `ParseTriggerWords`, the whole-word/Cyrillic matcher,
per-channel trigger loading (incl. the tenant-scope regression),
`resolveMessageSender`, the image_generation sentinel shape + mcp-def counter,
the caption-edit error match, and the message tool `edit`/`topic` actions
(context-channel preference, not-found handling). `go build ./...`,
`go build -tags sqliteonly ./...`, and `go vet` all pass.

## Operator notes (not code)

- Trigger words in groups require the bot's **Group Privacy = OFF** in BotFather
  (then re-add the bot), otherwise Telegram never delivers plain group messages.
- Trigger words are set in the agent's `IDENTITY.md` (no dedicated UI field yet —
  edited via the existing agent context-file editor).
- `channel_post` and topic editing require the bot to be a channel admin with the
  relevant rights.

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -718,6 +718,35 @@ func runGateway() {
 		if cs, ok := t.(tools.ChannelSenderAware); ok {
 			cs.SetChannelSender(channelMgr.SendToChannel)
 		}
+		if ce, ok := t.(tools.ChannelEditorAware); ok {
+			ce.SetChannelEditor(channelMgr.EditChannelMessage)
+		}
+		if tr, ok := t.(tools.TopicResolverAware); ok && pgStores != nil && pgStores.Contacts != nil {
+			contacts := pgStores.Contacts
+			tr.SetTopicResolver(func(ctx context.Context, channel, chatID, topicName string) (string, bool) {
+				list, err := contacts.ListContacts(ctx, store.ContactListOpts{
+					ChannelInstance: channel,
+					ContactType:     "topic",
+					Limit:           500,
+				})
+				if err != nil {
+					return "", false
+				}
+				want := strings.ToLower(strings.TrimSpace(topicName))
+				for _, c := range list {
+					if c.SenderID != chatID || c.ThreadID == nil || c.DisplayName == nil {
+						continue
+					}
+					if strings.ToLower(strings.TrimSpace(*c.DisplayName)) == want {
+						return *c.ThreadID, true
+					}
+				}
+				return "", false
+			})
+		}
+		if tp, ok := t.(tools.TopicPosterAware); ok {
+			tp.SetTopicPoster(channelMgr.PostToTopic)
+		}
 		if tc, ok := t.(tools.ChannelTenantCheckerAware); ok {
 			tc.SetChannelTenantChecker(channelMgr.ChannelTenantID)
 		}
--- a/docs/telegram-trigger-words.md
+++ b/docs/telegram-trigger-words.md
@@ -1,0 +1,69 @@
+# Telegram trigger words (wake an agent by name)
+
+In group chats a Telegram bot normally only reacts when it is `@mentioned`,
+addressed with a `/command@bot`, or replied to. **Trigger words** let an agent
+also wake up when a message names it by an alias — without an explicit mention.
+
+Trigger words are a property of the **agent**, not the channel, so they travel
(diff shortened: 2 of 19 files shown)


clark-cant (approved):
LGTM — clean multi-part PR.

**image_generation crash fix**: Root cause correct. The sentinel carried Type only with nil Function; multiple pipeline sites read td.Function.Name causing SIGSEGV. Populating Function with a name-only schema removes the whole nil-deref class at once. Tests cover the regression.

**Trigger words**: Clean implementation — ParseTriggerWords in bootstrap, Unicode-aware whole-word matcher, 60s cache per agent, tenant-scoped loading. Treated as @mention equivalent in the group gate. Tests cover parsing, matching (including Cyrillic), and tenant scope.

**channel_post**: Properly subscribes to channel_post updates, synthesizes sender via resolveMessageSender for posts without From, reuses group-style gate. Defensive recover() in goroutine prevents single malformed update from crashing gateway.

**message edit + topic posting**: Clean interface injection (ChannelEditor, TopicResolver, TopicPoster) following existing *Aware pattern. Forum topics learned from forum_topic_created service messages. Edit falls back to editMessageCaption for media. Tests cover edit, topic resolution, not-found, and message-id return.

**Risk**: Medium — large PR bundling 4 features + fix, but all coherent Telegram improvements. CI all green.

PR #1389, opened 2026-07-07
fix(exec): Unicode-aware boundary for privilege-escalation `su` deny (fixes non-ASCII false positives)

### LINKED ISSUE #1388: exec safety policy: privilege-escalation `\bsu\b` deny false-positives on non-ASCII text (e.g. Vietnamese "suất")
### Summary

The `exec` tool's `privilege_escalation` deny group contains the pattern `` `\bsu\b` `` (in `internal/tools/shell_deny_groups.go`) to block the Unix `su` command. Because Go's `regexp` (RE2) treats `\b` as an **ASCII-only** word boundary, this pattern matches the ASCII substring `su` whenever it is immediately followed (or preceded) by a **non-ASCII** letter. Common Vietnamese words such as **`suất`** (`hiệu suất`, `năng suất`) therefore match: the accented `ấ` (U+1EA5) is not an ASCII word character, so `\b` sees a spurious word boundary right after `su`.

Since the deny patterns are matched against the **entire command string** (including document text embedded in a heredoc), any exec command that merely *contains* such a word is rejected with:

```
command denied by safety policy: matches pattern \bsu\b
```

### Impact

An agent that embeds Vietnamese (or other diacritic-heavy) prose in an exec command — e.g. writing a report to a file, or building a PDF via a `python3 - <<'PY' … 'PY'` heredoc — is blocked at random depending on whether the text happens to contain a word like `suất`. The command is entirely benign and has nothing to do with privilege escalation. This is a hard blocker for non-English document/PDF generation workflows.

### Reproduction

```go
regexp.MustCompile(`\bsu\b`).MatchString("báo cáo hiệu suất")   // => true  (should be false)
regexp.MustCompile(`\bsu\b`).MatchString("năng suất")           // => true  (should be false)
regexp.MustCompile(`\bsu\b`).MatchString("su - root")           // => true  (correct)
```

### Root cause

RE2's `\b` is defined on ASCII `\w` (`[0-9A-Za-z_]`) only; Unicode letters/digits are treated as non-word, creating false boundaries adjacent to accented characters. This affects the short bare-word command patterns; `su` (two letters) is by far the most collision-prone and the one observed in practice.

### Suggested fix

Match the command name using a Unicode-aware boundary built from `\pL`/`\pN` (RE2 supports Unicode classes) instead of ASCII `\b`, so accented letters adjacent to the token count as word characters:

```go
`(?:^|[^\pL\pN_])(?:su)(?:[^\pL\pN_]|$)`
```

This still matches the real `su` command (start/space/`;`/`&&`/newline on either side) while ignoring `su` embedded in `suất`, `năng suất`, etc.

PR follows.

Buns0705:
Fixes #1388

### Problem

`privilege_escalation`'s deny pattern `` `\bsu\b` `` (in `internal/tools/shell_deny_groups.go`) uses Go RE2's ASCII-only `\b`. It matches the ASCII `su` inside non-ASCII words — e.g. the Vietnamese `suất` (`hiệu suất`, `năng suất`) — because the accented `ấ` is not an ASCII word char, so `\b` sees a spurious boundary right after `su`. Any exec command containing such a word (e.g. a Vietnamese document embedded in a `python3 … <<'PY'` heredoc to render a PDF) is wrongly rejected with `command denied by safety policy: matches pattern \bsu\b`.

### Fix

Add a small `unicodeCmdWord` helper that builds a **Unicode-aware** word boundary from `\pL`/`\pN` and use it for the `su` pattern:

```go
`(?:^|[^\pL\pN_])(?:su)(?:[^\pL\pN_]|$)`
```

Accented letters adjacent to the token now count as word characters, so `suất` no longer matches while the real `su` command (bounded by start/space/`;`/`&&`/newline/EOF) still does. Scope is intentionally limited to `su`, the only short bare-word pattern observed to collide with real text; the helper is reusable for other patterns if similar reports appear.

### Tests

Extended `TestPrivilegeEscalationGaps`:
- **still denied:** `su`, `su -`, `su root`, `su -l postgres`, `su admin` (existing cases unchanged)
- **now allowed (new):** `hiệu suất`, `năng suất`, `báo cáo hiệu suất tháng`

`go test ./internal/tools/ -run PrivilegeEscalation` passes. The `su` matching behaviour for real commands is unchanged.

### DIFF
--- a/internal/tools/shell_deny_groups.go
+++ b/internal/tools/shell_deny_groups.go
@@ -4,6 +4,17 @@ import (
 	"regexp"
 )
 
+// unicodeCmdWord matches a command name (or alternation) as a standalone
+// word using a Unicode-aware boundary. Go's RE2 \b is ASCII-only, so a
+// pattern like \bsu\b matches the "su" inside non-ASCII words such as the
+// Vietnamese "suất" (the accented 'ấ' is not an ASCII word char, so \b
+// sees a spurious boundary right after "su"). Treating Unicode letters and
+// digits (\pL, \pN) as word chars avoids those false positives while still
+// matching the real `su` command.
+func unicodeCmdWord(alt string) *regexp.Regexp {
+	return regexp.MustCompile(`(?:^|[^\pL\pN_])(?:` + alt + `)(?:[^\pL\pN_]|$)`)
+}
+
 // DenyGroup is a named set of shell command deny patterns.
 type DenyGroup struct {
 	Name        string           // machine name: "package_install"
@@ -83,7 +94,7 @@ var DenyGroupRegistry = map[string]*DenyGroup{
 		Default:     true,
 		Patterns: []*regexp.Regexp{
 			regexp.MustCompile(`\bsudo\b`),
-			regexp.MustCompile(`\bsu\b`),
+			unicodeCmdWord("su"),
 			regexp.MustCompile(`\bdoas\b`),
 			regexp.MustCompile(`\bpkexec\b`),
 			regexp.MustCompile(`\brunuser\b`),
--- a/internal/tools/shell_deny_test.go
+++ b/internal/tools/shell_deny_test.go
@@ -130,14 +130,19 @@ func TestPrivilegeEscalationGaps(t *testing.T) {
 	)
 
 	mustAllow(t, patterns,
-		"summit",    // not "su"
-		"sugar",     // not "su"
-		"surplus",   // not "su"
-		"issue",     // not "su"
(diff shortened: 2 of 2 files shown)


clark-cant (approved):
LGTM — clean Unicode-aware word boundary fix. The `unicodeCmdWord` helper correctly uses `\pL`/`\pN` classes to avoid ASCII-only `\b` false positives on non-ASCII text. Real `su` command matching is preserved. Tests cover both deny and allow cases including Vietnamese text. Minimal scope, no security regression.

PR #1392, opened 2026-07-08
feat(bitrix24): replace two MCP text inputs with a filtered dropdown

tech-synity:
## Problem

Bitrix24 channel creation used to demand two hand-typed strings — `mcp_server_name` and `mcp_base_url` — plus zero indication of which MCP servers can actually auto-onboard. Typos silently disabled provisioning and admin had to know from external documentation which servers implement `/api/auto-onboard`.

## Fix

Ship a single dropdown backed by `mcp_servers.require_user_credentials`, plus every backend/frontend piece needed to make it work end-to-end. Legacy configs keep working through a fallback path so this is a rolling upgrade, not a breaking change.

### Highlights across the two commits

**`feat(bitrix24): replace two MCP text inputs with a filtered dropdown`**

- Promotes `require_user_credentials` from `mcp_servers.settings` JSONB to a top-level column (PG migration 000089, SQLite v54). Backfill reads the legacy JSONB entry so no admin has to re-tick anything.
- Adds `MCPServerID` (UUID) to `bitrix24.InstanceConfig`. Provisioner prefers `GetServer(id)` and derives the base URL from the mcp_servers row; the legacy `mcp_server_name + mcp_base_url` pair stays wired as a fallback so half-migrated fleets keep booting.
- New `mcp-select` field type + `MCPServerSelect` component. Uses the shared `useMCP()` react-query cache and filters client-side to servers whose `require_user_credentials` is true. Explicit "None (disable MCP provisioning)" option so admins can clear the binding without editing config JSON.
- Frontend MCP form now sends the promoted flag both at the top level and inside the legacy `settings` JSONB so mid-rollout backends stay consistent.
- PG migration 000090 + SQLite v55 backfill existing `channel_instances.config` rows with `mcp_server_id` by resolving the legacy name against `mcp_servers`, tenant-scoped via `agents.tenant_id`. Idempotent — only touches rows already carrying `mcp_server_name` that lack `mcp_server_id`.

**`fix(bitrix24): derive auto-onboard base URL from mcp_servers.url origin`** — regression discovered in live-testing the first commit and fixed before shipping the PR.

`mcp_servers.url` stores the JSON-RPC endpoint the agent loop dials for tool calls (e.g. `https://mcp.example.com/mcp`), while the auto-onboard REST call the provisioner POSTs expects the ORIGIN only (`https://mcp.example.com`). The refactor passed the JSON-RPC URL verbatim to `mcp_client.newMCPClient`, which appends `/api/auto-onboard` → `.../mcp/api/auto-onboard` → 404 on every mint or refresh. `deriveAutoOnboardBaseURL` now strips the path/query/fragment; the legacy path is untouched because `MCPBaseURL` from channel config is already the origin.

## Verification

```
go vet ./internal/mcp/... ./internal/channels/bitrix24/... ./internal/store/... ./internal/http/...
go test ./internal/mcp/... ./internal/channels/bitrix24/...   → ok
```

New tests added in this branch:

- `TestParseEntityContext_*` on the prior chat-entity work (still green)
- `TestInitMCPProvisioner_MCPServerID` and two `DisabledModes` subtests for the new `mcp_server_id` factory path (valid UUID + row, invalid UUID string, valid UUID missing row)
- `TestDeriveAutoOnboardBaseURL` — six happy shapes + four error shapes covering origin stripping and rejection of unusable inputs

Live-tested against a local build on the goclaw-deploy postgres:

- Migrations 000089 + 000090 applied cleanly. Every existing `bitrix24` `channel_instances.config` picked up `mcp_server_id` while retaining its legacy pair.
- Provisioner boot log for every existing channel: `bitrix24 mcp: provisioning enabled mcp_server_id=<uuid> require_user_credentials=true`.
- A new bitrix24 channel created through the UI writes `mcp_server_id` at top level and does NOT emit `mcp_server_name` / `mcp_base_url` — the target long-term shape.
- Real user turn (`user_id=614`) triggered the expired-credential refresh branch. Before the fix commit: `auto-onboard failed: mcp auto-onboard: 404 Not Found`. After: `bitrix24 mcp: self-refreshed user credentials created=false` followed immediately by `mcp.user_tools_loaded user=614 tools=2`.

## Deferred (log-only, no PR follow-up needed to ship this)

- `MCPServerSelect` dropdown shows the placeholder if the currently-persisted `mcp_server_id` is not in the filtered list (server was unchecked or deleted). Config value is untouched unless the admin picks a new one. Minor UX polish.
- `sqlitestore` idempotency guard for migration 53 replaces the whole patch with `SELECT 1;` if the column already exists — the JSONB backfill won't run in that hand-migrated edge case. Consistent with the existing idempotency pattern in the file; documented, not a bug in this PR.
- No factory-side check that `mcp_server_id` points to a server whose `require_user_credentials` is true. The dropdown filters correctly, so this only bites operators editing config JSON by hand.

## Surface parity

- **Gateway server**: store + factory + provisioner + HTTP allowlist + two migrations.
- **API contract**: adds `require_user_credentials` + `mcp_server_id` as optional fields on existing routes. No new endpoints, no breaking shape changes.
- **Web UI**: MCP form + Bitrix24 channel form + shared types (web and desktop mirror).
- **CLI/runtime package**: N/A because no CLI subcommand reads `mcp_server_id` or the promoted flag; the `gateway` command comment was refreshed but no flag surface changed.

[B24:2794]

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -750,8 +750,9 @@ func runGateway() {
 		// the one used by pg.NewPGStores → NewPGBitrixPortalStore.
 		bitrixEncKey := os.Getenv("GOCLAW_ENCRYPTION_KEY")
 		// Use the MCP-aware factory variant so channels that opt into
-		// lazy per-user credential provisioning (via mcp_server_name +
-		// mcp_base_url in their instance config) can reach the partner's
+		// lazy per-user credential provisioning (via mcp_server_id — or
+		// the legacy mcp_server_name + mcp_base_url pair — in their
+		// instance config) can reach the partner's
 		// MCPServerStore. The MCP server authenticates each onboard call
 		// via the caller-supplied Bitrix access_token (the "Bitrix24
 		// OAuth → existing mcp_user_credentials bridge" — Bitrix-specific
--- a/internal/channels/bitrix24/factory.go
+++ b/internal/channels/bitrix24/factory.go
@@ -88,31 +88,43 @@ type bitrixInstanceConfig struct {
 
 	// Optional MCP lazy-provisioning binding (Phase C).
 	//
-	// When MCPServerName + MCPBaseURL are set AND the factory variant that
-	// accepts a MCPServerStore is used (FactoryWithPortalStoreAndMCP), the
-	// channel tries to mint per-user MCP credentials on first message:
+	// When MCPServerID (or the legacy MCPServerName+MCPBaseURL pair) is set
+	// AND the factory variant that accepts a MCPServerStore is used
+	// (FactoryWithPortalStoreAndMCP), the channel tries to mint per-user MCP
+	// credentials on first message:
 	//
 	//   1. Channel receives message from user U with OAuth tokens in event.
 	//   2. Channel looks up MCPUserCredentials(serverID, senderID). Present
-	//      → skip. Absent → POST /api/auto-onboard on MCPBaseURL forwarding
-	//      U's OAuth tokens. MCP server authenticates the call via Bitrix
-	//      `profile` against the supplied access_token — no shared admin
-	//      secret required — and responds with a per-user api_key, which
-	//      channel stores via SetUserCredentials (the "Bitrix24 OAuth →
-	//      existing mcp_user_credentials bridge" — Bitrix-specific glue,
-	//      not a generic MCP architecture pattern).
+	//      → skip. Absent → POST /api/auto-onboard on the resolved base URL
+	//      forwarding U's OAuth tokens. MCP server authenticates the call
+	//      via Bitrix `profile` against the supplied access_token — no
+	//      shared admin secret required — and responds with a per-user
+	//      api_key, which channel stores via SetUserCredentials (the
+	//      "Bitrix24 OAuth → existing mcp_user_credentials bridge" —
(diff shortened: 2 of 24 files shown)


clark-cant (approved):
LGTM — clean implementation of the Bitrix24 MCP dropdown refactor.

**Highlights:**
1. Proper promotion of `require_user_credentials` from JSONB to top-level column (PG migration 000092, SQLite v54) with backfill
2. `MCPServerID` preferred wiring with correct legacy fallback to `mcp_server_name + mcp_base_url`
3. `deriveAutoOnboardBaseURL` correctly strips path/query/fragment from `mcp_servers.url` to derive the origin for `/api/auto-onboard` — fixes the Phase 2 regression where JSON-RPC endpoint was passed verbatim (404)
4. Comprehensive test coverage: `TestInitMCPProvisioner_MCPServerID`, `TestDeriveAutoOnboardBaseURL` (6 happy + 4 error shapes), invalid UUID, missing row
5. Manager correctly checks both column and legacy JSONB (`srv.RequireUserCredentials || requireUserCreds(srv.Settings)`)
6. Migrations are idempotent; backward-compatible with half-migrated fleets
7. Frontend `MCPServerSelect` uses shared `useMCP()` cache with proper filtering

No critical or important issues found. Rolling upgrade, not a breaking change.

PR #1395, opened 2026-07-08
fix: resolve Zalo group by name + notify origin on failed forward

### LINKED ISSUE #3: feat(mcp): wire per-agent MCP server loading in managed mode
## Summary

`mcp.Manager.LoadForAgent()` exists but is never called. In managed mode, MCP servers are still loaded globally from `config.json` (`gateway.go:200-209`), not per-agent from DB.

## Current state (infrastructure is ready)

- **DB schema ready:** `mcp_servers`, `mcp_agent_grants` (with `config_overrides` JSONB), `mcp_user_grants`
- **Store methods ready:** `ListAccessible()`, `GrantToAgent()`, etc.
- **Manager method ready:** `LoadForAgent(ctx, agentID, userID)`
- **HTTP API ready:** `/v1/mcp/servers/{id}/grants/agent`

## What's missing

1. `ListAccessible()` doesn't SELECT `config_overrides` from `mcp_agent_grants` — the column exists but isn't included in the query result
2. `LoadForAgent()` doesn't merge `config_overrides.env` into server env before spawning
3. `wireManagedExtras()` / resolver doesn't call `LoadForAgent()` per agent — MCP is only initialized once globally
4. MCP tools register into global registry — need per-agent clone (similar to `DynamicToolLoader.LoadForAgent()` pattern)

## Use case

Running 2+ agents in managed mode with different MCP server configs. Example: separate NeuralMemory brains per agent via env var overrides in `config_overrides`:

```json
// mcp_agent_grants.config_overrides for agent "tho-ngoc"
{"env": {"NEURALMEMORY_DIR": "/home/user/.neuralmemory-tho-ngoc"}}

// mcp_agent_grants.config_overrides for agent "rua"
{"env": {"NEURALMEMORY_DIR": "/home/user/.neuralmemory-rua"}}
```

Both agents share the same `mcp_servers` record (command: `nmem mcp`) but get different env vars per grant → brain isolation without duplicate server definitions.

## Proposed approach

1. Add `ConfigOverrides []byte` to `MCPAccessInfo` and include `mag.config_overrides` in `ListAccessible()` SQL
2. In `LoadForAgent()`, parse `config_overrides.env` and merge into server env map before `connectServer()`
3. In resolver's `resolveAgent()`, create per-agent MCP Manager with cloned registry (after `DynamicToolLoader` clone), call `LoadForAgent()`
4. Handle MCP process lifecycle cleanup when agent loop stops

Happy to submit a PR if this aligns with your design intent.

### LINKED ISSUE #1394: Forwarding a message to a Zalo group by name silently fails while the agent reports success
## Summary

When an agent forwards a message to a different chat via the `message` tool (`forward=true`), and the destination is identified by a human display name rather than a real chat ID, the send fails downstream — but the tool already returned success, and nothing ever surfaces the real failure. The user is told the message was delivered when it never left the server.

Two separate root causes combine to produce this:

## 1. No way for the agent to resolve a group's display name to its real chat ID (Zalo)

`sessions_list` only returns session keys (`group:<numeric id>`), never a human-readable name:

```
{"key":"group:1234567890123","message_count":43,"updated":"2026-07-08T09:35:39Z"}
```

There's no tool exposing the account's group list with names, so when a user asks the agent to "forward this to the Ops group", the agent has no reliable way to turn that name into the chat ID `message`'s `target` parameter requires (per its own docstring: `"target": "Chat ID to send to"`). In the trace below, the agent ends up passing the display name itself as `target`.

## 2. `dispatchOutbound` silently swallows the resulting failure

`internal/tools/message.go`'s `Execute` treats `msgBus.PublishOutbound()` as immediate success — the tool returns `{"status":"sent",...}` to the model, and (for a `forward=true` cross-target send) `postCrossTargetNotice` announces the "forward" to the origin chat, *before* the async bus consumer (`dispatchOutbound` in `internal/channels/dispatch.go`) has even attempted the real send.

When that real send then fails, `dispatchOutbound`'s error handling only retries/notifies for **media** failures, and only against the **same (already broken) destination**:

```go
// Try to send a text-only error notification back to the chat.
// Only for media failures — text-only failures likely mean the chat
// is inaccessible (kicked, blocked, etc.) so retrying won't help.
if len(msg.Media) > 0 {
    ...
}
```

For a text-only forward with an invalid target, this assumption is wrong — the failure is a bad/unresolvable target, not an inaccessible chat, and it's entirely recoverable if the user is told. Instead the failure is dropped with only a `slog.Error` line; neither the agent nor the user ever finds out.

## Reproduction (production trace, redacted)

Agent trace, channel `bunny-zalo-personal` (Zalo Personal):

1. User (from Group A) asks the agent to forward a message + image into a different group ("Ops") and ping a teammate there.
2. Agent calls `sessions_list` + several `sessions_history` lookups trying to find a session matching that group name — no name field is exposed anywhere, so it can't reliably resolve it.
3. Agent calls `message` with `channel=zalo_personal` (wrong channel identifier — self-corrects on retry) then `channel=<the real channel instance name>, target="Ops", forward=true`.
4. Tool call returns `{"status":"sent","channel":"...","target":"Ops"}`.
5. Server log, ~20ms later:
   ```
   level=ERROR msg="error sending message to channel" channel=<channel> chat_id="Ops"
   error="zalo_personal: decrypt send response: zalo_personal: inner error code 114: Tham số không hợp lệ"
   ```
6. Agent tells the user the message was forwarded successfully — it never arrived.

## Expected behavior

- The agent should have a way to resolve a group's real chat ID from its display name before sending.
- If a forward's downstream delivery genuinely fails, the chat that requested the forward should be told, instead of the failure being dropped silently while a false "sent" confirmation stands.

## Environment

- goclaw self-hosted (Docker), `dev` branch
- Channel: Zalo Personal (`internal/channels/zalo/personal`)
- Go 1.26

Buns0705:
Fixes #1394.

## Problem

Forwarding a message to a different chat (`message` tool, `forward=true`) with a display name instead of a real chat ID fails downstream, but the agent already reported success and nothing surfaces the real failure — see #1394 for the full trace-backed root cause analysis. Two more related issues were found during live verification and are included here since they're part of the same forward-to-group incident chain.

## Changes

**1. New `zalo_list_groups` tool** — lets the agent resolve a Zalo group's display name to its real chat ID before sending/forwarding, instead of guessing.
- `channels.GroupInfo` + `channels.GroupListProvider` (mirrors the existing `GroupMember`/`GroupMemberProvider` pattern for `list_group_members`)
- `Manager.ListGroups` delegator
- `zalo/personal.Channel.ListGroups` implementation, wrapping the already-used (dashboard group picker) `protocol.FetchGroups`
- `ListGroupsTool` (`internal/tools/list_groups.go`), gated to `zalo_personal` via `RequiredChannelTypes`
- `message` tool's `target` parameter description now points at this when a display name is all that's known

**2. `dispatchOutbound` notifies the origin chat on a failed forward** — `message.go` tags cross-target forward sends with the origin channel/chat in `OutboundMessage.Metadata` (`bus.MetaForwardOriginChannel`/`MetaForwardOriginChatID`). `dispatchOutbound`'s error handling (extracted into `Manager.handleSendFailure` for testability) checks this first: if present, it notifies the ORIGIN chat with the real failure instead of retrying against the same broken destination, or — for text-only forwards — silently dropping it as before. Non-forward behavior (retry-notify the same chat, media-only) is unchanged.

**3. Warm the Zalo group-approval cache so forwards from non-group origins route correctly** — found while verifying #1/#2 live: forwarding from a DM into a group with a *correctly resolved* real group ID still silently went nowhere. `send.go` picks `ThreadTypeGroup` vs `ThreadTypeUser` from `BaseChannel`'s `approvedGroups` cache (or an explicit `group_id` metadata key). That cache is only ever populated by inbound group traffic under the `pairing` group policy — under `allowlist` (this integration's actual default) it stays empty for the process lifetime, and it's wiped on every restart (in-memory only). `message.go` only attaches `group_id` metadata when the *origin* session is itself a group (`isGroupContext(ctx)`), so a DM→group forward carries no such metadata and no cache entry — `Send()` defaults to `ThreadTypeUser`, addressing the message as if to a user account. It goes out with no error anywhere in the path, so nothing surfaces it. `ListGroups` now marks every returned ID as approved via `MarkGroupApproved`, and `Channel.Start` fires a best-effort `ListGroups` call right after connecting so the cache is warm from process start, not just after an explicit `zalo_list_groups` call.

**4. Expose a media tag for quote-forwarded images, not just vision access** — found while testing forwarding a quoted image to another chat. `extractQuoteMedia` downloads the image attached to a quoted message so the model can see it via vision, but the composed content only ever carried the literal `"[Quoted image]"` placeholder text — no `<media:image>` tag was ever inserted for it. The agent-side enrichment (`enrichImageIDs`/`enrichImagePaths`) only fills `id`/`path` attributes into a bare tag it finds already present in content; with no tag to find, the downloaded file had no path reference the model could hand to `message(MEDIA:<path>)` to forward it, even though the file was already on disk. `buildQuoteMediaTag` renders the same bare `<media:image>` tag a direct (non-quoted) attachment gets, appended after the quote wrapper text in all three call sites (`handleDM`, `handleGroupMessage`, `extractContentAndMediaWithQuote`).

## Testing

- `go build ./...`, `go vet`, `gofmt -l` on all touched files — clean
- New tests: `internal/channels/dispatch_test.go` (`TestHandleSendFailure_*`), `internal/channels/list_groups_test.go` (`TestManagerListGroups*`), `internal/tools/list_groups_test.go` (`TestListGroupsTool_*`), `internal/tools/message_test.go` (`TestMessageToolForward_TagsOriginMetadataOnGroupSend`), `internal/channels/zalo/personal/content_media_quote_tag_test.go` (`TestBuildQuoteMediaTag`)
- Full existing suite for `internal/tools`, `internal/channels` (incl. `zalo/personal`), `internal/bus`, `cmd` passes on a fresh clone of this branch — no regressions
- Fixes #1 and #2 verified live against a real Zalo Personal account/production trace (group-name resolution + forward succeeded with no silent drop, for a group-origin forward)
- Fixes #3 and #4 verified by build + clean restart on the live deployment; #3's underlying `protocol.FetchGroups` call and #4's `downloadFile` both require live network access (the latter is also blocked by the existing SSRF guard against `httptest`-style local test servers), so neither has a full integration test — happy to add one if maintainers can point at a preferred mocking seam

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -728,6 +728,12 @@ func runGateway() {
 			gl.SetGroupMemberLister(channelMgr.ListGroupMembers)
 		}
 	}
+	// Wire group lister on zalo_list_groups tool
+	if t, ok := toolsReg.Get("zalo_list_groups"); ok {
+		if gl, ok := t.(tools.GroupListerAware); ok {
+			gl.SetGroupLister(channelMgr.ListGroups)
+		}
+	}
 	// Wire Telegram manager on telegram_manager tool.
 	for _, toolName := range []string{"telegram_manager", "create_forum_topic"} {
 		if t, ok := toolsReg.Get(toolName); ok {
--- a/cmd/gateway_tools_wiring.go
+++ b/cmd/gateway_tools_wiring.go
@@ -67,6 +67,8 @@ func wireExtraTools(
 	toolsReg.Register(tools.NewSendFileTool(workspace, agentCfg.RestrictToWorkspace))
 	// Group members tool (list members in group chats)
 	toolsReg.Register(tools.NewListGroupMembersTool())
+	// Zalo group list tool (resolve a group's real chat ID from its display name)
+	toolsReg.Register(tools.NewListGroupsTool())
 	// Telegram manager tool (admin/forum/message management; gated by tool policy)
 	// create_forum_topic is kept as a backward-compatible wrapper for topic.create.
 	toolsReg.Register(tools.NewCreateForumTopicTool(nil))
--- a/internal/agent/systemprompt.go
+++ b/internal/agent/systemprompt.go
@@ -217,6 +217,7 @@ var coreToolSummaries = map[string]string{
 	"knowledge_graph_search": "Find people, projects, and their connections — use for relationship questions (who works with whom, project dependencies) that memory_search may miss",
 	"team_tasks":             "Team task board — track progress, manage dependencies (spawn auto-creates delegation tasks)",
 	"list_group_members":     "List all members of the current group chat (Feishu/Lark only)",
+	"zalo_list_groups":       "List all Zalo groups the account belongs to (group_id + name) — resolve a group's real ID before sending/forwarding to it by name (Zalo only)",
 	"create_forum_topic":     "Create a forum topic in a Telegram supergroup",
 	"delegate":               "Delegate a task to a linked agent (requires agent_links). See ## Delegation Targets for available agents",
 	"memory_expand":          "Retrieve full session details from episodic memory results — use after memory_search returns episodic hits",
--- a/internal/bus/types.go
+++ b/internal/bus/types.go
@@ -47,6 +47,18 @@ type OutboundMessage struct {
 	AgentOtherConfig []byte            `json:"agent_other_config,omitempty"` // agent's other_config for TTS voice/model
 }
 
+// Metadata keys on OutboundMessage.Metadata used to track the origin chat of
+// a cross-target forward (message tool, forward=true). The outbound dispatch
+// consumer runs async with no path back to the tool call that queued it, so
+// these let it notify the ORIGIN chat if delivery to the forward target
+// actually fails — otherwise a bad target (e.g. a display name instead of a
+// real chat ID) fails silently downstream while the tool already reported
+// success back to the model.
+const (
+	MetaForwardOriginChannel = "forward_origin_channel"
+	MetaForwardOriginChatID  = "forward_origin_chat_id"
+)
+
 // MediaAttachment represents a media file to be sent with a message.
 type MediaAttachment struct {
 	URL         string `json:"url"`                    // file path or URL
--- a/internal/channels/channel.go
+++ b/internal/channels/channel.go
@@ -724,6 +724,20 @@ type GroupMemberProvider interface {
 	ListGroupMembers(ctx context.Context, chatID string) ([]GroupMember, error)
 }
 
+// GroupInfo represents a group/channel chat the account belongs to.
+type GroupInfo struct {
+	GroupID     string `json:"group_id"`
+	Name        string `json:"name"`
+	TotalMember int    `json:"total_member,omitempty"`
+}
+
+// GroupListProvider is optionally implemented by channels that can list the
+// groups/chats the connected account belongs to — lets the agent resolve a
+// group's display name to its real chat ID instead of guessing.
+type GroupListProvider interface {
+	ListGroups(ctx context.Context) ([]GroupInfo, error)
+}
+
 // GroupTitleProvider is optionally implemented by channels that can resolve
 // a platform group/channel ID to a human-readable title.
 type GroupTitleProvider interface {
--- a/internal/channels/dispatch.go
+++ b/internal/channels/dispatch.go
@@ -85,53 +85,7 @@ func (m *Manager) dispatchOutbound(ctx context.Context) {
 			}
 
 			if err := channel.Send(sendCtx, msg); err != nil {
-				slog.Error("error sending message to channel",
-					"channel", msg.Channel,
-					"chat_id", msg.ChatID,
-					"content_len", len(msg.Content),
-					"content_preview", Truncate(msg.Content, 160),
-					"error", err,
-				)
-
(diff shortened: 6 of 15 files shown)


clark-cant (approved):
LGTM — clean two-part fix for #1394.

**1. zalo_list_groups tool**: Well-structured. Follows the existing GroupMemberProvider/GroupMemberLister pattern precisely. Gated to zalo_personal via RequiredChannelTypes. Wiring in gateway.go and gateway_tools_wiring.go is correct.

**2. Forward-failure notification**: handleSendFailure extraction is clean. Metadata propagation via MetaForwardOriginChannel/MetaForwardOriginChatID is the right approach — bus consumer is async with no callback path, so stashing origin in metadata is the only way to route failures back. Non-forward behavior preserved correctly.

**Tests**: Comprehensive — 4 new test files covering the actual bug scenario (display name as chatID → origin notification), non-forward regression, tool delegation, and metadata tagging. No phantom coverage.

**CI**: All green (release-versioning, go, web).

Risk: Low. Focused scope, follows existing patterns, well-tested.

PR #1397, opened 2026-07-08
feat(bootstrap): skip built-in USER.md for predefined agents with USER_PREDEFINED.md

bclermont:
Summary

  When a predefined agent has an operator-authored USER_PREDEFINED.md, the built-in per-user USER.md template is no longer seeded or injected into the system prompt. This gives operators full control over the user-context portion of the prompt and stops
  the per-turn "USER PROFILE INCOMPLETE" nag (name / timezone / pronouns).

  Motivation

  For predefined agents, the built-in internal/bootstrap/templates/USER.md was seeded per-user and injected into the system prompt regardless of whether the operator had authored a USER_PREDEFINED.md. Because the template ships with blank Name / Pronouns
  / Timezone fields, the system prompt fired a "learn the user's name/timezone" instruction on every turn until the Name field was filled — overriding operator intent and cluttering the prompt with content the operator can't control.

  USER_PREDEFINED.md was never consulted as a suppression condition anywhere, so authoring it did not stop the built-in template from imposing itself. This PR closes that gap.

  Changes

  - internal/bootstrap/seed_store.go — SeedUserFiles now detects an operator-authored USER_PREDEFINED.md in the agent-level files and, for predefined agents, skips seeding both USER.md and BOOTSTRAP.md (skipping USER.md alone would leave the onboarding
  nudge asking the model to recreate it). Covers both the normal seed loop and the channel-prefill path. Uses already-fetched data — no extra query.
  - internal/agent/loop_history.go — new dropBuiltinUserFileIfPredefined helper strips the built-in USER.md from resolved context files at inject time when a predefined agent has USER_PREDEFINED.md. Applied at all return paths of resolveContextFiles.
  This is a belt-and-suspenders layer covering pre-existing seeded users and the fact that the write_file tool can still recreate USER.md.
  - Tests — internal/bootstrap/seed_store_test.go (skip on normal + channel-prefill paths) and new internal/agent/loop_history_userpredefined_test.go (helper unit tests).

  Scope / safety

  - Only affects predefined agents that have USER_PREDEFINED.md.
  - Open agents (which never have USER_PREDEFINED.md) are untouched.
  - Predefined agents without USER_PREDEFINED.md keep current behavior.

  Surface parity

  - Gateway/bootstrap: ✅ covered (seeding + injection).
  - API contract / Web UI / CLI: N/A — no request/response shape, UI, or CLI command changes; this only alters which context file the backend seeds/injects.

  Verification

  - go build ./... ✅
  - go vet ./... ✅
  - go test ./internal/bootstrap/... ./internal/agent/... ✅
  - Web pnpm lint / pnpm build ✅ (pre-existing warnings only)

### DIFF
--- a/internal/agent/loop_history.go
+++ b/internal/agent/loop_history.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"path/filepath"
 	"strings"
 
 	"github.com/google/uuid"
@@ -307,14 +308,14 @@ func (l *Loop) buildMessages(ctx context.Context, history []providers.Message, s
 // but base-only files (like auto-injected delegation info) are preserved.
 func (l *Loop) resolveContextFiles(ctx context.Context, userID string) []bootstrap.ContextFile {
 	if l.contextFileLoader == nil || userID == "" {
-		return l.contextFiles
+		return dropBuiltinUserFileIfPredefined(l.agentType, l.contextFiles)
 	}
 	userFiles := l.contextFileLoader(ctx, l.agentUUID, userID, l.agentType)
 	if len(userFiles) == 0 {
-		return l.contextFiles
+		return dropBuiltinUserFileIfPredefined(l.agentType, l.contextFiles)
 	}
 	if len(l.contextFiles) == 0 {
-		return userFiles
+		return dropBuiltinUserFileIfPredefined(l.agentType, userFiles)
 	}
 
 	// Merge: start with per-user files, then append base-only files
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean two-layer fix. Seed layer correctly skips USER.md+BOOTSTRAP.md when USER_PREDEFINED.md exists for predefined agents. Inject layer (dropBuiltinUserFileIfPredefined) provides belt-and-suspenders coverage for pre-existing seeds and write_file recreations. Scope is tight: only predefined agents with USER_PREDEFINED.md are affected. Tests cover all three branches (with/without USER_PREDEFINED, open agent untouched) plus channel-prefill path.

PR #1399, opened 2026-07-08
feat(tools): seed missing built-in tools into builtin_tools catalog

bclermont:
Eleven built-in tools were registered in the runtime tool registry — so they appeared in agent system prompts and were subject to allow/deny policy — but were absent from builtinToolSeedData(). Since the agent-config UI's allow/deny picker is driven by
  the builtin_tools DB table (seeded by that function), operators had no way to see or toggle these tools. This seeds them so the UI catalog reflects the actual registered tool set.

  Motivation

  The prompt's tool list and the allow/deny UI are fed by two different sources: the prompt lists every tool in the runtime registry, while the UI lists only tools present in the builtin_tools table. These drifted out of sync — operators saw tools like
  datetime, vault_search, delegate, and heartbeat injected into agent prompts with no corresponding UI row to disable them, and were forced to type tool names manually into the deny field as a workaround.

  Changes

  cmd/gateway_builtin_tools.go — add the following entries to builtinToolSeedData():

  ┌────────────────────┬────────────┬───────────────────────────┐
  │        Tool        │  Category  │ Requires (advisory badge) │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ datetime           │ runtime    │ —                         │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ heartbeat          │ scheduling │ —                         │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ memory_expand      │ memory     │ memory                    │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ list_group_members │ messaging  │ —                         │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ zalo_list_groups   │ messaging  │ —                         │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ vault_search       │ vault      │ vault                     │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ vault_read         │ vault      │ vault                     │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ delegate           │ subagents  │ agent_links               │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ workstation_exec   │ runtime    │ workstation               │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ claude_remote      │ runtime    │ workstation               │
  ├────────────────────┼────────────┼───────────────────────────┤
  │ mcp_tool_search    │ skills     │ mcp                       │
  └────────────────────┴────────────┴───────────────────────────┘

  Notes / non-goals

  - telegram_manager intentionally excluded — an existing test (TestBuiltinToolSeedDataDoesNotIncludeTelegramManager) asserts it must not be seeded; it's configured via Telegram channel settings, not exposed as a togglable builtin. Left as-is.
  - Requires is advisory only. It renders as an informational badge in the UI and is not enforced anywhere in the backend — consistent with all existing seed entries (memory, teams, etc. aren't enforced either). Runtime registration conditions are
  unchanged: a tool whose subsystem is absent is simply never registered, so it won't appear in the prompt regardless of its catalog row.
  - stt dead entry left untouched per request (it's in the catalog but has no registered agent tool).

  Deployment

  No migration needed. seedBuiltinTools() runs on every gateway startup with INSERT ... ON CONFLICT (name) DO UPDATE, so a redeploy auto-inserts the new rows on existing databases. Existing rows keep their enabled/settings (operator toggles are
  preserved).

  Surface parity

  - Gateway/seed: ✅ covered.
  - Web UI: ✅ N/A — the UI already renders whatever builtin_tools returns; new rows appear automatically, no component change.
  - API contract / CLI: N/A — no request/response shape or command changes.

  Verification

  - go build ./... ✅
  - go vet ./... ✅
  - go test ./cmd/... ./internal/tools/... ✅

### DIFF
--- a/cmd/gateway_builtin_tools.go
+++ b/cmd/gateway_builtin_tools.go
@@ -26,6 +26,13 @@ func builtinToolSeedData() []store.BuiltinToolDef {
 			Metadata: json.RawMessage(`{"config_hint":"Config → Tools → Exec Approval"}`),
 		},
 		{Name: "wait", DisplayName: "Wait", Description: "Pause the current agent tool sequence for a bounded number of milliseconds", Category: "runtime", Enabled: true},
+		{Name: "datetime", DisplayName: "Date/Time", Description: "Get the current date and time with timezone support, for precise timestamps in scheduling and memory", Category: "runtime", Enabled: true},
+		{Name: "workstation_exec", DisplayName: "Workstation Exec", Description: "Execute an allowlisted command on a linked remote workstation", Category: "runtime", Enabled: true,
+			Requires: []string{"workstation"},
+		},
+		{Name: "claude_remote", DisplayName: "Claude Remote", Description: "Run Claude Code CLI on a remote workstation via workstation_exec", Category: "runtime", Enabled: true,
+			Requires: []string{"workstation"},
+		},
 
 		// web
 		{Name: "web_search", DisplayName: "Web Search", Description: "Search the web for information using a search engine (Brave or DuckDuckGo)", Category: "web", Enabled: true,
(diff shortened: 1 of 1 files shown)


clark-cant (approved):
LGTM — clean seed addition. All 11 entries follow existing pattern, correct categories, advisory Requires badges match runtime registration conditions. telegram_manager exclusion preserved (test-guarded). ON CONFLICT upsert means zero migration cost. Fixes real drift between prompt tool list and UI allow/deny catalog.

PR #1400, opened 2026-07-08
feat(providers): make provider request timeout configurable (default 30s)

bclermont:
The provider verify call ("Check" button) and the provider models-list call used hardcoded timeouts (30s and 15s respectively) that were too low for slow models. This introduces a single tenant-scoped, UI-editable setting —
  providers.request_timeout_sec (default 30) — that bounds both, so operators can raise it for slow models without a redeploy.

  Motivation

  When switching an agent to a slower model, the pre-save validation would time out. The limit was hardcoded in three places (/verify at 30s, /models at 15s ×2) with no way to change it. Operators had no knob short of editing source and rebuilding.

  Changes

  Backend
  - New internal/http/provider_request_timeout.go — loadProviderRequestTimeoutSec(ctx, sc), mirroring the existing loadTenantTTSTimeoutMs pattern. Reads providers.request_timeout_sec from the tenant's system_configs; falls back to
  defaultProviderRequestTimeoutSec (30) on nil store / missing / non-numeric / <=0.
  - provider_verify.go — verify chat call now uses the configurable timeout (was 30s hardcoded).
  - provider_models.go — both models-list timeouts now use the configurable value (were 15s hardcoded).
  - internal/config/config_channels.go — added RequestTimeoutSec to ProvidersConfig.
  - internal/config/config_system.go + cmd/gateway_system_config_sync.go — overlay/seed the value from config.json like tts.timeout_ms.
  - Unit test for the helper (unset / valid / invalid / zero / negative / nil-store).

  Frontend
  - System Settings modal gains a "Provider request timeout (seconds)" number field, wired to providers.request_timeout_sec via the existing generic /v1/system-configs endpoint. Empty = unset (uses default).
  - i18n label + hint added for all 4 locales (en, vi, zh, ko).

  Notes / non-goals

  - No new table or migration — reuses the existing system_configs KV store, tenant-scoped (write gated by the existing admin-role /v1/system-configs handler).
  - The 10s timeout in provider_verify.go was intentionally left alone — it bounds a local Claude CLI auth-status shell check, not a provider network call.
  - Existing deployments pick up the config-seeded default automatically on startup; operator-set values are preserved.

  Surface parity

  - Gateway/handlers: ✅ covered.
  - Web UI: ✅ covered (settings field + i18n).
  - API contract: N/A — reuses the existing /v1/system-configs endpoints; no new/changed request/response shapes.
  - CLI: N/A — no CLI surface for this setting.

  Verification

  - go build ./... ✅
  - go vet ./... ✅
  - go test ./internal/http/... ./internal/config/... ✅
  - npm run lint ✅ (0 errors) · npm run build ✅


clark-cant (approved):
LGTM — clean configurable timeout following the existing loadTenantTTSTimeoutMs pattern.

**Summary**: Replaces hardcoded 30s/15s timeouts in provider verify and models-list with a tenant-scoped system_config (providers.request_timeout_sec, default 30). Operators can now raise the limit for slow models without rebuilding.

**Review**:
- Helper (provider_request_timeout.go) properly guards nil store, empty, non-numeric, zero, and negative values — all fall back to default 30s.
- Test covers all 6 edge cases including nil store.
- Config overlay/seed path mirrors existing TTS timeout pattern exactly.
- Frontend: follows existing InitState/DEFAULTS/load/save pattern. UI field placed logically in Background Workers section.
- i18n complete for all 4 locales (en, vi, zh, ko).
- 10s local CLI auth-status timeout correctly left alone (not a provider network call).
- No migration needed — reuses existing system_configs KV store.
- 13 files, +122/-3 — scope matches description perfectly.

**Verdict**: Approve.

PR #1402, opened 2026-07-09
fix: warm Zalo group-approval cache + attach quoted-image content tag

### LINKED ISSUE #1401: Zalo group forward: approval cache never warms under allowlist policy; quoted images lose their content-tag path
Two follow-up bugs found while live-testing the fixes in #1394/#1395 (only the first two commits of that PR's branch were merged; these two were added afterward and never made it into a PR).

## Bug 1: group-approval cache never warms under `allowlist` group policy

`Send()` in the Zalo Personal channel picks `ThreadTypeGroup` vs `ThreadTypeUser` from an in-memory `approvedGroups` cache, or an explicit `group_id` metadata key on the outbound message. That cache is only ever populated from **inbound** group messages under the `"pairing"` group policy branch. Under `"allowlist"` (a supported, commonly-used policy), that branch never runs, so the cache stays empty forever regardless of chat history.

A cross-target forward whose origin session is a DM (not a group) has no way to set `group_id` metadata either — the tool only sets it when the *origin* context is a group. Combined, a DM→group forward silently addresses the message as a `ThreadTypeUser` send instead of `ThreadTypeGroup`. Zalo's API accepts this without erroring, so the tool reports success and no error is logged anywhere — but nothing arrives.

Fix: `ListGroups()` (added in #1395) now also calls `MarkGroupApproved` for every group it returns, and `Channel.Start()` fires a best-effort async `ListGroups()` call right after connecting so the cache is warm from process start.

## Bug 2: quoted images lose their content-tag path

Replying to an image and asking the agent to forward it fails with "this is a quoted image, no direct file to attach" even though the file was already downloaded to disk. `extractQuoteMedia()` downloads the quoted image and adds it to the `Media` list (used for vision), but the message content text only ever contains the literal placeholder `"[Quoted image]"` — no `<media:image>` tag, unlike a direct (non-quoted) attachment. The later enrichment step that attaches a real file path only works by finding an *existing* bare tag in content; with none present, the downloaded file is reachable for vision but has no path the agent can hand to the forwarding tool.

Fix: a new `buildQuoteMediaTag()` renders the same bare `<media:image>` tag per downloaded quoted image, appended after the quote-wrapper text.

Both fixes are live-verified in production (DM→group forward now delivers; quoted-image forward now attaches the file).

## Testing limitations

Same as noted in #1395: `protocol.FetchGroups`/`downloadFile()` require a live Zalo session / hit `tools.CheckSSRF()`'s loopback block, so full integration coverage isn't feasible without a larger refactor. A pure unit test is included for the quote-tag builder; the cache warm-up is verified via build + clean restart logs (no warm-up failure warning) plus live production confirmation.

Buns0705:
Fixes #1401.

Two commits carried over from the branch behind #1395 that were added after that PR was already merged (only its first two commits landed), so they never went through review:

1. **`ListGroups()` now warms the group-approval cache** (`MarkGroupApproved` per group) and `Channel.Start()` fires a best-effort async `ListGroups()` on connect — fixes DM→group forwards silently addressing the send as a user DM under `allowlist` group policy (no error anywhere, message just never arrives).
2. **`buildQuoteMediaTag()`** renders a bare `<media:image>` tag for each quoted-image download, so the existing content-tag enrichment step has something to attach a real file path to — fixes "no direct file to attach" when forwarding a quoted image.

Both live-verified in production against a real Zalo Personal account.

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./internal/channels/...` clean
- [x] New unit test `TestBuildQuoteMediaTag` (pure, no network)
- [x] Cache warm-up verified via clean restart logs (no warm-up-failure warning)
- [x] Live production: DM→group forward now delivers; quoted-image forward now attaches the file
- [ ] Full integration test not feasible without a live Zalo session / larger SSRF-test refactor (same limitation as #1395)

### DIFF
--- a/internal/channels/zalo/personal/channel.go
+++ b/internal/channels/zalo/personal/channel.go
@@ -96,6 +96,17 @@ func (c *Channel) getListener() *protocol.Listener {
 // to. Lets the agent resolve a group by name (e.g. "Ban Điều Hành") to the
 // chat ID the message tool actually requires, instead of guessing/passing
 // the display name as target.
+//
+// Also warms the approvedGroups cache (BaseChannel.MarkGroupApproved) for
+// every returned group ID. Send() uses that cache to decide whether a
+// target is a group (ThreadTypeGroup) or a user (ThreadTypeUser) when the
+// outbound message carries no explicit "group_id" metadata — which is
+// exactly the case for a message-tool forward whose ORIGIN session isn't
+// itself a group (e.g. forwarding from a DM into a group): the cache is
+// otherwise only populated by inbound group traffic under the "pairing"
+// group policy, so under "allowlist" (this deployment's default) a
+// same-session-never-messaged-before group would default to ThreadTypeUser
+// and the send would go nowhere, with no error surfaced anywhere.
 func (c *Channel) ListGroups(ctx context.Context) ([]channels.GroupInfo, error) {
 	sess := c.session()
 	if sess == nil {
@@ -108,6 +119,7 @@ func (c *Channel) ListGroups(ctx context.Context) ([]channels.GroupInfo, error) 
 	out := make([]channels.GroupInfo, len(groups))
 	for i, g := range groups {
 		out[i] = channels.GroupInfo{GroupID: g.GroupID, Name: g.Name, TotalMember: g.TotalMember}
+		c.MarkGroupApproved(g.GroupID)
 	}
 	return out, nil
 }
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean two-part fix for #1401.

1. **Cache warming**: ListGroups() now calls MarkGroupApproved per group, and Start() fires async ListGroups() on connect. Correctly solves the silent-fail problem where DM→group forwards under allowlist policy would default to ThreadTypeUser and send nowhere. Best-effort async with slog.Warn on failure is the right pattern.

2. **Quote media tag**: buildQuoteMediaTag() renders bare <media:image> tags so the enrichment pipeline (enrichImageIDs/enrichImagePaths) has something to attach real file paths to. Correctly applied in all three quote-handling paths (handleDM, handleGroupMessage, extractContentAndMediaWithQuote).

3. **Tests**: TestBuildQuoteMediaTag covers nil, single, and multiple paths. Good.

4. **Comments**: Excellent documentation explaining WHY — especially the ListGroups doc comment tracing the exact failure mode.

No security concerns, no breaking changes, follows existing patterns.

PR #1403, opened 2026-07-09
feat(bitrix24): add delete affordance for pending portals in channel creation form

tech-synity:
## Summary
- Add a delete icon for **pending** (not-yet-authorized) Bitrix24 portals in the Portal dropdown on the channel-creation form. Installed portals get no delete/edit here (out of scope, deferred to a possible future Settings page).
- Mouse-clickable and keyboard-reachable (Delete/Backspace on the focused option) — a nested `<button>` inside a Radix `SelectItem` isn't reachable via Radix's roving-tabindex keyboard navigation on its own, so an explicit `onKeyDown` handler was added.
- Safe by construction: a pending portal's Select value carries a sentinel prefix (never the bare portal name), so it can never become a channel's stored `config.portal` value — no in-use check needed client-side. Backend `bitrix.portals.delete` RPC keeps its own in-use guard + tenant scope as defense in depth.

## Test plan
- [x] 14 new unit tests (pure-logic, mirrors existing `voice-picker.test.tsx` convention — this project has no `@testing-library/react`)
- [x] Full frontend test suite green, no regressions
- [x] `pnpm build` clean (TypeScript compiles)
- [x] i18n keys added to en/vi/zh, validated as parseable JSON
- [x] Manually verified in browser (mouse click + keyboard Delete) against a live backend — trash icon shows only for pending portals, opens confirm dialog without triggering resume-authorize, cancels cleanly, no console errors
- [x] Two code-review passes (first flagged the keyboard-accessibility gap, second confirmed the fix is correct and robust)

PR #1404, opened 2026-07-09
fix(bitrix24): register newly-created portal in live router

tech-synity:
## Summary
- `bitrix.portals.create` only persisted the new portal row to Postgres — it never registered the portal into the process-wide `bitrix24.Router` used by the OAuth install webhook (`PortalByDomain`/`PortalByKey`). The only other `RegisterPortal` call site is `BootstrapPortals`, which runs once at gateway boot.
- Result: any portal created via the self-service UI while the gateway was already running (i.e. basically always, in production) was invisible to the install callback until the next full restart — the install attempt 404s with `"unknown portal"`. Hit live on a real portal (`web1trang.bitrix24.com`) during manual QA of #1403.
- Fix: inject `registerPortal`/`unregisterPortal` func fields into `BitrixPortalsMethods` (mirrors the existing `gatewayPublicURL func() string` pattern already on the struct), defaulting to `bitrix24.WebhookRouter()` + `NewPortal`/`RegisterPortal` — the same construction path `BootstrapPortals` uses — and `Router.UnregisterPortal` respectively. `handleCreate` calls `registerPortal` after a successful `Create`; `handleDelete` calls `unregisterPortal` after a successful `Delete`, for symmetry. Both are best-effort: a registration failure is logged (`slog.Warn`), not returned as an RPC error, since the DB row is valid either way and `BootstrapPortals` will eventually pick it up on the next restart.
- Injected via func fields (rather than calling `bitrix24.WebhookRouter()` inline) so tests can stub live-router registration without touching that process-wide singleton — its test-reset helper (`resetWebhookRouterForTest`) isn't exported outside the `bitrix24` package.

## Test plan
- [x] 4 new test cases: register-on-create, best-effort failure handling (create still succeeds + row persisted even if registration fails), unregister-on-delete, no-unregister-when-delete-is-blocked (in-use guard)
- [x] All 20 tests in `bitrix_portals_test.go` pass (15 existing + 4 new)
- [x] `go build ./...` and `go build -tags sqliteonly ./...` clean
- [x] `go vet ./internal/gateway/methods/... ./internal/channels/bitrix24/...` clean
- [x] Manually confirmed via live Docker logs that a restart's `BootstrapPortals` correctly picks up a portal that had been created mid-uptime (the workaround before this fix), and that the fix's `registerPortal` closure follows the identical `NewPortal` + `RegisterPortal` construction path
- [x] Independent code-review pass: DONE, no blocking issues

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -788,6 +788,7 @@ func runGateway() {
 				pgStores.BitrixPortals,
 				pgStores.ChannelInstances,
 				server.PublicURLSnapshot().Get,
+				bitrixEncKey,
 			).Register(server.Router())
 		}
 
--- a/internal/gateway/methods/bitrix_portals.go
+++ b/internal/gateway/methods/bitrix_portals.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/google/uuid"
 
+	"github.com/nextlevelbuilder/goclaw/internal/channels/bitrix24"
 	"github.com/nextlevelbuilder/goclaw/internal/gateway"
 	"github.com/nextlevelbuilder/goclaw/internal/i18n"
 	"github.com/nextlevelbuilder/goclaw/internal/permissions"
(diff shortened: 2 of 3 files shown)


clark-cant (approved):
LGTM — clean fix for a real production bug.

1. **Root cause is correct**: `handleCreate` only persisted the DB row but never registered into the process-wide `bitrix24.Router`. The only other `RegisterPortal` call site is `BootstrapPortals` (boot-only). So a freshly created portal's install callback 404s until restart. Confirmed by the live repro (web1trang.bitrix24.com).

2. **Fix pattern is sound**: Injecting `registerPortal`/`unregisterPortal` as func fields mirrors the existing `gatewayPublicURL func()` pattern on the struct. Defaults use the same `NewPortal` + `RegisterPortal` construction path as `BootstrapPortals` — verified by the test plan.

3. **Best-effort semantics are correct**: Registration failure is logged (`slog.Warn`) but not returned as RPC error. The DB row is valid either way, and `BootstrapPortals` will pick it up on restart. The warning message is specific enough for ops to diagnose.

4. **Symmetric unregister on delete**: Prevents stale in-memory routing after DB row deletion. Correctly placed after successful delete, and the test confirms blocked deletes don't trigger unregister.

5. **Test coverage**: 4 new tests covering register-on-create, best-effort failure (create still succeeds), unregister-on-delete, and no-unregister-when-delete-blocked. All 20 tests pass.

6. **`encKey` parameter addition**: Needed to construct `*bitrix24.Portal` for live router registration. Clean propagation from `cmd/gateway.go`. All existing test call sites updated consistently.

No issues found. Ship it.

PR #1405, opened 2026-07-09
fix(config): remove ignored session scoping controls

namnhfreelancer:
## Summary
- Remove the deprecated Session Scoping controls from the Config > Behavior UI.
- Stop sending `sessions.scope` / `sessions.dm_scope` from the Behavior save payload.
- Remove stale i18n keys and add a regression test for the Behavior patch payload.

## Bug
The UI exposed `Session Scope` and `DM Scope` as editable settings, but the gateway has ignored non-default values since `651072a9e`, logging them as deprecated and fixed to `per-sender` / `per-channel-peer`. This made the UI look configurable while runtime behavior never changed.

In practice, an operator could change Session Scoping in `/config`, save successfully, and still get the same session partitioning behavior. That is a UI contract bug: the dashboard advertised a setting that the backend no longer honors.

## Verification
- `cd ui/web && pnpm test -- behavior-section.test.ts` (52 files / 340 tests passed)
- `cd ui/web && pnpm lint` (0 errors; 4 unrelated existing warnings)
- `cd ui/web && pnpm build`

## Surface parity
- Gateway server: N/A because backend already deprecates and ignores these fields.
- API contract: N/A because this removes stale UI payload fields without changing backend contracts.
- Web UI: updated `/config` Behavior tab and i18n keys.
- CLI/runtime package: N/A because this was a dashboard-only stale control.


clark-cant (approved):
LGTM — clean UI contract fix.

**Summary**: Removes deprecated Session Scoping controls from the Config > Behavior UI. The gateway has ignored these fields since 651072a9e, so the UI was advertising settings that had no effect.

**Review**:
1. Root cause correct: UI exposed `sessions.scope` / `sessions.dm_scope` as editable but gateway hardcoded them to `per-sender` / `per-channel-peer`. This is a UI contract bug — operator changes were silently ignored.
2. Clean removal: `BehaviorSessionsCard` deleted entirely, i18n keys removed across all 4 locales (en/ko/vi/zh), state management removed from `BehaviorSection`.
3. Good refactor: `buildBehaviorPatch()` extracted as a pure function — makes the patch payload testable and prevents accidental re-introduction of stale fields.
4. Regression test: Verifies `sessions` key is absent from patch payload and validates the full expected structure. 340 tests pass.
5. No backend changes needed — gateway already deprecates and ignores these fields.

**Risk**: Low — UI-only removal of dead controls. No API contract changes, no backend impact.

PR #1409, opened 2026-07-09
fix(graph): preserve knowledge graph node colors

namnhfreelancer:
## Summary
- Preserve adapter-assigned colors for Knowledge Graph entity nodes in the shared Sigma graph renderer.
- Keep Vault document nodes using the existing theme-aware Vault color mapping.
- Add regression coverage for KG node colors and Vault node color behavior.

## Related issues
- None found for `knowledge graph color`.

## Test plan
- [x] `cd ui/web && pnpm test -- src/components/graph/sigma-graph-container.test.ts`
- [x] `cd ui/web && pnpm build`

### DIFF
--- a/ui/web/src/components/graph/graph-node-colors.ts
+++ b/ui/web/src/components/graph/graph-node-colors.ts
@@ -1,0 +1,11 @@
+import { getVaultNodeColor } from "@/adapters/vault-graph-adapter";
+
+export function resolveGraphNodeDisplayColor(attrs: Record<string, unknown>, isDark: boolean): string {
+  const docType = typeof attrs.docType === "string" ? attrs.docType : "";
+  if (docType) {
+    return getVaultNodeColor(docType, isDark);
+  }
+  return typeof attrs.color === "string" && attrs.color
+    ? attrs.color
+    : getVaultNodeColor("other", isDark);
+}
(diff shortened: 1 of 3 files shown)


clark-cant (approved):
LGTM — clean fix for KG node color preservation.

1. **Root cause correct**: `nodeReducer` was unconditionally overwriting all node colors via `getVaultNodeColor(docType)`, which clobbered adapter-assigned colors on KG entity nodes. KG adapter sets per-entityType colors that were silently discarded.
2. **Clean extraction**: `resolveGraphNodeDisplayColor()` in new `graph-node-colors.ts` correctly bifurcates — Vault document nodes (have `docType`) get theme-aware Vault colors; KG entity nodes (no `docType`, have `color` from adapter) preserve their adapter-assigned color.
3. **Fallback sound**: Missing `color` attribute falls back to `getVaultNodeColor("other", isDark)` — consistent with existing behavior for unknown node types.
4. **Test coverage**: 2 regression tests — one verifying KG entity color preservation (`#E85D24` passes through), one verifying Vault document nodes still get theme-mapped colors (`#8b5cf6` → `#0891b2` for teal document type).
5. **No breaking changes**: Only the `nodeReducer` call site changes; all downstream rendering code is unaffected.
6. **3 files, +32/-2**: Minimal surface area, well-scoped.

PR #1413, opened 2026-07-09
fix(kg): count newly inserted dedup candidates

namnhfreelancer:
## Summary
- count KG dedup candidates only when `ON CONFLICT DO NOTHING` actually inserts a row
- apply the same behavior to PostgreSQL and SQLite stores
- add PG integration and SQLite unit coverage for dismissed/pre-existing candidate conflicts

## Validation
- `go test -tags sqliteonly ./internal/store/sqlitestore -run TestSQLiteKGScanDuplicatesCountsOnlyInsertedCandidates -count=1`
- `go test ./internal/store/pg -run TestDoesNotExist -count=1`
- `TEST_DATABASE_URL="${TEST_DATABASE_URL:-postgres://postgres:test@localhost:5433/goclaw_test?sslmode=disable}" go test -tags integration ./tests/integration -run TestStoreKG_ScanDuplicatesCountsOnlyInsertedCandidates -count=1 -v` (test package passed; target PG test skipped locally because localhost:5433 rejected postgres/test auth)

Surface parity: Web UI/API/CLI N/A because this only corrects store-layer return counts for KG dedup scan/flag operations.

### DIFF
--- a/internal/store/knowledge_graph_store.go
+++ b/internal/store/knowledge_graph_store.go
@@ -96,7 +96,7 @@ type KnowledgeGraphStore interface {
 	// Auto-merges at high similarity (>0.98 + name match), flags medium (>0.90) as candidates.
 	DedupAfterExtraction(ctx context.Context, agentID, userID string, newEntityIDs []string) (merged int, flagged int, err error)
 	// ScanDuplicates scans ALL entities with embeddings for duplicates (self-join).
-	// Flags candidates above threshold. Used for on-demand bulk scanning of existing data.
+	// Returns the number of newly inserted review candidates above threshold.
 	ScanDuplicates(ctx context.Context, agentID, userID string, threshold float64, limit int) (int, error)
 	// ListDedupCandidates returns pending dedup candidates for review.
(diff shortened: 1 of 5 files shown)


clark-cant (approved):
LGTM — clean fix for inflated dedup candidate counts.

1. Root cause correct: `ON CONFLICT DO NOTHING` silently skips pre-existing rows, but the old code incremented the counter regardless. `ScanDuplicates` and `DedupAfterExtraction` both reported inflated counts.
2. Fix is minimal and correct: `insertDedupCandidate` now returns `(bool, error)` using `RowsAffected()`. Callers only increment when `inserted == true`.
3. PG/SQLite parity: Identical pattern in both stores — `res.RowsAffected() > 0` check.
4. Interface doc updated: ScanDuplicates comment now accurately says "Returns the number of newly inserted review candidates".
5. Test coverage: Both SQLite unit test and PG integration test verify the exact semantic — seed dismissed candidate → expect 0, delete → expect 1.
6. No breaking changes: Return value is more accurate (lower but correct). Purely store-layer, no UI/API/CLI impact.

PR #1415, opened 2026-07-09
tracing: include tool calls and hook executions in traces; fix usage_events FK drop

bclermont:
Traces are now useful for troubleshooting agent runs: each model-call span visibly contains the tool calls it triggered, and every hook execution (pre/post tool use, session start/stop, etc.) now produces a trace span with its input, output, decision,
  and error. Also fixes a pre-existing race that was silently dropping usage_events (token/cost) rows.

  Motivation

  Every model call already created a trace, but the trace didn't contain the tool calls, and hook execution detail lived only in the hook_executions audit table — invisible when investigating what an agent actually did. The goal is first-class
  troubleshooting of agent + tool execution via the trace.

  Changes

  1. Tool calls nested under their producing LLM call

  - Added RunState.CurrentLLMSpanID / CurrentToolSpanID and threaded them through the pipeline.
  - tool_call spans now parent to the llm_call span that produced them. pre_tool_use hook spans parent to the llm_call; post_tool_use hook spans parent to the tool_call.
  - Resulting span tree per turn:
  agent
  └─ llm_call
     ├─ hook.<handler>.pre_tool_use
     ├─ tool_call
     │  └─ hook.<handler>.post_tool_use
     └─ (next llm_call …)
  - (Parallel-path post-tool-use hooks nest under the LLM span, because the per-tool span ID isn't reachable from the pipeline package.)

  2. Hook executions now emit trace spans

  - EmitHookSpan (internal/hooks/tracing.go) — previously defined + tested but never called in production — is now wired into dispatcher.go's writeExec (the single chokepoint for sync + async hook fires).
  - Each hook span carries: input (the hook Event JSON), console output, decision, error, duration_ms, and hook_id/hook_name/handler_type in Metadata.
  - No-op when tracing is disabled. Reuses SpanTypeEvent — no schema change.

  3. Fix usage_events_span_id_fkey violations (pre-existing)

  - Root cause: recordToolUsageEvent inserted a usage_events row synchronously referencing the tool span's ID, but that span was only buffered and flushed ~5s later by the collector → FK failed for fast tools → token/cost data silently dropped
  (intermittent).
  - Fix: usage events are routed through tracing.Collector.EmitUsageEvent and flushed after BatchCreateSpans in the same flush cycle, so the referenced span row always exists first. FK is preserved; no migration.

  Testing

  - go build ./..., go build -tags sqliteonly ./..., go vet ./... pass.
  - go test ./internal/tracing/... ./internal/agent/... ./internal/store/... pass. (tests/integration compile under the integration tag; not run here — needs pgvector on :5433.)
  - Web trace viewer renders spans purely by parent_span_id with no span_type filter, so event spans and the new nesting display correctly.

  Notes

  - No database migration; store.SpanData already supported all fields used.
  - Surface parity: Gateway ✓; API contract N/A (reused SpanData); Web UI N/A (data layer); Subagent traces N/A (already nest llm→tool under their own agent span); CLI N/A.

### DIFF
--- a/cmd/gateway_setup.go
+++ b/cmd/gateway_setup.go
@@ -303,7 +303,7 @@ func wireTracingAndCron(
 ) (*tracing.Collector, *tracing.SnapshotWorker) {
 	var traceCollector *tracing.Collector
 	if stores.Tracing != nil {
-		traceCollector = tracing.NewCollector(stores.Tracing)
+		traceCollector = tracing.NewCollector(stores.Tracing, stores.UsageEvents)
 		traceCollector.OnFlush = func(traceIDs []uuid.UUID) {
 			ids := make([]string, len(traceIDs))
 			for i, id := range traceIDs {
--- a/internal/agent/loop_pipeline_callbacks.go
+++ b/internal/agent/loop_pipeline_callbacks.go
@@ -391,6 +391,9 @@ func (l *Loop) makeCallLLM(req *RunRequest, emitRun func(AgentEvent)) func(ctx c
 			opts = append(opts, withProvider(provider.Name()))
 		}
 		spanID := l.emitLLMSpanStart(ctx, start, state.Iteration+1, chatReq.Messages, opts...)
+		if spanID != uuid.Nil {
+			state.CurrentLLMSpanID = &spanID
+		}
 		recordUsageCapAttempt := func(reservation *usagecaps.Reservation) {
 			if reservation != nil {
 				opts = append(opts, withUsageCapMetadata(reservation.TraceMetadata()))
--- a/internal/agent/loop_pipeline_tool_callbacks.go
+++ b/internal/agent/loop_pipeline_tool_callbacks.go
@@ -13,6 +13,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/providers"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
 	"github.com/nextlevelbuilder/goclaw/internal/tools"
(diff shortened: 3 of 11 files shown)


clark-cant (approved):
LGTM — clean tracing improvement with a real bug fix.

1. **Span tree nesting correct**: RunState.CurrentLLMSpanID/CurrentToolSpanID threaded through pipeline. Tool spans parent to LLM span, pre_tool_use hooks parent to LLM span, post_tool_use hooks parent to tool span. Resulting tree is exactly what you'd want for troubleshooting.

2. **Hook span emission wired correctly**: EmitHookSpan was defined+tested but never called in production. Now wired into dispatcher.go's writeExec — the single chokepoint for sync+async hook fires. Each span carries input, output, decision, error, duration_ms, hook identity. No-op when tracing disabled.

3. **usage_events FK fix is the right approach**: Root cause correct — usage events were inserted synchronously referencing a span flushed ~5s later, so FK failed for fast tools. Routing through collector and flushing AFTER spans in same cycle preserves the FK without schema migration. Clean ordering fix.

4. **EmitHookSpan signature change is clean**: Caller-measures-duration pattern avoids double-measurement. cfg/input/output params give the span full context. Tenant fallback to MasterTenantID is correct.

5. **Test coverage**: Unit tests updated for new signature. Integration test updated. Existing tracing tests pass.

6. **No breaking changes**: Reuses existing SpanData fields. No migration. Backward compatible.

One minor note: parallel-path post-tool-use hooks nesting under LLM span (rather than per-tool span) is documented and correct given the architectural constraint.

Squash-merge recommended.

PR #1417, opened 2026-07-10
feat(bitrix24): ship per-user OAuth re-authorization flow

tech-synity:
## What changed

This PR ships the Bitrix24 per-user OAuth re-authorization flow and the follow-up fixes needed to make it work in the current dev setup.

- add per-user OAuth re-auth flow for Bitrix24 users
- send authorize links from the bot and complete MCP credential minting on callback
- keep the callback compatible when the Bitrix app still redirects to `/bitrix24/install`
- fold in review-feedback fixes around the OAuth flow

## Why

When a Bitrix24 user had no usable per-user MCP credentials, the bot needed a way to ask that specific user to authorize and then mint or refresh credentials without reinstalling the whole app.

In dev we also hit a compatibility problem: some Bitrix app settings still redirected the browser back to `/bitrix24/install`, while the newer flow expected `/bitrix24/handler`. That produced `invalid state format` even though the signed OAuth state itself was valid.

## Impact

- users can re-authorize from the bot flow instead of relying on portal-wide reinstall steps
- MCP per-user credentials can be minted/refreshed after the callback completes
- legacy dev app settings that still point to `/bitrix24/install` continue to work

## Root cause

The old install handler only understood install-state in the `<tenant>:<portal>` shape. The newer user OAuth flow signs a different state payload. When Bitrix redirected that newer payload back to the legacy install URL, the request was treated as a broken install callback instead of a user OAuth callback.

## Validation

- `go test ./internal/channels/bitrix24/...`
- live-verified on dev by sending the authorize link to Bitrix24 user `7`
- observed successful callback handling and persisted credential refresh in DB/logs

### DIFF
--- a/docs/journals/260708-2003-oauth-per-user-flow-planning.md
+++ b/docs/journals/260708-2003-oauth-per-user-flow-planning.md
@@ -1,0 +1,87 @@
+# Bitrix24 Per-User OAuth Flow: Planning Complete
+
+**Date**: 2026-07-08 20:03
+**Severity**: High
+**Component**: Bitrix24 channel, OAuth, provisioning, MCP credential minting
+**Status**: Pending Implementation (6-phase plan drafted, all decisions locked)
+
+## What Happened
+
+Bitrix24 support confirmed via official investigation (2026-07-08) that the `ONIMBOTMESSAGEADD` webhook no longer attaches `auth[access_token]` for new users — this is **expected behavior, not a regression**. Imbot subscriptions bind with `USER_ID=0` at registration time, so Bitrix never requests/includes a real OAuth token in the top-level `auth` block; only the bot's own token (`data[BOT][botID][AUTH]`) is guaranteed present.
+
+**Proof**: goclaw's `mcp_user_credentials` table shows 3 users (614, 1, 610) successfully minted credentials between 2026-05-07 and 2026-05-25. Zero new users since. DB + support confirmation align perfectly.
+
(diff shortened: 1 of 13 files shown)


clark-cant (approved):
LGTM — clean per-user OAuth re-auth flow for Bitrix24.

**Review summary:**

1. **Design quality**: Stateless HMAC-signed state with 10-min TTL is the right call — eliminates DB janitor needs and has tight blast radius. Planning doc shows good iteration (v1→v2 trigger precision, DM vs group delivery debate).

2. **Correctness**: 
   - Identity validation (payload.UserID == token.User.ID) prevents account takeover
   - Domain validation prevents redirect to wrong portal
   - Debounce sweep on acquire prevents unbounded map growth
   - Fallback chain (DM → group → release debounce on total failure) is sound

3. **Security**: HMAC-SHA256, no secrets in logs, identity + domain checks, stateless codec. All correct.

4. **Test coverage**: 996 lines of tests across 4 test files. Meaningful assertions (not phantom coverage). Edge cases covered: DM send fails, total failure releases debounce, expired entries swept, identity mismatch rejected.

5. **No breaking changes**: Purely additive. provisionIfMissing signature change is internal to bitrix24 package.

6. **Strategic value**: Fixes hard block on Bitrix24 multi-user expansion. No new tables, no migration, ~230 LOC production code as estimated in planning doc.

7. **Legacy compatibility**: /bitrix24/install now handles user OAuth state too — good catch on the dev environment redirect mismatch.

PR #1423, opened 2026-07-11
fix: honor model override for sync subagents

zezaeoh:
## Summary
Fix synchronous `spawn` calls silently ignoring their per-task `model` override. The sync path now passes the requested model into `SubagentTask`, matching the existing async behavior and preserving the documented model precedence.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration)

The focused regression test passes with the race detector. A full `go test -race ./...` run was also allowed to complete; it failed in pre-existing long-running tests outside this change: `internal/hooks/handlers` exceeded 11 minutes, and `internal/providers` timed out after 10 minutes in `TestOpenAIEmbedding_BatchSplitting`. The changed `internal/tools` package passed.

## Test Plan
- `go build ./...`
- `go vet ./...`
- `go test ./internal/tools -run TestRunSyncHonorsPerTaskModelOverride -count=1`
- `go test -race ./internal/tools -run TestRunSyncHonorsPerTaskModelOverride -count=1`
- `go test -race ./...` — completed with unrelated timeouts in `internal/hooks/handlers` and `internal/providers`; `internal/tools` passed
- Added `TestRunSyncHonorsPerTaskModelOverride`, which configures conflicting manager, parent, and subagent-config models and verifies that the explicit per-task model reaches the provider.

### DIFF
--- a/internal/tools/subagent_spawn.go
+++ b/internal/tools/subagent_spawn.go
@@ -127,7 +127,7 @@ func (sm *SubagentManager) RunSync(
 	ctx context.Context,
 	parentID string,
 	depth int,
-	task, label string,
+	task, label, modelOverride string,
 	channel, chatID string,
 ) (string, int, error) {
 	cfg := sm.effectiveConfig(ctx)
@@ -160,6 +160,7 @@ func (sm *SubagentManager) RunSync(
 		Label:            label,
 		Status:           "running",
 		Depth:            depth + 1,
+		Model:            modelOverride,
 		OriginChannel:    channel,
 		OriginChatID:     chatID,
 		OriginLocalKey:   ToolLocalKeyFromCtx(ctx),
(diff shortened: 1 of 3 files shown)


clark-cant (approved):
LGTM — clean, minimal fix.

1. Correct bug: sync spawn path was silently dropping the per-task model override. Now `modelOverride` flows through `RunSync` → `SubagentTask.Model`, matching async behavior.
2. Call site in `executeSubagentSync` correctly extracts `model` from args.
3. Test is well-designed: configures conflicting manager/parent/config models and verifies the explicit per-task override wins. Good use of recording provider to assert the actual model reaching the provider.
4. No breaking changes: additive parameter, backward compatible.
5. Small scope, clean code. Ship it.

PR #1424, opened 2026-07-11
fix: persist explicit None TTS provider instead of dropping it

### LINKED ISSUE #1080: [Bug] TTS provider cannot be switched back to "None (Disabled)" — setting not persisted after save
## Bug Report

**Version:** v3.11.3 (Docker Compose)

### Description

When the TTS provider is set to **Edge (Free)**, it is impossible to switch it back to **None (Disabled)**. Clicking "Save Changes" appears to succeed (no error shown), but after reloading the page, the setting reverts to Edge (Free) instead of reflecting the saved change.

Changes made **within** Edge (Free) settings (e.g., voice, rate, pitch, volume) are saved correctly. The issue only occurs when attempting to change the **primary provider** selection to `None (Disabled)`.

### Steps to Reproduce

1. Open TTS settings page
2. Ensure current provider is set to **Edge (Free)**
3. Click **None (Disabled)**
4. Click **Save Changes** (success message or no error appears)
5. Reload the page

### Expected Behavior

The provider should be saved as `None (Disabled)` and TTS should be fully disabled after reload.

### Actual Behavior

After reload, the provider reverts to **Edge (Free)**. The `None (Disabled)` selection is not persisted.

### Environment

- GoClaw version: **v3.11.3**
- Deployment: **Docker Compose**
- TTS Provider currently active: **Edge (Free)**

### Additional Notes

- No error is shown when saving — the save action appears to complete successfully
- Only the provider field fails to persist; other fields within Edge (Free) save correctly
- This may indicate the save handler skips persisting the provider when the selected value is `None`/empty/falsy (treating it as "no change")

mvanhorn:
## Summary
Setting the TTS provider to None (empty/explicit no-provider) now persists correctly instead of being silently dropped, so a user who disables TTS keeps that setting after save.

## Type
- [x] Bug fix

## Target Branch
Targets `dev` (default for bugfixes).

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Web UI builds: `cd ui/web && pnpm build` (if UI changes) — N/A, no UI change
- [x] New user-facing strings added to all 3 locales (en/vi/zh) — N/A, no new strings
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration) — N/A, no migration

## Test Plan
Reported in #1080: the TTS config update path in `internal/http/tts_config.go` guarded the provider write with `if req.Provider != ""`, so an explicit empty/None provider was never written and the previous value stuck. The field is now a `*string` pointer so an explicitly-provided empty value is distinguished from "field omitted" and persisted. Added a regression test in `internal/http/tts_config_test.go` asserting that submitting an empty provider clears the stored value, while omitting the field leaves it unchanged. `go test -race ./internal/http/...` passes.

Fixes #1080

### DIFF
--- a/internal/http/tts_config.go
+++ b/internal/http/tts_config.go
@@ -220,7 +220,7 @@ func (h *TTSConfigHandler) handleGet(w http.ResponseWriter, r *http.Request) {
 
 // ttsConfigSaveRequest is the request body for POST /v1/tts/config.
 type ttsConfigSaveRequest struct {
-	Provider   string                  `json:"provider"`
+	Provider   *string                 `json:"provider"`
 	Auto       string                  `json:"auto"`
 	Mode       string                  `json:"mode"`
 	MaxLength  int                     `json:"max_length"`
@@ -269,7 +269,7 @@ func (h *TTSConfigHandler) handleSave(w http.ResponseWriter, r *http.Request) {
 		set := func(key, val, label string) bool {
 			return saveOrFail(w, ctx, h.systemConfigs.Set, key, val, label)
 		}
-		if req.Provider != "" && !set("tts.provider", req.Provider, "provider") {
+		if req.Provider != nil && !set("tts.provider", *req.Provider, "provider") {
 			return
 		}
 		if req.Auto != "" && !set("tts.auto", req.Auto, "auto") {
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
## Review — github-maintain

**Summary**: Fixes #1080 by changing `Provider string` to `*string` pointer in `ttsConfigSaveRequest`, allowing explicit empty/None provider to be distinguished from omitted field.

**Risk level**: Low — small, targeted fix with proper regression tests.

**Mandatory gates**:
- Duplicate/prior implementation: Clear — no prior fix for this specific issue
- Project standards: Follows existing patterns in `internal/http/tts_config.go`
- Strategic necessity: Clear value — fixes real user-facing bug where disabling TTS doesn't persist

**Findings**:
- **Correctness**: ✓ Pointer semantics correctly handle nil vs empty string
- **Testing**: ✓ Three subtests cover: explicit empty (disables), omitted (preserves), non-empty (replaces)
- **Security**: ✓ No new attack surface; parameterized queries unchanged
- **Breaking changes**: ✓ None — JSON unmarshaling handles both old clients (sending string) and new semantics

**Verdict**: **Approve**

Clean fix with proper test coverage. The pointer pattern is the correct Go idiom for distinguishing absent vs empty in JSON APIs.

*Posted by github-maintain at $(date -u +"%Y-%m-%dT%H:%M:%SZ")*

PR #1425, opened 2026-07-11
fix(bitrix24): normalize ws/wss scheme in public URL detection

tech-synity:
## Summary
- The gateway's public-URL snapshot (used to build the Bitrix24 OAuth install link) trusted `X-Forwarded-Proto` verbatim. When a WebSocket upgrade request (`/ws`) passes through a proxy that sets `X-Forwarded-Proto: wss`, the snapshot got poisoned with scheme `wss://`, producing an unreachable install URL (`ERR_UNKNOWN_URL_SCHEME` in the browser).
- Same bug duplicated in `internal/channels/bitrix24/public_url.go` (used directly by the install handler to persist the portal's public URL).
- Fix: normalize `ws`/`wss` to `http`/`https` (same transport, just describes a WebSocket upgrade) instead of trusting the header value as-is; unrecognized values fall back to the TLS-derived default as before.

## Test plan
- [x] `go build ./internal/gateway/... ./internal/channels/bitrix24/...`
- [x] `go test ./internal/gateway/... ./internal/channels/bitrix24/...` — includes new `websocket_upgrade_wss_normalizes_to_https` / `websocket_upgrade_ws_normalizes_to_http` cases in both packages' existing table-driven tests
- [x] Verified live on a tose.sh deployment: Bitrix24 portal authorize link now resolves to `https://` instead of `wss://`

### DIFF
--- a/internal/channels/bitrix24/public_url.go
+++ b/internal/channels/bitrix24/public_url.go
@@ -24,7 +24,10 @@ var (
 // the URL Bitrix24 will use for all subsequent event callbacks.
 //
 // Scheme resolution priority:
-//  1. X-Forwarded-Proto header (set by Cloudflare Tunnel / nginx)
+//  1. X-Forwarded-Proto header (set by Cloudflare Tunnel / nginx), normalized:
+//     "http"/"https" pass through as-is; "ws"/"wss" map to "http"/"https"
+//     (same transport, describes a WebSocket upgrade rather than a scheme
+//     change); any other value is ignored.
 //  2. r.TLS != nil → https
 //  3. Otherwise → http (honest to what we observed). If a reverse proxy is
 //     terminating TLS without forwarding the proto header, fix the proxy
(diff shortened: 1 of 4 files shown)


clark-cant (approved):
LGTM — clean, correct bug fix.

1. Bug is real: X-Forwarded-Proto: wss from WebSocket upgrade poisons public URL with wss:// scheme → ERR_UNKNOWN_URL_SCHEME in browser.
2. Fix is correct: normalizes ws→http, wss→https (same transport, just describes WebSocket upgrade).
3. Applied consistently in both derivePublicURL locations (bitrix24 public_url.go + gateway public_url_snapshot.go).
4. Logic restructure is sound: TLS-derived default first, then header override only for recognized schemes. Unknown values fall through to TLS default — strictly better than blindly trusting any header value.
5. Tests cover both ws and wss cases in both packages.
6. No breaking changes: existing http/https behavior preserved.
7. No security concerns — purely additive URL scheme normalization.
8. Scope appropriate: 4 files, +54/-7, focused fix.
9. PR rebased against dev, all CI green.

PR #819, opened 2026-04-10
feat(whatsapp): add slash command handling and text feedback

reski-rukmantiyo:
## Summary
- Add WhatsApp slash commands: `/reset`, `/stop`, `/stopall`, `/menu`
- Commands are intercepted before the normal message pipeline in `inbound.go`
- New `commands.go` handles routing and publishes appropriate `InboundMessage` events via the bus
- New `sendText()` helper in `outbound.go` for direct plain-text replies (used for `/menu` and `/reset` confirmation)
- `/stop` and `/stopall` delegate feedback to the consumer after cancel result is known

## Files changed
- `internal/channels/whatsapp/commands.go` (new) — command handler with menu help text
- `internal/channels/whatsapp/inbound.go` — command interception before media/pipeline processing
- `internal/channels/whatsapp/outbound.go` — `sendText()` helper for direct text delivery

## Test plan
- [ ] Send `/menu` via WhatsApp and verify help text is returned
- [ ] Send `/reset` and confirm conversation history is cleared with confirmation message
- [ ] Send `/stop` during an active task and verify task cancellation
- [ ] Send `/stopall` with multiple active tasks and verify all are cancelled
- [ ] Send `/unknown` and verify it passes through to the normal agent pipeline (not intercepted)
- [ ] Send a regular message (no `/`) and verify it reaches the agent normally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1426, opened 2026-07-11
fix(whatsapp): add missing protobuf import breaking dev build

tech-synity:
## What changed

Adds the missing `google.golang.org/protobuf/proto` import in `internal/channels/whatsapp/outbound.go`.

## Why

` sendText ` calls ` proto.String(text) ` (added in #819) without importing the ` proto ` package, breaking ` go build ./... ` on ` dev ` since that merge. This currently fails CI on every open PR against ` dev ` (confirmed on #1425's ` go ` check and on ` dev `'s own CI run).

## Validation

- ` go build ./... ` and ` go build -tags sqliteonly ./... ` — both clean
- ` go vet ./internal/channels/whatsapp/... ` — clean
- ` go test ./internal/channels/whatsapp/... ` — pass

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -218,5 +218,5 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20260128011058-8636f8732409 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260128011058-8636f8732409 // indirect
 	google.golang.org/grpc v1.78.0 // indirect
-	google.golang.org/protobuf v1.36.11 // indirect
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
LGTM — correct build fix.

#819 introduced proto.String(text) in outbound.go without importing the proto package, breaking dev build. This PR:

1. Adds the missing google.golang.org/protobuf/proto import in outbound.go
2. Promotes protobuf from indirect to direct dependency in go.mod (correct since it's now directly imported)

Minimal, focused, fixes CI for all open PRs against dev. No concerns.

PR #1429, opened 2026-07-12
feat(dingtalk): DingTalk channel (Stream mode + AI Card streaming)

connermo:
## Overview

Adds `dingtalk` as a first-class GoClaw channel type. Inbound via DingTalk **Stream mode** (WebSocket — no public IP, callback domain, or ICP filing required), outbound via **AI Card streaming** in both groups and DMs with plain text / markdown fallback.

Ported from the official [`dingtalk-openclaw-connector`](https://github.com/DingTalk-Real-AI/dingtalk-openclaw-connector) (TypeScript, for OpenClaw) — used purely as a protocol spec, since its config schema is already ~1:1 with GoClaw's channel config (both descend from OpenClaw). Transport uses the official `github.com/open-dingtalk/dingtalk-stream-sdk-go` (diverging from Feishu's hand-rolled WS codec, since DingTalk ships a maintained Go SDK).

Plan: `plans/260709-2230-dingtalk-channel/plan.md` — all 7 phases complete, with an adversarial audit pass (`reports/`).

## What's included

- **Phase 1–2:** Channel type registration + factory, Stream transport, dual-host API client
- **Phase 3:** Inbound path with ack-then-process
- **Phase 4–5:** Outbound text/markdown, inbound + outbound media
- **Phase 6:** AI Card streaming (posts on first token, configurable repaint interval, thinking reaction on the user's message)
- **Phase 7:** Cross-surface parity (docs, web UI channel schema/config) + live verification
- Slash commands: `/new`, `/stop`, `/stopall` with a fail-closed group gate
- Pins the Stream SDK past a process-killing panic in v0.9.1

## Surface parity

- **Gateway:** new `internal/channels/dingtalk/` package, registered in factory + capabilities
- **Web UI:** channel schema, constants, config bindings updated (`ui/web/src/pages/channels/`, `constants/channels.ts`)
- **Docs:** `docs/05-channels-messaging.md` updated with DingTalk setup

## Testing

Extensive unit coverage across the package (card, client, commands, config, emotion, factory, inbound, media, outbound, parse). Live-verified against a real DingTalk app (see phase-07 report).


feixiang-lex:
# DingTalk Channel PR Review

Hi Conner — thanks for the substantial work here. Overall, this is a thoughtful implementation: the fail-closed permission checks, ACK-then-process flow, deduplication, config validation, SDK pin, and test coverage are all especially solid.

I did find several issues that I think should be addressed before merging:

## Blocking issues

### 1. [Security] Secrets may leak into logs

`oapiToken()` puts `appsecret` in the query string, and `doOAPI()` does the same with `access_token`. When `http.Client.Do()` fails, the resulting `*url.Error` contains the full URL, including its query string. These errors eventually reach structured logs through the outbound dispatcher.

The signed OSS download URL has the same issue: a failed download can log the complete signed URL.

Relevant locations:

- `internal/channels/dingtalk/client.go:143`
- `internal/channels/dingtalk/client.go:271`
- `internal/channels/dingtalk/media_inbound.go:125`
- `internal/channels/dispatch.go:108`

Could we sanitize query parameters at the HTTP client boundary before wrapping errors? Doing it centrally would be safer than relying on every logging caller to redact them.

### 2. [Reliability] AI Card writes synchronously block event broadcasting

`Manager.HandleAgentEvent()` is documented as needing to remain non-blocking, but `cardStream.Update()` synchronously performs several HTTP calls when materializing the first card: create, deliver, state update, and streaming push.

`MessageBus.Broadcast()` invokes subscribers synchronously and serially, so a slow DingTalk API call can block subsequent agent events and every later subscriber—potentially for multiple HTTP timeout periods.

Relevant locations:

- `internal/channels/events.go:17`
- `internal/channels/dingtalk/card.go:181`
- `internal/bus/bus.go:120`

I think Card writes need a per-stream, ordered async worker that can coalesce stale repaint frames.

### 3. [Lifecycle] A timed-out start can later mark a stopped channel healthy

The instance loader calls `Stop()` after a start timeout. However, if `transport.Start()` later returns successfully, `Channel.Start()` still unconditionally calls `SetRunning(true)` and `MarkHealthy()`.

That can leave the transport and history flusher stopped while the dashboard reports the channel as healthy.

Relevant locations:

- `internal/channels/instance_loader.go:428`
- `internal/channels/dingtalk/dingtalk.go:195`

Could this use an explicit starting/running/stopped state, or at least re-check whether the channel was stopped before marking it healthy?

### 4. [Permissions] `group_sender` session scope leaks into group-level identities

With `group_session_scope=group_sender`, `chatID()` appends the sender ID. That derived ID is then used for group pairing and the file-writer permission scope.

This changes pairing from group-level to per-member pairing and prevents permissions granted to the real group scope from matching `/new`.

Relevant locations:

- `internal/channels/dingtalk/inbound.go:299`
- `internal/channels/dingtalk/policy.go:42`
- `internal/channels/dingtalk/commands.go:138`

Group pairing and permission scopes should use the original `ConversationID`; only the session key should include the sender suffix.

## Other issues

### 5. The command path can block forever on a saturated inbound bus

Normal messages use cancellable `TryPublishInbound()` retries, but `handleCommand()` directly calls blocking `PublishInbound()`. During shutdown, a command goroutine can therefore remain stuck indefinitely.

- `internal/channels/dingtalk/commands.go:99`

Could this reuse `Channel.publish(ctx, msg)` and only acknowledge the command if publishing succeeds?

### 6. `/stop` and `/stopall` send duplicate and potentially contradictory feedback

The DingTalk handler immediately replies with “task stopped,” while the shared consumer later sends feedback based on the actual cancellation result—including “no active task.”

Users can therefore receive two messages whose claims disagree.

- `internal/channels/dingtalk/commands.go:86`
- `cmd/gateway_consumer_handlers.go:469`

As with Telegram, it would be better for stop commands to let the consumer send the single authoritative result.

### 7. Card creation and termination have a race

`ensureCard()` checks `created` and releases the mutex before performing network calls. Concurrent updates can both create cards and overwrite `outTrackID`.

There is also an Update/Stop race: `Stop()` can observe `created=false`, mark the stream done, and return while an in-flight `ensureCard()` subsequently delivers a card. That card may remain permanently in the INPUTING state.

- `internal/channels/dingtalk/card.go:210`
- `internal/channels/dingtalk/card.go:371`

An explicit `creating` state or a serialized per-card lifecycle would close both windows.

### 8. The configured 60-second file download timeout is ineffective

All requests use an `http.Client` with a 30-second timeout, so the 60-second file-download context can never be reached. The same client-level limit also applies to larger chunk uploads.

- `internal/channels/dingtalk/client.go:73`
- `internal/channels/dingtalk/media_inbound.go:115`

A dedicated media client or transport-level timeouts combined with per-request context deadlines would preserve the intended budgets.

### 9. The v1.0 API token has no fallback TTL

`apiToken()` caches the token using `expireIn` directly. If DingTalk omits the field or returns zero, the cached token expires immediately, so every API operation fetches another token. The token-cache mutex prevents a refresh stampede, but it also serializes every request behind that extra network round trip.

- `internal/channels/dingtalk/client.go:126`
- `internal/channels/dingtalk/client.go:136`

`oapiToken()` already handles the equivalent case with `defaultOAPITokenTTL`. The v1.0 path should apply the same kind of positive fallback TTL.

### 10. Proactive group delivery has an unsafe fallback when conversation metadata is missing

For group messages, `postProactive()` requires the original DingTalk `openConversationId`. If `dingtalk_conversation_id` is missing, it falls back to `msg.ChatID`. Under `group_session_scope=group_sender`, however, that value is the derived session identity (`conversation:sender`), not a valid DingTalk conversation ID.

- `internal/channels/dingtalk/outbound.go:132`
- `internal/channels/dingtalk/outbound.go:135`

The normal inbound-reply path preserves both metadata fields together, so this mainly affects metadata-less or externally initiated outbound messages. Deriving the conversation ID by trimming a suffix would be ambiguous; it would be safer to reject proactive group delivery when the required metadata is absent, or to carry the original conversation ID as an explicit routing field for cron/delegate sends.

## Lower-priority observations

- `takeCard()` removes the card before repaint succeeds. A repaint failure loses the handoff and also prevents media delivery; a later retry can then post a duplicate standalone message instead of repainting the card.
- Outbound attachments are fully loaded with `os.ReadFile()` before chunking, so large files can cause significant memory spikes.
- Blank updates advance `lastPush`, which can delay the first real token by one repaint interval.

I also checked a couple of possible false positives:

- Moving `github.com/ollama/ollama` from indirect to direct is correct—the project imports its API package directly.
- The `runCtx` lifetime is currently safe because the loader intentionally passes a long-lived context rather than its start-timeout context.

`go test -race ./internal/channels/dingtalk` passes, but the current tests do not appear to cover late start completion, concurrent Card Update/Stop, a saturated command bus, or the end-to-end stop feedback path.

Thanks again—there is a lot of good engineering in this channel implementation. The main things I’d want resolved before merge are the credential logging risk, event-loop blocking, lifecycle race, and group-scope identity issue.

clark-cant:
## 🤖 Maintainer Triage (github-maintain)

**Status:** Blocked — merge conflicts detected (DIRTY state)

**Assessment:**
- 6,952 additions across 48 files — substantial new channel implementation
- DingTalk Stream mode + AI Card streaming is a valuable contribution
- Merge state is DIRTY: branch needs rebase onto current `dev`

**Action required from author:**
1. Rebase `feat/dingtalk-channel` onto latest `origin/dev`
2. Resolve merge conflicts
3. Push updated branch

**After rebase:**
- CI checks will run automatically
- Maintainers will review the implementation

**Labels added:** `pr:oversized` (for reviewer attention), `status:blocked` (merge conflicts)

---
*Posted by github-maintain at 2026-07-15T00:30:00Z*

clark-cant:
## 🔍 Maintainer Triage (github-maintain)

**PR type:** Feature — new channel (DingTalk with Stream mode + AI Card streaming)
**Size:** 48 files, +6952/-19 lines — oversized for automated review
**Merge state:** DIRTY (needs rebase onto dev)
**CI:** No checks reported

### Assessment
This is a substantial feature PR adding a new first-class channel type. The scope and quality look impressive — Stream transport, AI Card streaming, media handling, slash commands, web UI parity, and extensive unit tests.

### Blockers before review
1. **Rebase required** — branch is DIRTY, needs rebase onto current dev
2. **Oversized for automated review** — 6952 additions across 48 files exceeds bounded cron review budget. This will need maintainer deep-review.

### Recommendation
- Author: please rebase onto latest dev
- Maintainer: schedule dedicated review session for this PR (not suitable for 15-min cron slot)

### Notes
- Plan document referenced (plans/260709-2230-dingtalk-channel/plan.md) — good practice
- Uses official DingTalk Stream SDK (good choice over hand-rolled WS)
- Companion docs update included

---

clark-cant:
🦸‍♂️ **Maintainer check — github-maintain**

This PR adds a substantial new channel (DingTalk with Stream mode + AI Card streaming) — 48 files, +6952 lines. Well-documented with test coverage.

**Current blockers:**
1. **Merge conflicts** — branch has conflicts with `dev` base (mergeState: DIRTY). Please rebase/merge latest `dev` and resolve conflicts.
2. **No CI checks** — GitHub Actions have not run on this PR. CI needs to pass before review can proceed.

**Next steps:**
1. Rebase onto latest `dev` and resolve conflicts
2. Push to trigger CI
3. Once CI is green, we'll do a full review

The PR structure looks solid (channel package follows existing patterns like Feishu/Discord, factory registration, web UI schema updates). Just needs the branch cleanup before we can proceed with review.

PR #1432, opened 2026-07-13
fix(bitrix24): mount shared webhook route independent of bot readiness

tech-synity:
## What changed

Mounts the shared Bitrix24 webhook router (`/bitrix24/install`, `/bitrix24/handler`, `/bitrix24/events`) unconditionally at gateway boot, independent of whether any Bitrix24 channel instance loaded successfully.

## Why

The webhook route was only mounted when a Bitrix24 `Channel` loaded via `WebhookHandlers()`, and the channel factory requires `portal` + `bot_code` + `bot_name` to already be set (`internal/channels/bitrix24/factory.go`). Completing portal OAuth requires Bitrix to reach `/bitrix24/install`, and the admin UI won't let you select an uninstalled portal when creating a bot (`bitrix-portal-select.tsx`, shows "Pending install - click to resume" instead).

Net effect: on any fresh deployment, the first Bitrix24 portal can never finish OAuth, because the route it needs isn't mounted until a bot exists, and a bot can't be created until OAuth finishes. Every request to `/bitrix24/*` silently falls through to the SPA catch-all instead (returns the dashboard HTML, no error).

## Fix

Claim and mount `bitrix24.WebhookRouter()` directly in the boot sequence (`cmd/gateway_lifecycle.go`), right after the existing per-channel mount loop. `ClaimWebhookRoute()` is idempotent (first-claim-wins via `CompareAndSwap` on `Router.routeTaken`), so this is a no-op if a real Channel already claimed the route in the loop above - no double-registration risk.

## Validation

- `go build ./...` and `go build -tags sqliteonly ./...` - clean
- `go vet ./cmd/... ./internal/channels/bitrix24/...` - clean
- `go test ./internal/channels/bitrix24/...` - pass
- `go test ./cmd/...` - same single pre-existing failure with and without this change (`TestShellDenyGroupsConfigReload_ReplacesDBClaudeCLIProvider`, missing test fixture binary on Windows, unrelated to this diff)
- Reproduced live on a fresh TOSE deployment: boot log showed `failed to load channel instance ... requires portal, bot_code, and bot_name` followed by `no channels enabled` / `channels=[]`, and every `/bitrix24/*` request returned the dashboard SPA HTML instead of a handler response

### DIFF
--- a/cmd/gateway_lifecycle.go
+++ b/cmd/gateway_lifecycle.go
@@ -10,6 +10,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/bus"
 	"github.com/nextlevelbuilder/goclaw/internal/cache"
 	"github.com/nextlevelbuilder/goclaw/internal/channels"
+	"github.com/nextlevelbuilder/goclaw/internal/channels/bitrix24"
 	"github.com/nextlevelbuilder/goclaw/internal/config"
 	"github.com/nextlevelbuilder/goclaw/internal/edition"
 	"github.com/nextlevelbuilder/goclaw/internal/heartbeat"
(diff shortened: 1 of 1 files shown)


clark-cant (approved):
LGTM. Clean fix for the Bitrix24 webhook boot deadlock. ClaimWebhookRoute is idempotent, so the extra mount is safe. All CI green.

PR #1437, opened 2026-07-15
fix(ollama): resolve context window per model via POST /api/show

bclermont:
Summary

  Ollama-backed agents were being served a 4096-token context window regardless of the model's real capacity, causing every prompt larger than a few thousand tokens to fail with 400 
  exceed_context_size_error. The startup probe meant to auto-detect each model's context length never worked: it used the wrong HTTP method, queried a hardcoded model name, and parsed
  a response shape Ollama never returns. This PR makes context resolution actually work, per model, at request time.

  Root cause

  Three independent bugs stacked in the Ollama context-resolution path:

  1. FetchOllamaModelContext issued a GET to /api/show. That endpoint is POST-only — a GET is answered with 405 method not allowed, so every lookup silently fell back to the default.
  (Also produced continuous 405 warning spam in logs.)
  2. It parsed a flat model_info.context_length field. A real Ollama server never returns that — it namespaces the key by model architecture: gemma4.context_length,
  qwen35.context_length, llama.context_length, etc. So even a successful call yielded 0.
  3. Resolution ran once at startup against a hardcoded model ("llama3.3"). The model an agent actually uses isn't known until it sends a request, so a wrong guess resolved to nothing
  — and the resolved value never tracked the real model.

  Additionally, Ollama's OpenAI-compatible /v1/chat/completions shim silently ignores options.num_ctx — only the native /api/chat endpoint (which OllamaProvider already uses) honors
  it. This is now documented inline so the constraint isn't reintroduced.

  Changes

  - internal/providers/ollama_context.go
    - FetchOllamaModelContext now POSTs {"model": "<name>"} with a JSON body.
    - New extractContextLength scans model_info for context_length or any *.context_length (arch-namespaced) key; the bare key is still accepted for fixtures.
  - internal/providers/ollama.go
    - New OllamaProvider.resolveNumCtx(ctx, model): resolves the context window for the model actually requested, memoised per model under a sync.RWMutex.
    - Priority: explicit operator-configured num_ctx wins (the lever for capping KV cache to fit VRAM); otherwise the model's own context length from /api/show, bounded by 
  OllamaDefaultNumCtx so an enormous advertised window (e.g. Qwen3.5 reports 262144) can't balloon the KV-cache allocation; on any failure it falls back to OllamaDefaultNumCtx, never
  to Ollama's silent 4096.
    - buildRequest now takes ctx and calls resolveNumCtx.
  - cmd/gateway_providers.go
    - resolveOllamaNumCtx simplified to the explicit-settings override only; the broken startup /api/show probe (with its hardcoded model + 5s timeout) is removed. Per-model resolution
  now lives in the provider.
  - internal/providers/error_classify.go
    - Classify Ollama's native exceed_context_size 400 as context_overflow, so the pipeline's context-overflow recovery (emergency compaction + retry) engages instead of surfacing a
  raw error. Belt-and-suspenders: the num_ctx fix should stop this 400 from occurring at all.

  Tests

  - TestFetchOllamaModelContext_Success updated to assert the POST contract (method + JSON body).
  - TestFetchOllamaModelContext_ArchNamespacedKey — real-world gemma4.context_length response.
  - TestOllamaProviderResolvesNumCtxPerModel — resolves for the requested model (not a hardcoded name) and serves the second call from cache.
  - TestOllamaProviderExplicitNumCtxWins — configured num_ctx wins and skips the lookup.
  - TestClassifyContextWindowOllama — Ollama 400 classified as context overflow.

  Checklist: gofmt, go vet ./internal/providers/ ./cmd/, go build ./..., go build -tags sqliteonly ./..., and go test ./internal/providers/ all clean.

  Cross-surface parity

  - Gateway/provider runtime: the change (fixed).
  - API contract: N/A — request/response structs and pkg/protocol unchanged; internal provider behavior only.
  - Web UI: N/A — num_ctx is resolved server-side from the model; no UI field exists for it.
  - CLI/runtime package: N/A — resolveOllamaNumCtx is internal startup wiring; signature simplified but not user-facing.

  Notes

  - No config or migration changes; no new dependencies.
  - Behavior for non-Ollama providers is unchanged.

### DIFF
--- a/cmd/gateway_providers.go
+++ b/cmd/gateway_providers.go
@@ -316,7 +316,7 @@ func registerProvidersFromDB(registry *providers.Registry, provStore store.Provi
 			if host == "" {
 				host = "http://localhost:11434"
 			}
-			numCtx := resolveOllamaNumCtx(&p, config.DockerLocalhost(host), "")
+			numCtx := resolveOllamaNumCtx(&p)
 			prov := providers.NewOllamaProvider(p.Name, config.DockerLocalhost(host), "llama3.3", numCtx, nil).
 				WithThinkingEnabled(store.ParseThinkingEnabled(p.Settings))
 			registry.RegisterForTenant(p.TenantID, prov)
@@ -397,7 +397,7 @@ func registerProvidersFromDB(registry *providers.Registry, provStore store.Provi
 			if base == "" {
 				base = "https://ollama.com"
 			}
-			numCtx := resolveOllamaNumCtx(&p, base, p.APIKey)
+			numCtx := resolveOllamaNumCtx(&p)
 			prov := providers.NewOllamaProvider(p.Name, base, "llama3.3", numCtx, nil).
 				WithThinkingEnabled(store.ParseThinkingEnabled(p.Settings))
 			registry.RegisterForTenant(p.TenantID, prov)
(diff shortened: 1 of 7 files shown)


clark-cant (approved):
LGTM. Clear root cause analysis, focused fix, good test coverage. Builds appropriately on prior work (#1307).

PR #1438, opened 2026-07-15
fix(cron): guard nil job + tenant-consistent readback in cron create

tech-synity:
## Symptom
Creating a cron job over a connection with **no tenant in context** (gateway-token / CLI, or a master-scope UI session) panics with a nil-pointer dereference in `handleCreate`.

## Root cause — tenant asymmetry between insert and readback
1. `internal/store/pg/cron_crud.go` `AddJob` INSERT uses `tenantIDForInsert(ctx)`, which falls back to `store.MasterTenantID` when the context tenant is `uuid.Nil` → row inserted OK under master tenant.
2. `internal/store/pg/cron_scan.go` `scanJob` (and `lockCronJobForMutation` in `cron_update.go`) reject a nil-tenant context with `"tenant_id required"`. So the readback with the same nil-tenant context **errors**.
3. `GetJob` returns `(nil, false)` → `AddJob` swallows it and returns `(nil, nil)` (job nil, err nil).
4. `internal/gateway/methods/cron.go` `handleCreate` then dereferences `job.ID` at the post-create `UpdateJob` patch → **panic**. Even without the panic, that patch would silently fail for the same nil-tenant reason, leaving the cron with an incomplete payload.

## Fix (defense-in-depth)
- **`internal/gateway/methods/cron.go` `handleCreate`**: add a `job == nil` guard before any `job.ID` use — returns `ErrInternal` instead of panicking. Apply the post-create stateless/command patch under a tenant-consistent context `store.WithTenantID(ctx, job.TenantID)` so `lockCronJobForMutation` finds the row for nil-tenant callers.
- **`internal/store/pg/cron_crud.go` + `internal/store/sqlitestore/cron_crud.go` `AddJob`**: read back under `store.WithTenantID(ctx, tenantIDForInsert(ctx))` so insert and readback share the same tenant; return an explicit error if readback fails instead of swallowing it. Fixed in both PG and SQLite stores for cross-surface parity (identical asymmetry existed in both).

## Tests
- `internal/store/sqlitestore/cron_crud_test.go`: `AddJob` with a nil-tenant context returns a **non-nil** job + nil error, tenant = MasterTenantID (real SQLite temp-file DB). Fails before the fix (returned `(nil, nil)`).
- `internal/gateway/methods/cron_test.go`: `handleCreate` does not panic and does not call `UpdateJob` when the store returns `(nil, nil)`.

## Verification
- `go build ./...` and `go build -tags sqliteonly ./...` — clean.
- `go vet ./internal/store/pg/... ./internal/gateway/methods/...` — clean.
- `go test ./internal/gateway/methods/ -run Cron` — pass.
- `go test -tags sqliteonly ./internal/store/sqlitestore/ -run Cron` — pass.
- `go test ./internal/store/pg/ -run Cron` — pass. (3 unrelated pre-existing `TestBuildSkillInfo` failures on Windows are path-separator issues, confirmed present on the base commit and untouched by this change. PG integration tests requiring a live DB were not run — no `TEST_DATABASE_URL` available.)

### DIFF
--- a/internal/gateway/methods/cron.go
+++ b/internal/gateway/methods/cron.go
@@ -107,6 +107,13 @@ func (m *CronMethods) handleCreate(ctx context.Context, client *gateway.Client, 
 		client.SendResponse(protocol.NewErrorResponse(req.ID, protocol.ErrInvalidRequest, err.Error()))
 		return
 	}
+	// Anti-panic guard: a store may return (nil, nil) if the insert succeeded but
+	// the tenant-scoped readback failed (e.g. tenant asymmetry). Never dereference
+	// a nil job below.
+	if job == nil {
+		client.SendResponse(protocol.NewErrorResponse(req.ID, protocol.ErrInternal, i18n.T(locale, i18n.MsgInternalError, "cron job created but could not be loaded")))
+		return
+	}
 
 	// Apply extra fields not in AddJob signature via an immediate patch.
 	// Default stateless=true for new crons (saves tokens); override with explicit false.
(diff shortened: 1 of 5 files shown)


clark-cant (approved):
## Review — github-maintain

**Summary:** Defense-in-depth fix for a nil-pointer panic in cron create when the store's AddJob returns (nil, nil) due to tenant-asymmetry between insert and readback.

**Risk level:** Low — contained fix, well-scoped, no API changes.

**Mandatory gates:**
- Duplicate / prior implementation: clear — no prior PR covers this specific tenant-asymmetry path
- Project standards: clear — follows existing store patterns, consistent PG/SQLite parity
- Strategic necessity: clear value — fixes a production panic for nil-tenant callers (gateway-token / CLI / master-scope UI)

**Findings:**
- Fix is correct: nil guard before dereference + tenant-consistent readback context in both stores
- Tests cover both the nil guard (handleCreate) and the readback fix (SQLiteCronStore)
- Comments are clear and explain the why, not just the what

**Verdict: Approve** ✅

PR #1443, opened 2026-07-16
feat(bitrix24): recognize reply/quote context across all message types

tech-synity:
## Problem

When a user **replies to / quotes** an earlier message, the Bitrix24 channel showed the agent nothing about the quoted message. `events.go` parsed a `REPLY_TO_MESSAGE_ID` field that Bitrix never sends, so the reply signal was silently dropped and the agent answered without the context it was clearly meant to use.

Bitrix actually ships the quoted message inline on `ONIMBOTMESSAGEADD` as `data[PARAMS][REPLY_MESSAGE][ID|AUTHOR_ID|MESSAGE]` (plus `PARAMS[PARAMS][REPLY_ID]`). For a **text** original the full body is in `MESSAGE`; for a **media** original (image/audio/file/video) `MESSAGE` is empty and no file id is included.

## Approach (Hybrid)

- **Parse** `REPLY_MESSAGE` (form + JSON) into `EventParams.ReplyMessage`, with `PARAMS[REPLY_ID]` as a fallback id. Removes the dead `ReplyToMID` field + `bitrix_reply_to_mid` meta.
- **`reply_context.go` (new)** — `resolveReplyContext`:
  - **Text quote** → sanitize BBCode (reuse the existing user-mention rewriter) + truncate, prepend `[Đang trả lời tin của <author>: "..."]`. No API call.
  - **Media quote** → `im.dialog.messages.get` to recover the quoted attachment, then re-inject it through the **same download pipeline** as a direct upload (STT / vision / read_document), bounded by the existing `media_max_mb`. Always prepends a note; falls back to note-only on any fetch/download failure.
  - `im.dialog.messages.get` is used deliberately: `imbot.v2.Chat.Message.get` returns `BOT_TYPE_NOT_ALLOWED` for `TYPE=bot` bots, whereas `im.dialog.messages.get` (scope `im`) works with the bot token.
- **`handle.go`** — wire the note + extra media into message assembly.

Covers text, image, audio, file, video quotes. Non-reply events are a no-op; one extra REST call only when the quote is media.

## Safety / DRY

- Reuses `downloadEventFiles` (size cap, SSRF-safe client, temp-file streaming) — no new size/MIME logic.
- Best-effort: fetch/download failure never blocks the user's own message from reaching the agent.
- No recursion (does not resolve quote-of-quote). One `im.dialog.messages.get` call, only for media quotes.

## Tests

[...]

PR #1444, opened 2026-07-16
fix(bitrix24): label quoted-media reply note from downloaded MIME

tech-synity:
Follow-up to #1443. For a media-only quoted message, `im.dialog.messages.get` frequently returns a blank type/name (observed live for a voice note), so the reply note fell back to the generic **"một tệp"** even though the file was correctly downloaded and tagged `<media:audio>`.

## Fix

Build the note *after* downloading and derive the kind label from the downloaded file's MIME (reusing `classifyMediaType`), so it reads **"một tin nhắn thoại" / "một hình ảnh" / "một video"** — matching the `<media:*>` tag the agent receives. Falls back to the Bitrix file type, then `tệp`, when no download is available (e.g. oversized). Also adopts the downloaded filename when `messages.get` omitted it.

## Tests

MIME-preference path + `mediaKindVN`. Compile (PG + sqliteonly) + `go vet` + `go test ./internal/channels/bitrix24/...` all green.

## Surface parity

Gateway-only; API contract / Web UI / CLI N/A.

### DIFF
--- a/internal/channels/bitrix24/reply_context.go
+++ b/internal/channels/bitrix24/reply_context.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/nextlevelbuilder/goclaw/internal/bus"
+	"github.com/nextlevelbuilder/goclaw/internal/channels/media"
 )
 
 // quotedTextMaxRunes caps how much of a quoted text message we inline into the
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
Clean fix for the blank metadata edge case. Good use of downloaded MIME as the authoritative source, proper fallback chain, and solid regression test coverage. LGTM.

PR #1445, opened 2026-07-16
feat(bitrix24): send pairing reply with code instead of silent drop

tech-synity:
## Problem

With an agent's DM (or Group) policy set to **"pairing (require code)"**, an unpaired Bitrix24 sender got **silence**. The channel only had a Phase-03 stub, `logPairingNeeded`, which logged *"pairing required (Phase 07 will send pairing reply)"* and dropped the message. No pairing code was ever generated or sent, so an admin had **nothing to approve** — the pairing gate was effectively "block everyone" on Bitrix.

Bitrix24 was the **only** channel missing this — Discord / Zalo / WhatsApp / Slack / Feishu all already send a pairing reply.

## Fix

Replace the stub with `sendPairingReply`, mirroring the other channels:

- `RequestPairing()` to generate the code — keyed by `senderID` for DMs and `"group:<chatID>"` for groups, so the generated code matches what `CheckGroupPolicy` / `CheckDMPolicy` later verify via `IsPaired`.
- Render the shared `systemmessages.KeyPairingAccountRequired` message (code + how to approve).
- Send it back into the chat via the **v2** API (`imbot.v2.Chat.Message.send`) — consistent with the welcome / OAuth-invite messages — **not** the legacy `imbot.message.add` v1.
- Keep the existing per-sender debounce so a spammy sender can't flood the chat.

## Tests

`pairing_test.go`: the reply fires on the **v2** endpoint carrying the code; no send without a pairing service; the second call within the debounce window is suppressed.

Compile (PG + sqliteonly) + `go vet` + `go test ./internal/channels/bitrix24/...` all green.

## Surface parity

Gateway/channel-only. API contract / Web UI / CLI: N/A — inbound policy handling + an outbound chat message; no request/response struct, UI, or CLI change.

### DIFF
--- a/internal/channels/bitrix24/handle.go
+++ b/internal/channels/bitrix24/handle.go
@@ -13,6 +13,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/channels"
 	"github.com/nextlevelbuilder/goclaw/internal/channels/media"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
+	"github.com/nextlevelbuilder/goclaw/internal/systemmessages"
 )
 
 // mentionMatcher is the compiled-once regex + rendered tag string used to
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
Excellent work bringing Bitrix24 to pairing parity with other channels. Clean implementation, proper error handling, and solid test coverage. All CI green and merge-ready.

PR #1447, opened 2026-07-17
feat(audio): openai_compat TTS/STT provider for self-hosted endpoints

bclermont:
## What

Adds `internal/audio/openaicompat` — a TTS + STT provider pair speaking the
OpenAI audio wire format (`POST /audio/speech`, `POST /audio/transcriptions`)
against **any** OpenAI-compatible endpoint: gpu-manager, Speaches, vLLM,
LocalAI, llama.cpp, Ollama. A self-hosted engine no longer needs a bespoke HTTP
shim implementing a proprietary `/transcribe_audio` contract.

Enabled by a new `tts.openai_compat` config block; setting `api_base` registers
**both** providers, since a compatible engine serves speech-out and speech-in at
one endpoint.

## Why a new package instead of reusing `openai`

Both speak the same wire format, but `IsVoiceCompatible` applies OpenAI's voice
allowlist to any provider named `"openai"`, and `FilterVoiceForProvider` then
silently substitutes `"alloy"` for anything outside it. A self-hosted voice ID
such as Piper's `fr_FR-gilles-low` would be replaced with no error and no log
line — the caller hears the wrong language. Providers without validation rules
pass through untouched, which is correct for an engine whose voice namespace is
its own. There is a regression test asserting exactly this contrast.

## Deliberate departures from the `openai` package

- **`api_base` is required.** No public endpoint to fall back to; defaulting to
  `api.openai.com` would send self-hosted traffic to a vendor. `api_base` is the
  enable switch, not an API key.
- **Empty `api_key` omits the `Authorization` header** rather than sending an
  empty Bearer token — self-hosted engines commonly have no auth.
- **Empty voice/model are omitted** from the request, not sent blank, since some
  engines resolve the model from the requested voice.
- **The multipart filename is never blank** — several engines type the audio by
  extension rather than `Content-Type` and reject what they cannot type.
- Endpoint error bodies are preserved in the returned error so a format mismatch
  (e.g. an engine that cannot encode mp3) is diagnosable.

The STT chain is now built from the providers that actually registered, instead
of a hardcoded `{elevenlabs, proxy}`. `openai_compat` goes first when present;
`proxy` stays last for `BridgeLegacySTT`. Behaviour is unchanged when
`tts.openai_compat.api_base` is unset — the chain still resolves to
`{elevenlabs, proxy}`.

## Surface parity

- **Gateway server:** provider pair + registration in `setupTTSManager` /
  `setupAudioExtras`. ✅
- **API contract:** `TtsConfig` gains `openai_compat` (`TtsOpenAICompatConfig`). ✅
- **Web UI:** N/A — TTS provider selection is config/DB-driven; no UI enum change
  required for this provider to function. Adding it to any provider picker is a
  follow-up.
- **CLI/runtime:** N/A — no CLI command references the TTS provider list.

## Tests

`internal/audio/openaicompat` — 83.1% coverage. `gofmt`/`go vet` clean;
`go build ./...` and `go build -tags sqliteonly ./...` both green.

Verified end-to-end against a live gpu-manager (Piper `fr_FR-gilles-low`):
`/v1/audio/speech` returned valid WAV, and feeding it back through
`/v1/audio/transcriptions` returned French text.

### DIFF
--- a/cmd/gateway_agents.go
+++ b/cmd/gateway_agents.go
@@ -7,6 +7,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/audio/elevenlabs"
 	geminiaudio "github.com/nextlevelbuilder/goclaw/internal/audio/gemini"
 	minimaxaudio "github.com/nextlevelbuilder/goclaw/internal/audio/minimax"
+	"github.com/nextlevelbuilder/goclaw/internal/audio/openaicompat"
 	"github.com/nextlevelbuilder/goclaw/internal/bus"
 	"github.com/nextlevelbuilder/goclaw/internal/config"
 	"github.com/nextlevelbuilder/goclaw/internal/memory"
@@ -273,6 +274,26 @@ func setupTTS(cfg *config.Config) *tts.Manager {
 		}))
 	}
 
+	// OpenAI-compatible self-hosted endpoint. api_base is the enable switch —
+	// unlike the vendor providers there is no API key to gate on, since these
+	// endpoints commonly have no auth.
+	if base := ttsCfg.OpenAICompat.APIBase; base != "" {
+		provider, err := openaicompat.NewTTSProvider(openaicompat.Config{
+			APIBase:   base,
+			APIKey:    ttsCfg.OpenAICompat.APIKey,
+			TTSModel:  ttsCfg.OpenAICompat.Model,
+			TTSVoice:  ttsCfg.OpenAICompat.Voice,
+			TTSFormat: ttsCfg.OpenAICompat.Format,
+			TimeoutMs: ttsCfg.TimeoutMs,
+		})
+		if err != nil {
+			slog.Warn("audio.tts: openai_compat not registered", "error", err)
+		} else {
+			mgr.RegisterProvider(provider)
+			slog.Info("audio.tts: openai_compat registered", "api_base", base)
+		}
+	}
+
 	if key := ttsCfg.ElevenLabs.APIKey; key != "" {
 		mgr.RegisterProvider(tts.NewElevenLabsProvider(tts.ElevenLabsConfig{
 			APIKey:    key,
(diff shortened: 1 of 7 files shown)


clark-cant (approved):
## Summary

Adds `internal/audio/openaicompat` — a TTS + STT provider pair speaking the OpenAI audio wire protocol against any compatible self-hosted endpoint (gpu-manager, Speaches, vLLM, LocalAI, llama.cpp, Ollama). Well-scoped, well-documented, and deliberately distinct from the `openai` package to avoid the voice-allowlist trap.

## Risk Level: **Low**

Contained new package, no changes to existing provider behavior, all CI green.

## Mandatory Gates

- **Duplicate / prior implementation:** ✅ Clear — no prior openai_compat provider exists. #1005 (jahrulnr) touched STT routing but was closed and unrelated.
- **Project standards:** ✅ Follows existing audio provider patterns (elevenlabs, gemini, minimax). Config integration consistent with `tts.*` conventions.
- **Strategic necessity:** ✅ Clear value — self-hosted TTS/STT is a core requirement for air-gapped and privacy-sensitive deployments. Fills a real gap.

## Findings

No Critical or Important issues found.

### Suggestions (non-blocking)

1. **`stt.go` — `writeAudio` path**: The file path branch does `os.Stat` then `os.Open` with a TOCTOU gap. Practically irrelevant for a server-side audio file the operator configured, but worth noting for completeness.

2. **`tts.go` — `Synthesize` reads entire response into memory**: For very long TTS outputs this could be large. Current 30s timeout caps practical size. Consider streaming for future iterations if long-form synthesis is requested.

3. **`config.go` — `TimeoutMs` validation**: `withDefaults()` treats `TimeoutMs <= 0` as "use default". An explicit `TimeoutMs: 0` from config would silently get 30s. Minor — matches the documented "defaults to 30000" behavior.

## Verdict: **Approve**

Solid implementation. The PR description is excellent — clearly explains why a separate package was needed, documents deliberate departures from the `openai` package, and notes the regression test for the voice-validation trap. Test coverage is thorough (filename derivation, error paths, size limits, auth omission, language fallback).

*Posted by /ck:review-pr at 2026-07-17T15:50:04Z*

PR #1461, opened 2026-07-21
fix(ui): align workstation web contract with gateway API shape

### LINKED ISSUE #1258: Add Workstation dialog is broken: snake_case/camelCase field mismatch + dead identity_file field
## Summary

The **Add Workstation** dialog cannot create any workstation (SSH or Docker). Every submit fails with:

```json
{"error":{"code":"INVALID_REQUEST","message":"workstationKey is required"}}
```

even when the Workstation Key field is filled correctly. Root cause is a JSON field-name casing mismatch between the web UI and the backend — the UI sends `snake_case`, the backend decodes `camelCase`, so the keys are silently dropped.

This breaks the feature end-to-end on both transports (WebSocket RPC and HTTP).

## Root cause

**Frontend sends snake_case** — `ui/web/src/pages/workstations/workstation-create-dialog.tsx`:

```ts
await onCreate({ workstation_key: key.trim(), name: name.trim(), backend_type: backend, metadata });
```

`ui/web/src/pages/workstations/hooks/use-workstations.ts` passes the params straight through to the RPC with no remap:

```ts
const res = await ws.call(Methods.WORKSTATIONS_CREATE, params as unknown as Record<string, unknown>);
```

and `CreateWorkstationParams` is declared with `workstation_key` / `backend_type`.

**Backend decodes camelCase** — `internal/gateway/methods/workstations.go`:

```go
WorkstationKey string                   `json:"workstationKey"`
Name           string                   `json:"name"`
BackendType    store.WorkstationBackend `json:"backendType"`
```

`internal/http/workstations.go` (handleCreate) uses the same camelCase tags. So `workstation_key` → `WorkstationKey` stays empty → the `workstationKey is required` guard fires. `backend_type` → `BackendType` would fail next for the same reason.

## Secondary issue — `identity_file` is a dead field

The dialog has an **Identity File (optional)** input and packs it into metadata as `identity_file`:

```ts
...(identityFile.trim() ? { identity_file: identityFile.trim() } : {}),
```

But `store.SSHMetadata` (`internal/store/workstation_store.go`) has no `identity_file` (nor `identityFile`) field. SSH auth only accepts an **inline PEM `privateKey`** or a `password`:

```go
type SSHMetadata struct {
    Host       string `json:"host"`
    Port       int    `json:"port"`
    User       string `json:"user"`
    PrivateKey string `json:"privateKey,omitempty"`
    Password   string `json:"password,omitempty"`
    ...
}
// UnmarshalSSHMetadata: returns "privateKey or password is required" if both empty
```

So even after the casing fix, the UI cannot supply working SSH credentials — an identity-file path is ignored, and the dial fails `privateKey or password is required`.

## Reproduction

1. Dashboard → Workstations → Add Workstation.
2. Fill Display Name, Workstation Key (`aether`), Backend SSH, Host, Port, SSH User, Identity File.
3. Create → `{"error":{"code":"INVALID_REQUEST","message":"workstationKey is required"}}`.

## Impact

Add Workstation is non-functional for all users. The only way to create a workstation is to call the API directly with the correct camelCase keys and an inline `privateKey` in metadata.

## Suggested fix

1. Align the field-name contract — either remap UI params to `workstationKey` / `backendType` in `use-workstations.ts` (and the `metadata` keys for Docker, e.g. `docker_host`), or add `json:"workstation_key"` aliases on the backend structs. Pick one convention and apply it across both transports.
2. Resolve the credentials UX: replace the "Identity File" input with a "Private Key (PEM)" field that maps to `metadata.privateKey` (and/or a password field), so the dialog can produce a dialable SSH workstation.

Refs: #941 (feature that introduced this).

eddydao:
Fixes #1258.

## Problem

The workstations page was written against snake_case field names the gateway never served. Go's json decoder drops unknown members silently rather than erroring, so the mismatch surfaced as confusing validation failures instead of decode errors:

- **Create failed every time.** The dialog sent `workstation_key` / `backend_type`; the handler reads `workstationKey` / `backendType` (`internal/gateway/methods/workstations.go:125-127`). Both landed as zero values → `"workstationKey is required"`, regardless of what was typed.
- **List rendered garbage.** The `Workstation` interface declared `workstation_key` / `backend_type` / `created_at`, but the API serializes camelCase (`internal/store/workstation_store.go:14-26`) → blank key column, `backend.undefined`, `Invalid Date`.
- **SSH credentials never validated.** The dialog sent `metadata.identity_file`, which `store.SSHMetadata` does not accept; `UnmarshalSSHMetadata` requires `privateKey` or `password`.

## Fix

Aligned the TypeScript contract on the camelCase shape the WS handler, HTTP handler, and store model already share. No backend changes — per the maintainer's decision on the issue, adding snake_case aliases in Go would leave list rendering and future update flows exposed to a second parallel contract.

- Renamed `Workstation` / `CreateWorkstationParams` fields and every read on the page.
- Replaced the dead Identity File input with a private key (PEM) or password credential path.
- Extracted `buildWorkstationCreatePayload()` so the wire contract is unit-testable without mounting React, following the existing `cli-credentials` helper precedent.
- Moved validation strings into the en/vi/zh catalogs (89 keys, parity verified).

## Two findings beyond the issue text

**1. Docker create was equally broken — not mentioned anywhere in #1258.** The dialog sent `{container, docker_host}`, but `store.DockerMetadata` requires `image` plus `host` or `socketPath`, and stores the container name in `host` (surfaced back as `metadataSummary.containerName`). Every Docker workstation creation failed with `"image is required"`. Added the missing `image` field and mapped container → `host`, daemon endpoint → `socketPath`.

Flagging this explicitly because it expands the issue's stated scope. Fixing only what #1258 describes would have closed it with Docker creation still 100% broken. Happy to split it into its own PR if you'd prefer the boundary kept tight.

**2. Update was broken for a second, independent reason.** `useWorkstations.updateWorkstation` sent `{id, ...params}` flattened, but the handler reads a nested `updates` map (`internal/gateway/methods/workstations.go:180-182`) → `"no updates provided"`. Item 2 of the issue's scope covers this, but the cause was never diagnosed, so a rename-only fix would have left update dead.

## Tests

New `workstation-contract.test.ts` — 7 tests covering create payload shape, SSH credential mapping, Docker metadata, and list rendering against a verbatim `SanitizedView()` response body.

Verified the guards actually fail without the fix: reintroducing the snake_case payload produces `expected [ 'backend_type', … ] to deeply equal [ 'backendType', … ]`.

Full-suite baseline comparison — 6 failing files / 2 failing tests both before and after, an identical pre-existing set (jsdom `localStorage` and network stubs in `http-client`, `channels`, `cli-credentials`). This change adds +7 passing and introduces no new failures. `tsc --noEmit`, `eslint --max-warnings=0`, and `pnpm build` all clean.

## Surface parity

- **Web UI** — fixed.
- **Gateway / API contract** — N/A, unchanged by design; the UI was the wrong side.
- **CLI / runtime package** — N/A, `workstations.create` has exactly one caller in the repo (`ui/web/src/pages/workstations/hooks/use-workstations.ts`); no CLI or desktop surface exists.
- **Docs** — N/A, no documentation describes this dialog.


clark-cant (approved):
## Review — github-maintain

**Summary:** Aligns the workstation web UI TypeScript contract with the actual gateway API shape (camelCase), fixes Docker create metadata, and replaces the dead Identity File input with proper privateKey/password auth. Closes #1258.

**Risk level:** Low — UI-only changes, no backend modifications, well-scoped with regression tests.

### Mandatory gates

- **Duplicate / prior implementation:** clear — no prior PR addresses this contract mismatch
- **Project standards:** docs found — follows existing patterns (cli-credentials helper precedent for extracted payload builder, i18n catalog structure matches other pages)
- **Strategic necessity:** clear value — the workstation create flow was 100% broken for both SSH (snake_case silently dropped by Go decoder) and Docker (missing image field). This is a real user-facing bug.

### Findings

**Positive observations:**
- Excellent PR description with root cause analysis and scope expansion disclosure (Docker create was equally broken — flagged explicitly)
- `buildWorkstationCreatePayload()` extracted into a pure, testable helper — follows the existing `cli-credentials` pattern
- 7 regression tests that actually verify the wire contract, including a test that proves the guards fail without the fix
- i18n parity across en/vi/zh with 89 keys verified
- Update endpoint fix (`{ id, updates: params }` instead of flattened) addresses the second half of #1258 that was never diagnosed
- No anti-slop patterns: no unnecessary abstractions, no defensive paranoia, no phantom coverage

**Suggestion:**
- The `privateKeyHint` i18n string says "encrypted at rest" — ensure this is actually true in the backend storage layer before shipping. If not, soften to "stored securely".

**Verdict: Approve** — clean, well-tested fix for a completely broken workflow. The scope expansion beyond #1258 is well-justified and explicitly flagged.

_Merge deferred: mergeStateStatus=UNSTABLE, no CI checks reported yet. Will merge once CI is green._

*Reviewed by github-maintain at $(date -u +'%Y-%m-%dT%H:%M:%SZ')*

clark-cant (approved):
## Maintainer review — github-maintainnn**Summary:** Fixes #1258 — aligns TypeScript workstation contract with gateway camelCase API shape. +421/-73 across 8 files. Well-documented with tests.nn**Mandatory gates:**n- Duplicate/prior work: clear — no overlap foundn- Project standards: follows existing patterns (cli-credentials helper precedent)n- Strategic necessity: clear value — fixes broken create/list/update for workstationsnn**Findings:** None critical or important. Clean, focused fix.nn**Verdict:** Approvenn*Posted by /ck:review-pr at $(date -u +'%Y-%m-%dT%H:%M:%SZ')*

PR #1464, opened 2026-07-22
perf(channels): shard outbound dispatch per conversation

connermo:
## Problem

`Manager.dispatchOutbound` ran as a single goroutine calling `channel.Send()` synchronously. Every reply in the process — every channel, every user — queued behind the slowest send.

Two consequences:

- **Blast radius.** One media upload taking a few seconds froze all outbound traffic for that long, across unrelated channels.
- **Throughput ceiling.** Sustained delivery was capped at one Send round-trip (~100-200ms for a text send), regardless of how much `LaneMain` capacity was configured.

## Why serialization was not required

The ordering that loop actually protected is **per-conversation**: a run emits block replies, a possible `Provider busy, retrying...` notice, and the final answer to one `ChatID`, and those must arrive in that order. There are 41 `PublishOutbound` call sites, several of which target the same chat within one run.

A *global* order was never needed. Ordering across different conversations carries no meaning.

## Change

Dispatch now shards on `channel + ChatID`:

- A conversation always maps to the same shard and is delivered serially there → ordering guarantee unchanged.
- Unrelated conversations proceed in parallel.
- `GOCLAW_OUTBOUND_SHARDS` tunes worker count (default 8).

The channel name is part of the shard key because `ChatID`s are only unique within a channel.

### Temp media needed a real fix alongside it

Serial dispatch deduped temp media *implicitly*: the first delivery removed the file, and a later message carrying the same path found `os.Stat` failing and skipped it. The existing comment (`"already sent by another dispatch"`) states the assumption.

Across concurrent shards that check alone is a TOCTOU race — two shards can both stat a live file and upload it twice. Claims are now taken atomically and released after the send. Non-temp (workspace) media passes through unclaimed, since it is never removed after delivery.

## Also included

Pool limits that bound this same path, made configurable with **defaults unchanged**:

| Env | Default |
|-----|---------|
| `GOCLAW_PG_MAX_OPEN_CONNS` | 25 |
| `GOCLAW_PG_MAX_IDLE_CONNS` | 10 |
| `GOCLAW_HTTP_MAX_IDLE_CONNS` | 100 |
| `GOCLAW_HTTP_MAX_IDLE_CONNS_PER_HOST` | 10 |

The per-host limit binds when one provider takes nearly all traffic: every concurrent request past it pays a fresh TCP+TLS handshake instead of reusing a pooled connection. A deployment raising `LaneMain` needs to be able to lift it. Self-hosted / in-network providers have no reason to be throttled at 10.

## Tests

`internal/channels/dispatch_shard_test.go`:

- shard mapping is stable and bounded; channel name participates in the key
- **50 messages to one chat arrive strictly in order**
- **a stalled conversation does not block an unrelated one** — the regression this change exists for; one chat's `Send` is held hostage and the other must still be delivered
- 16 goroutines racing to claim the same temp file → exactly one wins
- non-temp media passes through unclaimed; missing files are skipped
- releasing drops both the file and the claim, so the claim set stays bounded

## Verification

- `go test -race` green for `channels`, `scheduler`, `store/pg`, `providers`

Note: `internal/channels/telegram` `TestDisplayWidth` fails on this branch, but it fails identically on a pristine `origin/dev` worktree — pre-existing and unrelated.

### DIFF
--- a/docs/00-architecture-overview.md
+++ b/docs/00-architecture-overview.md
@@ -380,6 +380,35 @@ flowchart TD
 | `team` | 100 | `GOCLAW_LANE_TEAM` | Agent team/delegation executions |
 | `cron` | 30 | `GOCLAW_LANE_CRON` | Scheduled cron jobs |
 
+### Scaling Beyond the Lane Defaults
+
+`main` bounds how many users can be answered at once, but raising it alone just
+moves the queue downstream. These knobs need to move with it:
+
+| Setting | Default | Env Override | Why it binds |
+|---------|:-------:|--------------|--------------|
+| Postgres max open conns | 25 | `GOCLAW_PG_MAX_OPEN_CONNS` | Held per query, not per run — undersizing shows up as burst latency when many runs load context at once |
+| Postgres max idle conns | 10 | `GOCLAW_PG_MAX_IDLE_CONNS` | Clamped to max open |
+| Provider idle conns | 100 | `GOCLAW_HTTP_MAX_IDLE_CONNS` | Total across all provider hosts |
+| Provider idle conns per host | 10 | `GOCLAW_HTTP_MAX_IDLE_CONNS_PER_HOST` | When one provider takes nearly all traffic, every request past this limit pays a fresh TCP+TLS handshake. Self-hosted/in-network providers should match `main`. |
+| Outbound dispatch shards | 8 | `GOCLAW_OUTBOUND_SHARDS` | See below |
+
+Channel-side rate limits are metered by the *remote* platform and are not
+covered by any of the above — DingTalk's AI Card API, for instance, meters per
+app, so `card_update_interval_ms` decides how many conversations can stream
+concurrently regardless of local capacity.
+
+### Outbound Dispatch
+
+Outbound delivery is sharded by `channel + ChatID`. A conversation always maps
+to the same shard and is delivered serially there, so the ordering a run
+depends on — block replies, retry notices, then the final answer — is
+preserved. Unrelated conversations run in parallel, so one slow send (a media
+upload of several seconds) no longer stalls every other reply in the process.
+
+Raising `GOCLAW_OUTBOUND_SHARDS` past the point where the remote API meters the
+gateway only moves queueing from this process into theirs.
+
 ### Session Queue Concurrency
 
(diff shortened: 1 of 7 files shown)


clark-cant (approved):
## Maintainer Review — github-maintain

**Summary:** Shards outbound dispatch by channel+ChatID to eliminate the single-goroutine bottleneck while preserving per-conversation ordering. Also fixes a TOCTOU race in temp media deduplication and makes connection pool limits configurable.

**Risk level:** Low-Medium — touches core dispatch path but well-tested with comprehensive regression coverage.

### Mandatory gates

| Gate | Result |
|------|--------|
| Duplicate / prior work | Clear — no prior PR addresses dispatch sharding |
| Project standards | Excellent — follows existing patterns, well-documented |
| Strategic necessity | High — removes a real throughput ceiling that blocks all outbound traffic behind the slowest send |

### Code review

**Correctness:** Solid. The shard key (channel+ChatID) correctly preserves the ordering guarantee that runs depend on — block replies, retry notices, and final answer to one ChatID must arrive in order. Unrelated conversations proceed in parallel as intended.

**Temp media fix:** The atomic claim mechanism (`sync.Map` LoadOrStore) properly closes the TOCTOU race that serial dispatch masked implicitly. The claim set is bounded by the send duration, so it won't grow unbounded.

**Pool configurability:** Sensible defaults unchanged. The documentation explaining when each limit binds (per-host for single-provider deployments, per-query not per-run for Postgres) is excellent operator guidance.

**Test coverage:** Comprehensive — 50-message ordering test, slow-chat-does-not-block regression test (the exact scenario this change exists for), 16-goroutine temp media race test, non-temp passthrough verification.

**Quality:**
- Code is well-structured with clear separation (shard index calculation, claim/release helpers)
- Comments explain *why* not just *what*
- No anti-slop patterns: no unnecessary abstractions, no defensive paranoia
- Documentation in `docs/00-architecture-overview.md` is operator-facing and actionable

### Verdict: APPROVE ✅

This is a high-quality performance improvement that addresses a real bottleneck. The code is well-tested, well-documented, and follows project conventions. Safe to merge once CI is green.

*Reviewed by github-maintain at 2026-07-31T14:58Z*

PR #1367, opened 2026-07-05
feat(providers): allow LAN Ollama hosts via GOCLAW_OLLAMA_ALLOWED_HOSTS

bclermont:
Problem

  Provider validation hardcodes an allowlist for ollama/acp provider types, only permitting localhost, 127.0.0.1, ::1, and host.docker.internal. Any other host (e.g. a LAN-hosted Ollama server) is rejected with:
  provider type "ollama" only allows localhost URLs (localhost, 127.0.0.1, ::1, host.docker.internal), got host "192.168.3.31"
  There was no way to whitelist additional hosts short of a code change.
  
  Change

  - Added GOCLAW_OLLAMA_ALLOWED_HOSTS — a comma-separated env var of additional hostnames/IPs accepted for ollama/acp provider URL validation, on top of the existing hardcoded localhost allowlist.
  - Case-insensitive host matching, consistent with existing comparison style in validateProviderURL.
  - Follows the same lazy-load (sync.OnceValue) pattern as the existing GOCLAW_ALLOW_PRIVATE_PROVIDER_URLS var.

  Files changed

  - internal/http/providers.go — new env parser + extended allow-check in validateProviderURL, updated error message/doc comment.
  - internal/http/providers_url_validate_test.go — new TestValidateProviderURL_OllamaAllowedHosts covering unset/rejected, configured/accepted, not-in-list/rejected, and localhost-still-works cases; existing test updated to pin default (unset) behavior.

  Verification

  - go build ./... ✅
  - go build -tags sqliteonly ./... ✅
  - go vet ./... ✅
  - go test ./internal/http/... ✅ (all pass, no regressions)

  Surface parity

  - Gateway server: ✅ implemented.
  - API contract: N/A — env var only, no request/response shape change.
  - Web UI: N/A — this is an operator/env-level config, not exposed in provider settings UI (raw SQL/API edits already required a restart for this path; no UI surface for env vars).
  - CLI/runtime package: N/A — no CLI command touches provider URL validation.

  Usage

  GOCLAW_OLLAMA_ALLOWED_HOSTS=192.168.3.31

PR #1465, opened 2026-07-22
feat(security): let operators un-block CIDRs behind a transparent proxy

connermo:
## Problem

SSRF protection resolves a hostname and judges the resulting IP. That model assumes **DNS resolution describes where the traffic actually goes** — which stops being true behind a TUN / fake-IP proxy.

In fake-IP mode the proxy answers every DNS query with a synthetic address out of a reserved range, then routes that address to the real public host. The resolved IP is a *handle*, not a destination.

Observed on a machine behind such a proxy:

```
WARN tool error tool=web_fetch error="SSRF protection: hostname news.sina.cn resolves to private IP 198.18.0.236"
WARN tool error tool=web_fetch error="SSRF protection: hostname best-ai.org  resolves to private IP 198.18.0.234"
```

All three were verified reachable from the same container — `HTTP 200` — so the network path was fine; only the check rejected them.

The block list is compiled in, so **no configuration can fix this**, and `web_fetch` is simply unusable. Worse, the agent keeps burning iterations retrying URLs that can never succeed: one turn took 85s with 15 tool calls, 3 of which were guaranteed failures.

## Change

`GOCLAW_SSRF_ALLOWED_CIDRS` lets an operator name the ranges their proxy hands out:

```
GOCLAW_SSRF_ALLOWED_CIDRS=198.18.0.0/15
```

- **Empty by default** — nothing changes for deployments that do not set it.
- Accepted *and* refused entries are logged at startup. This widens what LLM- and admin-supplied URLs can reach, so it should be visible rather than silent.

### Ranges that can never be allowlisted

Link-local (including cloud metadata at `169.254.169.254`), multicast, and unspecified are refused at parse time — checked **in both directions**, so neither an exact entry nor a wider range that swallows one gets through. `0.0.0.0/0` is refused.

This mirrors what `exemptableCIDRs` already declines to cover:

> an allowlist entry must never open a path to the cloud metadata endpoint

Verified against the running binary with a deliberately dangerous value:

```
WARN security.ssrf_allowlist          allowed=[198.18.0.0/15]
WARN security.ssrf_allowlist_rejected rejected="[169.254.0.0/16 (overlaps 169.254.0.0/16)]"
```

## Two implementation notes

**Applied inside `isBlocked`, not just `validate()`.** `NewSafeClient` re-checks the pinned IP at dial time through the same function, so relaxing only the pre-flight check would pass validation and then fail to connect.

**`internal/tools` has its own SSRF gate.** `web_shared.go` carries a separate private-range list used by `web_fetch` / `web_search`, and the two lists are not identical (`tools` has `100.64.0.0/10` and `fec0::/10`; `security` has `224.0.0.0/4`). It must consult the same operator setting, or relaxing one gate leaves the other rejecting exactly the traffic the operator permitted — this was caught during testing, when the first version of the change had no effect on `web_fetch` at all.

Unifying the two lists is deliberately **not** attempted here; it is a wider and riskier change. Flagging it as pre-existing debt: adding a range to one list today silently misses the other.

## Tests

`internal/security/ssrf_operator_allowlist_test.go` (6 cases) and `internal/tools/web_shared_ssrf_allowlist_test.go` (2 cases):

- empty / whitespace / comma-only specs are inert
- the benchmarking range is accepted
- link-local, cloud-metadata host, a wider range swallowing link-local, `0.0.0.0/0`, multicast, unspecified, IPv6 link-local are all refused
- malformed entries are reported, not silently dropped, while valid siblings still apply
- `isBlocked` and `Validate` honour the allowlist while keeping metadata and RFC 1918 blocked
- `CheckSSRF` (the `tools` gate) honours it too, and still blocks metadata / RFC 1918 / loopback

## Verification

- `go test -race` green for `security`, `tools`, `mcp`
- End-to-end against a live gateway: `news.sina.cn` fetched successfully; `169.254.169.254` and `10.0.0.1` still rejected

### DIFF
--- a/cmd/gateway_lifecycle.go
+++ b/cmd/gateway_lifecycle.go
@@ -16,6 +16,7 @@ import (
 	"github.com/nextlevelbuilder/goclaw/internal/heartbeat"
 	"github.com/nextlevelbuilder/goclaw/internal/sandbox"
 	"github.com/nextlevelbuilder/goclaw/internal/scheduler"
+	"github.com/nextlevelbuilder/goclaw/internal/security"
 	"github.com/nextlevelbuilder/goclaw/internal/store"
 	"github.com/nextlevelbuilder/goclaw/internal/tasks"
 	"github.com/nextlevelbuilder/goclaw/internal/tools"
@@ -293,6 +294,16 @@ func (d *gatewayDeps) runLifecycle(
 		slog.Info("cors: allowed_origins configured", "origins", d.cfg.Gateway.AllowedOrigins)
 	} else if !edition.Current().IsLimited() {
 		slog.Warn("security.cors_open: no allowed_origins configured — all WebSocket origins accepted. Set gateway.allowed_origins or GOCLAW_ALLOWED_ORIGINS for production")
+	}
+	if allowed, rejected := security.OperatorAllowlistStatus(); len(allowed) > 0 || len(rejected) > 0 {
+		if len(allowed) > 0 {
+			slog.Warn("security.ssrf_allowlist: SSRF protection relaxed for operator-configured ranges — tool- and admin-supplied URLs may reach them",
+				"env", security.SSRFAllowedCIDRsEnv, "allowed", allowed)
+		}
+		if len(rejected) > 0 {
+			slog.Warn("security.ssrf_allowlist_rejected: entries refused; cloud-metadata, multicast and unspecified ranges can never be allowlisted",
+				"env", security.SSRFAllowedCIDRsEnv, "rejected", rejected)
+		}
 	}
 
 	if err := d.server.Start(ctx); err != nil {
--- a/internal/security/ssrf.go
+++ b/internal/security/ssrf.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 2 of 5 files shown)


clark-cant (approved):
## Review — PR #1465: feat(security): let operators un-block CIDRs behind a transparent proxy

**Risk level:** Medium (security surface widening, opt-in with strong guardrails)

**Mandatory gates:**
- Duplicate / prior implementation: clear — #1248 (MCP host allowlist) and #1367 (Ollama hosts) cover different surfaces (hostname-based, not CIDR-based). No prior CIDR-based operator escape hatch exists.
- Project standards: clear — follows existing `internal/security` patterns, comments are thorough, test coverage is real.
- Strategic necessity: clear value — operators behind TUN/fake-IP proxies have no workaround today; this is a real deployment blocker.

**Findings:**

*Suggestion:*
- The `internal/tools` vs `internal/security` SSRF list divergence is correctly flagged as pre-existing debt. The PR body documents this well. No action needed here.

**Verdict: Approve**

Well-designed escape hatch with strong guardrails: never-allowlistable ranges (cloud metadata, multicast, unspecified), bidirectional overlap check, opt-in via env var, startup logging of both accepted and rejected entries. Applied inside `isBlocked` so both pre-flight and dial-time checks honor the allowlist. Tests are thorough (8 cases covering empty specs, benchmarking range acceptance, protected range refusal, malformed entry reporting, and end-to-end validation). The `SetOperatorAllowlistForTest` helper correctly documents the non-parallel requirement.

*Posted by github-maintain at 2026-07-31T13:28:25Z*

PR #1468, opened 2026-07-23
fix(tools): return authoritative datetime fields

itsddvn:
## Summary
- return server-computed UTC and local calendar fields, including weekday and ISO weekday
- mark calendar fields as authoritative so Qwen does not infer weekdays incorrectly from date strings
- keep existing `utc`, `unix_ms`, `local`, and `timezone` fields unchanged, making the response additive and backward compatible
- add deterministic coverage for local timezone output, UTC-only output, invalid timezones, and zero-value/nil receivers

## Root cause
The datetime tool returned timestamps but not explicit calendar fields. Models such as Qwen therefore inferred the weekday from the date and could produce an incorrect answer. Computing these fields in Go makes the tool result authoritative.

## Validation
- [x] `go test ./internal/tools -run ^'^TestDateTimeTool' -count=1`
- [x] `go test ./internal/tools ./internal/agent ./internal/pipeline -count=1`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `git diff --check`

## Surface parity
- Gateway server: updated builtin datetime tool response
- API contract: additive tool-result fields; existing fields preserved
- Web UI: N/A because the UI does not parse datetime tool payloads
- CLI/runtime package: N/A because the builtin tool is executed server-side

### DIFF
--- a/internal/tools/datetime.go
+++ b/internal/tools/datetime.go
@@ -10,9 +10,11 @@ import (
 // DateTimeTool provides precise current date/time with timezone support.
 // Models use this before creating cron jobs or any time-sensitive operation
 // instead of guessing timestamps from the system prompt's date-only field.
-type DateTimeTool struct{}
+type DateTimeTool struct {
+	now func() time.Time
+}
 
-func NewDateTimeTool() *DateTimeTool { return &DateTimeTool{} }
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
Approve: additive authoritative UTC/local calendar fields with deterministic tests. No Critical or Important findings; duplicate gate clear, project conventions followed, and all reported CI checks are green.

PR #1469, opened 2026-07-23
fix(sessions): store last-call context tokens instead of run-cumulative total

thotam:
## Summary
The sessions list "Context" bar (and compaction calibration) read `last_prompt_tokens`, but the value stored was the run-cumulative `TotalUsage.PromptTokens` — summed across every think→act→observe iteration — instead of the last LLM call's actual context size. Tool-heavy runs displayed e.g. **901K / 258K** and over-triggered post-run compaction (observed: 83 compactions on a 4-message session).

## Type
- [x] Bug fix

## Target Branch
`dev`

## Root cause & fix
- `ThinkStage` accumulates `TotalUsage.PromptTokens` per iteration; `FinalizeStage` passed that total to `UpdateMetadata` → `SetLastPromptTokens`, whose contract is *"actual prompt tokens from the last LLM response"*.
- Once the inflation exceeded the 40% overhead clamp in `estimateOverhead`, `maybeSummarize` triggered compaction regardless of real history size.
- Fix: snapshot the final iteration's own usage (`ThinkState.LastUsage`), and persist its real context size via new `Usage.ContextTokens()`:
  - `prompt + cache_read + cache_creation` when the provider excludes cached segments from `input_tokens` (Anthropic-style), plain `prompt` otherwise (OpenAI-style);
  - `+ completion tokens`, since the final reply joins history and occupies the next request's prompt — consistent with the `msgCount` passed alongside, which already counts the flushed reply.
- The run-cumulative total still feeds `AccumulateTokens` (billing) unchanged. In-loop pruning (real BPE counts) and emergency compaction (API overflow signal) were never affected.

Existing sessions keep the stale metadata value only until their next run: `UpdateMetadata` runs before `MaybeSummarize` in `FinalizeStage`, so the corrected value is in place before the next compaction decision.

Surface parity: Web UI / API contract / CLI N/A — `estimatedTokens` field shape unchanged, only the backend-written value now matches its semantics; desktop (SQLite) shares the same pipeline path.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] Web UI builds: `cd ui/web && pnpm build` (if UI changes) — N/A, no UI changes
- [x] SQL queries use parameterized `$1, $2` (no string concat) — N/A, no SQL changes
- [x] New user-facing strings added to all 3 locales (en/vi/zh) — N/A, no user-facing strings
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration) — N/A, no migration

## Test Plan
TDD — new tests written failing-first against the buggy behavior:
- `internal/providers/usage_context_tokens_test.go` — `Usage.ContextTokens()` for Anthropic-style vs OpenAI-style cache accounting
- `internal/pipeline/think_stage_last_usage_test.go` — `LastUsage` holds only the final iteration's usage (not the sum); kept when the final response carries no usage
- `internal/pipeline/finalize_stage_metadata_test.go` — `UpdateMetadata` receives last-call usage separately from the run-cumulative total

Verified locally: `go build ./...`, `go build -tags sqliteonly ./...`, `go vet`, `go test -race` on `internal/{pipeline,agent,providers,sessions,store/...}` (sqlitestore with `-tags sqliteonly`) — all pass.

### DIFF
--- a/internal/agent/loop_pipeline_callbacks.go
+++ b/internal/agent/loop_pipeline_callbacks.go
@@ -93,7 +93,7 @@ type pipelineCallbackSet struct {
 	checkReadOnly      func(state *pipeline.RunState) (*providers.Message, bool)
 	sanitizeContent    func(string) string
 	flushMessages      func(ctx context.Context, sessionKey string, msgs []providers.Message) error
-	updateMetadata     func(ctx context.Context, sessionKey string, usage providers.Usage, msgCount int) error
+	updateMetadata     func(ctx context.Context, sessionKey string, usage, lastUsage providers.Usage, msgCount int) error
 	bootstrapCleanup   func(ctx context.Context, state *pipeline.RunState) error
 	maybeSummarize     func(ctx context.Context, sessionKey string)
 }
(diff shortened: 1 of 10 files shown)


clark-cant (approved):
Reviewed in cron-safe maintenance run: no Critical/Important findings; duplicate/prior-work gate clear; project standards gate satisfied by repository docs and conventions; strategic value clear for correcting session context/compaction metadata. CI checks are green.

PR #1478, opened 2026-07-24
fix: persist mid-loop compaction to stop re-compaction loop and revive episodic memory

nguyenha935:
## Summary

Mid-loop context compaction (`prune_stage` + the final-request guard) only mutates the run's in-memory message buffer, never the session store. Every turn reloads full history and re-compacts from scratch: `message_tokens` climb 129k→156k across turns even though each turn compacts back down to ~60k. This is wasteful, and because each run sees a different lossy compaction the agent degrades.

The same missing persistence stalls episodic memory. The cumulative compaction count never advances, so the episodic worker's idempotency key (`sessionKey:count`) stays pinned and every cycle after the first is skipped. Observed on live traffic: 8 `run.completed` since deploy, **0** new episodic summaries.

This PR fixes both by reusing the existing `maybeSummarize` → `TruncateHistory` + `IncrementCompaction` path — **no new store methods, no migrations**.

- **Bug A (cumulative count):** `emitSessionCompleted` reads cumulative `GetCompactionCount` (matching the legacy v2 emit path) instead of the per-run counter that resets to 0 each run.
- **Bug B / anti-loop:** `finalize` passes `state.Prune.MidLoopCompacted` into `maybeSummarize`; under mid-loop pressure it lowers the trigger to a unit-aligned threshold (`compactionInputCap - overhead`, the same `MaxRequestShare` the guard uses) so the compaction is **persisted**. A defensive floor prevents over-compaction on pathological config, and tool-result-only bloat (history-only estimate under threshold) still skips — so we never truncate a session that didn't actually grow.
- **Bug C (dedup):** `SourceID` embeds the count (`sessionKey:count`) so the eventbus dedup key advances per compaction cycle instead of swallowing rapid same-session turns within the 5m TTL. Safe for the worker, which builds its own idempotency key from the payload and never parses `SourceID`.

## Type

- [x] Bug fix

## Target Branch

`dev`

## Checklist

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` (pipeline, agent, consolidation, tokencount, usage, tools, store — all `ok`)
- [x] SQL queries use parameterized `$1, $2` (no new SQL — reuses existing store methods)
- [x] Migration version bumped in `internal/upgrade/version.go` — N/A (no new migration)

## Test Plan

- **Bug A:** unit test asserts emit reads cumulative `GetCompactionCount`, not the per-run counter.
- **Bug B / anti-loop:** a run with `MidLoopCompacted=true` triggers `maybeSummarize` below the old threshold → `TruncateHistory` + `IncrementCompaction` exactly once; per-session `TryLock` still blocks concurrent summarize.
- **No over-compaction:** `MidLoopCompacted=true` but history-only estimate ≤ pressure threshold (tool-result bloat) → still SKIP, no session truncation.
- **Threshold alignment:** with the same history, pressure threshold (`compactionInputCap - overhead`) < baseline floor, so a history between the two marks persists under pressure and skips otherwise — asserts the boundary.
- **Bug C:** two `session.completed` with same `sessionKey` but different `count` both pass dedup; same `count` → only one passes.
- **Media preservation** unchanged when truncating via the pressure branch.

### Post-deploy verification (read DB, not Debug logs)

- `episodic_summaries` gains rows with `source_id = sessionKey:N`, N increasing, for sessions under compaction pressure.
- `sessions.compaction_count` advances instead of standing still.
- Guard log: `message_tokens` at run start no longer climbs monotonically across turns (the re-compaction loop is broken).


clark-cant (approved):
Reviewed against CONTRIBUTING.md and repository architecture. The PR addresses the reported compaction persistence and episodic dedup bugs with focused tests; CI (go, web, release-versioning) is green and merge state is CLEAN. No Critical or Important findings identified.

PR #1479, opened 2026-07-25
test(pipeline): lock mid-loop compaction pressure flag end-to-end

nguyenha935:
## Summary

Adds an end-to-end pipeline test that locks the mid-loop-compaction → post-turn-persist handoff, closing a coverage gap left by #1478.

`FinalizeStage` passes `state.Prune.MidLoopCompacted` into `MaybeSummarize` so a compaction that fired mid-loop gets PERSISTED (`TruncateHistory` + `IncrementCompaction`) instead of being thrown away and redone next turn. Nothing asserted that the flag actually arrives: the existing `TestFinalizeStage_MaybeSummarize_Called` discards it (`func(_ context.Context, sessionKey string, _ bool)`), so hardcoding the argument to `false` — which silently re-introduces the per-turn re-compaction loop and stalls the cumulative compaction count episodic depends on — keeps the whole suite green.

This test drives the real `NewDefaultPipeline().Run()` (ContextStage → PruneStage → ThinkStage → … → FinalizeStage) with a scripted LLM and a history large enough to force the prune-stage compaction, then asserts the flag observed by `MaybeSummarize` is `true`. A companion case asserts the negative: a small history compacts nothing and reports `false`, so the flag can't just be pinned true.

**Mutation-verified.** With `MaybeSummarize(ctx, key, false)` hardcoded in `finalize_stage.go`:

| suite | result |
|---|---|
| new test | **exit 1 (fails — regression caught)** |
| pre-existing pipeline + agent tests | exit 0 (passes — gap confirmed) |

Test-only change; no production code touched.

## Type
- [ ] Feature
- [x] Bug fix (test-only: regression guard)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev` — per CONTRIBUTING, all features/bugfixes target `dev`.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` (`internal/pipeline`, `internal/agent`)
- [x] SQL queries use parameterized `$1, $2` (no string concat) — no SQL in this change
- [x] New user-facing strings added to all 3 locales (en/vi/zh) — no user-facing strings
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration) — no migration
- [x] Web UI builds — no UI change

## Test Plan

`internal/pipeline/compaction_pressure_e2e_test.go`, both cases running the full pipeline via `NewDefaultPipeline().Run()`:

1. **`TestPipelineE2E_MidLoopCompaction_PropagatesPressureToMaybeSummarize`** — history of 50 messages × 100 tokens (5000) against `budget = ContextWindow(1000) − overhead(100) − MaxTokens(100) = 800` forces prune-stage compaction. Asserts, in order: `state.Prune.MidLoopCompacted` is set; `CompactMessages` ran; `MaybeSummarize` was called exactly once **with `midLoopCompacted=true`**.
2. **`TestPipelineE2E_NoCompaction_ReportsNoPressure`** — a 3-message history stays under budget. Asserts `CompactMessages` never ran and `MaybeSummarize` received `false`.

Run logs confirm the real stages executed rather than a stubbed path (`context.preflight_budget … history_tokens=5000 budget=800`, `final_context.guard … action=allow`).

Verified with `-race -count=1`; also ran the pre-existing `internal/pipeline` and `internal/agent` suites to confirm no regression.

### DIFF
--- a/internal/pipeline/compaction_pressure_e2e_test.go
+++ b/internal/pipeline/compaction_pressure_e2e_test.go
@@ -1,0 +1,133 @@
+package pipeline
+
+import (
+	"context"
+	"testing"
+
+	"github.com/nextlevelbuilder/goclaw/internal/providers"
+)
+
+// This file closes the last uncovered link in the anti-recompaction-loop chain.
+//
+// The chain has three hops:
+//  1. prune_stage / final_request_guard compact mid-loop and set
+//     state.Prune.MidLoopCompacted  — covered by
+//     TestPruneStage_Compaction_PreservesPending + stages_test.go.
+//  2. finalize_stage forwards that flag into deps.MaybeSummarize             — THIS FILE.
+//  3. maybeSummarize lowers its threshold and PERSISTS the compaction        — covered by
+//     internal/agent/loop_maybe_summarize_pressure_test.go.
+//
+// Hop 2 was effectively untested: TestFinalizeStage_MaybeSummarize_Called binds
+// the flag as `_ bool` and never asserts it, so hardcoding `false` at the
+// finalize call site would keep every existing test green while silently
+// restoring the unbounded re-compaction loop (each turn recompacts from scratch
+// because nothing is ever written back to the session store).
+//
+// These tests drive the REAL NewDefaultPipeline().Run() end to end — no stage is
+// invoked directly — so the flag has to survive the actual run wiring.
+
+// pressureE2EDeps builds deps for a full-pipeline run whose token math forces a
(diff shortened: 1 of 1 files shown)


clark-cant (approved):
Approve: test-only regression coverage for the mid-loop compaction pressure handoff. The full-pipeline E2E cases cover both true and false propagation, CI is green, merge state is CLEAN/MERGEABLE, and no Critical or Important findings were found.

PR #1379, opened 2026-07-06
feat: add team work classification routing

nguyenha935:
## Summary
Adds an optional Team Work Classification flow that uses embedding evidence plus an LLM arbiter to decide whether an agent should handle a request directly or route it through team/delegate workflow.

The routing is permission-aware: team leads may assign tasks, while team members can only route through `team_tasks` when member requests and auto-dispatch are enabled; otherwise the agent falls back to self-handling.

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Hotfix (targeting `main`)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`

## Checklist
- [x] `go build ./...` passes
- [x] `go build -tags sqliteonly ./...` passes (if Go changes)
- [x] `go vet ./...` passes
- [ ] Tests pass: `go test -race ./...`
- [x] Web UI builds: `cd ui/web && pnpm build` (if UI changes)
- [x] No hardcoded secrets or credentials
- [x] SQL queries use parameterized `$1, $2` (no string concat)
- [x] New user-facing strings added to all 3 locales (en/vi/zh)
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration)

## Test Plan
- `git apply --check /tmp/team-work-classify.patch` on clean `upstream/dev`
- `git diff --cached --check`
- `docker run --rm -v /tmp/goclaw-pr-team-work-classify:/src -v /tmp/goclaw-go-cache:/tmp/go-cache -v /tmp/goclaw-go-mod:/tmp/go-mod -w /src -e GOCACHE=/tmp/go-cache -e GOMODCACHE=/tmp/go-mod golang:1.26 go test ./internal/teamworkclassify ./internal/agent ./internal/gateway/methods ./cmd ./internal/pipeline`
- `docker run --rm -v /tmp/goclaw-pr-team-work-classify:/src -v /tmp/goclaw-go-cache:/tmp/go-cache -v /tmp/goclaw-go-mod:/tmp/go-mod -w /src -e GOCACHE=/tmp/go-cache -e GOMODCACHE=/tmp/go-mod golang:1.26 go build ./...`
[...]

PR #1485, opened 2026-07-28
feat(teams): enforce classified team work through durable workflows

nguyenha935:
## Summary

Team work classification (#1379) decided **how** a request should be routed — self, single owner, or multi-role — but nothing downstream held the system to that decision. A multi-role request could be classified correctly and still be answered by one agent, and a routing decision left no record once the turn ended. This PR makes the classification binding and auditable.

## Type
- [x] Feature

## Target Branch
`dev`

## What changes

- **Durable workflows.** A multi-role decision materialises as a `team_workflows` row plus one `team_tasks` row per step, so the plan survives a restart and can be inspected, actioned and refreshed rather than living only in a turn's memory.
- **Classification audit.** Every decision records the shape that was *verified*, the mode that was *requested* versus the one that took *effect*, and the stage at which anything degraded — so a wrong route can be explained after the fact instead of guessed at.
- **Independent review.** A step's critic is never the agent that produced the step, and a workflow only reaches a terminal state once its result has actually been delivered.
- **Plans shaped by the real roster.** The plan is built from the team that exists rather than a fixed step template, and the planner is told which roles cannot own a step so it stops assigning work to the coordinator.
- **Failure handling that does not lose work.** A step hitting a transient provider error is requeued within a dispatch budget instead of discarding a validated plan; a step that produced no usable result is not settled as complete; and internal provider error text never reaches the requester's delivery.
- **Label reconciliation.** Role and decision synonyms are normalised rather than used as grounds to throw away an otherwise valid plan.

## Surface parity

| Surface | Covered |
|---|---|
| Gateway (RPC + HTTP) | 36 files under `cmd/`, handlers + workflow actions |
| PG migrations | `000096`–`000098`; `RequiredSchemaVersion` 95 → 98 |
| SQLite | `schema.sql` + incremental patches in `schema.go`; `SchemaVersion` 58 → 61 |
| Web UI | 40 files — team board, task detail, workflow views, types |
| i18n | 6 new keys in **all five** catalogs (en/vi/zh/ko/ru) |

New tables `team_workflows` and `team_work_classification_audits` both carry `tenant_id NOT NULL REFERENCES tenants(id) ON DELETE CASCADE`, and every index leads with `tenant_id`. All new store queries are parameterized; no string-concatenated SQL.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` — see note below
- [x] SQL queries use parameterized `$1, $2`
- [x] New user-facing strings added to all locales
- [x] Migration version bumped in `internal/upgrade/version.go`

**On `go test -race ./...`:** passes. (An earlier version of this note claimed six `internal/http` Ollama tests fail as pre-existing debt. That was misleading: they fail only inside my local `golang:1.26-bookworm` container, where `host.docker.internal` does not resolve. `internal/http` passes on the CI runner.)

## Test Plan

66 new test files covering the classifier decision paths, the workflow store (PG + SQLite), step dispatch and settlement, and the finalization/delivery paths.

Beyond unit coverage, this has been running on a live deployment driven through the real `chat.send` ingress (not `/v1/chat/completions`, which bypasses the classifier gate), exercising all three routing shapes plus the member-in-a-team case. Several of the fixes above came directly out of that testing — the requeue-on-transient-failure path, the no-usable-result guard, and the error-text redaction were each found by observing real workflows rather than by unit tests.

**Status note for reviewers:** live acceptance is still in progress on my side. I am opening this now so review can start on the design and the surface coverage; I will report any further findings from live testing on this PR rather than amending silently.


nguyenha935:
CI caught a real inconsistency in the first push, now fixed in 68fae696.

The six new `MsgWorkflow*` keys had been added to `catalog_ko.go` and `catalog_ru.go` as untranslated English while `vi` and `zh` were translated, and two tests were written against that English text. I translated both catalogs while preparing this PR but did not update the tests, so `TestWorkflowActionMessagesHaveFiveLocaleStructuralParity` and `TestWorkflowHandlersLocalizePhase8ValidationAcrossFiveLocales/{ko,ru}` failed.

I kept the translations rather than reverting them, because the same test table already expects translated ko/ru text for the invalid-ID message (`잘못된 teamId ID`, `неверный ID teamId`) — so these are not English-fallback locales, the six workflow keys were simply missed. The tests now assert the translations.

`internal/i18n` and `internal/gateway/methods` both pass locally with `-race`. The only remaining failure in a full `go test -race ./...` is `internal/http`, which is the pre-existing Ollama-URL set described in the PR body — it fails identically on a clean `upstream/dev` worktree.

nguyenha935:
Correction to the note in the PR description about `internal/http`.

I wrote that six `TestProvidersHandlerListProviderModelsOllama*` tests fail and that this is pre-existing. The baseline comparison was right — they fail identically on a clean `upstream/dev` worktree — but my framing was misleading, because it implies CI would fail too. It does not: `internal/http` passes on the CI runner (`ok ... 6.082s`).

The failures are an artifact of **my** local harness only. I run the Go gates inside a `golang:1.26-bookworm` container, where these tests try to reach `host.docker.internal`, which does not resolve. CI runs directly on the Linux runner, so `localhost` works and the tests pass.

So there is no pre-existing failure for reviewers to discount here. The only CI failure on this PR was the ko/ru i18n one, fixed in 68fae696.

PR #933, opened 2026-04-16
fix(providers): strip Gemma 4 reasoning leak

badgerbees:
## Summary
Fixes Gemma 4+ reasoning leaks by treating those models as known reasoning leakers when `reasoning_effort` resolves to `off`, so the existing strip-thinking path suppresses internal thoughts before they reach users. This also adds regression coverage for the Gemma 4 family while keeping the Gemini tool-response `name` handling intact.

Fixes Issue #843 and #844.

## Type

- [ ] Feature
- [x] Bug fix
- [ ] Hotfix (targeting `main`)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev`

## Checklist
- [x] `go build ./...` passes
- [ ] `go build -tags sqliteonly ./...` passes (if Go changes)
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...`
- [ ] Web UI builds: `cd ui/web && pnpm build` (if UI changes)
- [ ] No hardcoded secrets or credentials
- [ ] SQL queries use parameterized `$1, $2` (no string concat)
- [ ] New user-facing strings added to all 3 locales (en/vi/zh)
- [ ] Migration version bumped in `internal/upgrade/version.go` (if new migration)

## Test Plan
`go test ./internal/providers -run "TestResolveReasoningDecisionStripsForLeakyModels|TestModelLeaksReasoning|TestBuildRequestBody_ToolMessageIncludesName|TestBuildRequestBody_ToolMessageNameLookupUsesRawID|TestBuildRequestBody_ToolMessageWithoutMatchingCallOmitsName" -count=1`

Verified the Gemma 4 reasoning allowlist change and the Gemini tool-response name handling path with targeted provider tests.

Issue #843, opened 2026-04-11
[Bug] HTTP 400: function_response.name cannot be empty – Gemma4's strict tool calling differs from other AI models


ptran17203:
An INVALID_ARGUMENT error (HTTP 400) occurs when the orchestrator sends a tool response back to the Gemma4 API without providing the function name. Unlike many other AI models (e.g., GPT-4, Claude, etc.), Gemma4 enforces a stricter contract: every function_response part must include a name field that exactly matches the function_call.name previously requested by the model. When the orchestrator (built for more lenient models) omits this field or sends an empty value, the API rejects the request.

Error Log:

text
HTTP 400: gemini-native: [{ 
  "error": { 
    "code": 400, 
    "message": "* GenerateContentRequest.contents[3].parts[0].function_response.name: Name cannot be empty.\n", 
    "status": "INVALID_ARGUMENT" 
  } 
}]
(The same error repeats in the log.)

Root Cause Analysis:

Other AI models (OpenAI, Anthropic, etc.) typically accept function_response with just the result payload; the name field is often optional or inferred from context.

Gemma4 (and other Gemini-native models) requires an explicit, non-empty name field in every function_response that matches the corresponding function_call.name. This is part of its native tool-calling state machine.

The current orchestrator does not populate the name field when forwarding tool results back to Gemma4, leading to a validation error at the API level.

Impact:

The tool-calling loop breaks entirely.

Any workflow requiring sequential or parallel tool calls fails with HTTP 400.

Inconsistent behavior across models reduces system reliability.

badgerbees:
hey, this and issue #844 should be fixed with my PR being merged into dev

viettranx:
Fixed by PR #933 (merged 2026-04-17) `fix(providers): strip Gemma 4 reasoning leak`. Closing as resolved.

PR #1498, opened 2026-08-03
fix(providers): carry tool name on tool results for Gemini

justintruong29:
## Why

Gemini drops `tool_call_id` and pairs `functionCall` with `functionResponse` **by function name**, so its OpenAI-compat shim requires a non-empty `FunctionResponse.name`.

Today that name is only recoverable through a reverse `id -> name` lookup over the assistant `tool_calls` still present in the request window. That lookup fails as soon as history is **pruned, truncated, or tool_calls are collapsed** — and at that point every follow-up iteration returns HTTP 400 `Name cannot be empty`. In practice a long-running Gemini session dies mid-conversation, exactly when the context has grown enough to matter.

## What changed

- Add `Message.ToolName`, set it at every tool-result creation site, and preserve it through context pruning; prefer it when serializing.
- Fall back to the existing reverse index for history persisted before the field existed, so old sessions keep working without a reset.
- When neither source resolves a name, **drop** the unlabelled tool result instead of emitting an empty one. An empty name is a guaranteed 400, and a synthetic name would match no prior `functionCall`. This path is only reachable for legacy history whose assistant `tool_call` has already left the window, so nothing dangles.
- Gated behind the existing Gemini detection, so other OpenAI-compat hosts (Together, Groq, vLLM) that pair by `tool_call_id` are unaffected.

The field is JSON-optional: no session-history migration is needed and older binaries ignore it on rollback.

## Scope

`internal/providers/` (new field + serialization), plus the tool-result creation sites in `internal/agent/`, `internal/pipeline/`, `internal/tools/`.

## Tests

`internal/providers/openai_tool_message_name_test.go` covers name propagation, the legacy-index fallback, and the drop path.

### DIFF
--- a/internal/agent/loop_history_sanitize.go
+++ b/internal/agent/loop_history_sanitize.go
@@ -152,6 +152,7 @@ func sanitizeHistory(msgs []providers.Message) ([]providers.Message, int) {
 						Role:       "tool",
 						Content:    "[Tool result missing — session was compacted]",
 						ToolCallID: tc.ID,
+						ToolName:   tc.Name,
 					})
 					dropped++
 				}
--- a/internal/agent/loop_pipeline_tool_callbacks.go
+++ b/internal/agent/loop_pipeline_tool_callbacks.go
@@ -168,6 +168,7 @@ func (l *Loop) makeExecuteToolRaw(req *RunRequest) func(ctx context.Context, tc 
 			Role:       "tool",
 			Content:    result.ForLLM,
 			ToolCallID: tc.ID,
+			ToolName:   tc.Name,
 			IsError:    result.IsError,
 		}
 		return msg, &toolRawResult{result: result, duration: dur, start: start, spanID: spanID, toolName: registryName, rawName: tc.Name}, nil
--- a/internal/agent/loop_tools.go
+++ b/internal/agent/loop_tools.go
@@ -158,6 +158,7 @@ func (l *Loop) processToolResult(
 		Role:       "tool",
 		Content:    result.ForLLM,
 		ToolCallID: tc.ID,
+		ToolName:   tc.Name,
 		IsError:    result.IsError,
 	}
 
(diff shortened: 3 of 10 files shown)


clark-cant (approved):
## Review: PR #1498 — fix(providers): carry tool name on tool results for Gemini

**Summary:** Fixes a production bug where Gemini (and other Gemini-native models) return HTTP 400 `Name cannot be empty` on follow-up turns after context pruning/truncation. The old reverse `id -> name` lookup over assistant `tool_calls` fails once that call leaves the request window; this PR carries the tool name on the message itself (`Message.ToolName`), sets it at every tool-result creation site, preserves it through pruning, and drops unresolvable legacy results instead of emitting an empty name.

**Risk level:** Low — 10 files, +135/−17, JSON-optional field (no migration, rollback-safe), Gemini-gated serialization change.

**Mandatory gates:**
- Duplicate / prior implementation: ✅ clear — prior work #933 (merged) introduced the reverse-index approach but cannot survive pruning; this PR extends it rather than duplicating. Related closed bug report #843 is the same root cause, already closed as resolved by #933; this PR closes the remaining pruned-history gap.
- Project standards: ✅ docs found — follows AGENTS.md conventions (Go, slog, raw SQL untouched, no ORM)
- Strategic necessity: ✅ clear value — fixes hard failure (HTTP 400) that kills long-running Gemini sessions mid-conversation

**Findings:**
- No critical or important issues.
- `ToolName` set at all creation sites (executor, callbacks, memory flush, preflight blocks, sanitize/compaction, subagent exec) — consistent coverage.
- Serialization precedence is correct: carried `ToolName` → reverse index fallback → drop (never emit empty name for Gemini).
- Drop path is safe: only reachable for legacy history whose assistant tool_call is already out of window; a synthetic name would match no prior `functionCall`.
- Gated to Gemini (`supportsThoughtSignature`), so Together/Groq/vLLM that pair by `tool_call_id` are unaffected — verified by dedicated non-Gemini regression test.
- Tests are strong: name survives missing assistant, carried name beats stale index, unresolvable dropped (Gemini) vs kept (non-Gemini).

**Verdict: Approve** — Clean, well-scoped, well-tested fix for a real production failure.

*Posted by github-maintain at 2026-08-03T17:10:00Z*

PR #1499, opened 2026-08-03
fix(vault): resolve vault_read paths under tenant workspace root

justintruong29:
## Why

`vault_read` is a singleton wired once at boot with the **global** workspace root, then joins `doc.Path` onto it. But `doc.Path` is stored relative to the **tenant** root.

For the master tenant both roots coincide, so the bug is invisible. Every other tenant gets ENOENT, because its files live under `tenants/<slug>/`.

The symptom is misleading rather than obvious: `vault_search` is DB-only and keeps returning the document, so users see "search finds it, read fails" instead of a missing file.

## What changed

- Resolve the tenant-scoped root **per request** via `TenantLayer`. This is a no-op for master and preserves existing single-tenant behaviour.
- Passing the tenant root into `resolvePath` also narrows the boundary check, so it now doubles as cross-tenant isolation.
- Runs originating from channels or cron may not carry the tenant slug in context. Without it, `TenantLayer` falls back to a UUID-named directory that does not match the slug-named one the write path creates. A nil-safe `TenantStore` lookup covers that case only.

## Scope

`internal/tools/vault_read.go`, `cmd/gateway_vault_wiring.go`.

## Tests

`internal/tools/vault_read_test.go` covers master (unchanged behaviour), a sub-tenant read, the slug-missing fallback, and the boundary check rejecting a path outside the tenant root.

### DIFF
--- a/cmd/gateway_vault_wiring.go
+++ b/cmd/gateway_vault_wiring.go
@@ -36,6 +36,10 @@ func wireVault(stores *store.Stores, toolsReg *tools.Registry, workspace string,
 	if stores.Episodic != nil {
 		vaultReadTool.SetEpisodicStore(stores.Episodic)
 	}
+	// Tenant slug fallback for runs whose context carries no slug (channels, cron).
+	if stores.Tenants != nil {
+		vaultReadTool.SetTenantStore(stores.Tenants)
+	}
 	toolsReg.Register(vaultReadTool)
 
 	// Build VaultSearchService: fan-out across vault + episodic + KG.
--- a/internal/tools/vault_read.go
+++ b/internal/tools/vault_read.go
@@ -22,6 +22,7 @@ type VaultReadTool struct {
 	vaultStore    store.VaultStore
 	kgStore       store.KnowledgeGraphStore
 	episodicStore store.EpisodicStore
+	tenantStore   store.TenantStore
 	workspace     string
 }
 
@@ -74,8 +75,14 @@ func (t *VaultReadTool) SetKGStore(kg store.KnowledgeGraphStore) { t.kgStore = k
 // lookup when vault lookup misses. nil is safe.
 func (t *VaultReadTool) SetEpisodicStore(es store.EpisodicStore) { t.episodicStore = es }
 
-// SetWorkspace injects the tenant workspace root (wired at boot).
+// SetWorkspace injects the global workspace root (wired at boot). This is the
+// base for every tenant — the tenant-scoped root is resolved per request in
+// tenantWorkspace, since one tool instance serves all tenants.
 func (t *VaultReadTool) SetWorkspace(ws string) { t.workspace = ws }
+
+// SetTenantStore injects an optional TenantStore used to resolve the tenant
+// slug when the request context does not carry one. nil is safe.
+func (t *VaultReadTool) SetTenantStore(ts store.TenantStore) { t.tenantStore = ts }
 
 func (t *VaultReadTool) Name() string { return "vault_read" }
 
@@ -145,8 +152,8 @@ func (t *VaultReadTool) Execute(ctx context.Context, args map[string]any) *Resul
 		return ErrorResult(fmt.Sprintf("vault_read does not support %s files — use read_image/read_audio/read_video/read_document", ext))
 	}
 
-	// Resolve path under tenant workspace root.
-	fullPath, err := t.resolvePath(doc.Path)
+	// Resolve path under the tenant workspace root.
+	fullPath, err := t.resolvePath(t.tenantWorkspace(ctx, tenantID), doc.Path)
 	if err != nil {
 		return ErrorResult(err.Error())
 	}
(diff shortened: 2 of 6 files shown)


justintruong29:
Pushed a follow-up commit that moves the vault integration fixtures under the tenant workspace root. Flagging it explicitly because it touches existing tests, and the reason is the part worth reviewing.

**What broke.** Four existing tests failed against this branch:

```
TestVaultRead_SharedAtWorkspaceRoot
TestVaultRead_OversizeTruncation
TestVaultRead_OutlinksScopeMatrix
TestVaultNamespaceFix_thuyTienScenario
```

All with the same shape: `open <tmp>/tenants/<uuid>/<file>: no such file or directory`.

**Why.** Each of them seeds a **non-master** tenant via `seedTenantAgent`, calls `tool.SetWorkspace(ws)` with a `t.TempDir()`, and writes the fixture straight to `ws`. That treats the injected workspace as the tenant root. `wireVault` runs once at gateway boot, before any tenant context exists, so in production the injected workspace can only be the **global** root — and `doc.Path` is stored relative to the tenant root. The tests passed only because reads ignored the tenant layer entirely, which is the bug this PR fixes.

**What I did not do.** I considered falling back to the base workspace when no slug resolves, which would have kept the tests untouched. That is wrong: `config.TenantScopedDir` deliberately falls back to an id-named directory precisely so an empty slug cannot resolve upward, and for a sub-tenant the base root is master's area. Falling back there would turn a missing-file bug into a cross-tenant read.

**What changed.** Fixtures now go to `config.TenantWorkspace(base, tenantID, slug)` and the slug travels in context, matching a real run. `tenantSlug` is now shared with `seedTenantAgent`, so a test's on-disk layout cannot drift from what the seed inserts. `TestVaultRead_PersonalCrossAgentDenied` and `TestVaultRead_MediaRejected` were passing either way (they deny before the read), but I moved their fixtures too so every test in the file reflects the same production layout.

Happy to split the test commit into its own PR if you would rather review the two separately. Also glad to be told the intended contract is the opposite — that `doc.Path` is meant to be relative to the global root — in which case the fix belongs on the write path instead and I will redo it that way.

clark-cant (approved):
## Review: PR #1499 — fix(vault): resolve vault_read paths under tenant workspace root

**Summary:** Fixes a multi-tenant bug where `vault_read` returns ENOENT for every non-master tenant. The tool is a singleton wired once at boot with the **global** workspace root, but `doc.Path` is stored relative to the **tenant** root (`<root>/tenants/<slug>/`). Master is invisible to the bug because its roots coincide. The fix resolves the tenant-scoped root per request via `TenantLayer`, with a nil-safe `TenantStore` slug fallback for channel/cron runs that carry no slug in context, and passes the tenant root into `resolvePath` so the boundary check doubles as cross-tenant isolation.

**Risk level:** Low — 2 production files (+38/−7), no API contract changes, master-tenant behaviour preserved as no-op, fallback is nil-safe.

**Mandatory gates:**
- Duplicate / prior implementation: ✅ clear — #948 (initial wiring) and #959 (id-namespace collision, merged) address different bugs; no prior PR/commit resolves tenant-root path resolution.
- Project standards: ✅ clean Go, follows existing `internal/tools/` + `cmd/gateway_*_wiring.go` store-injection patterns.
- Strategic necessity: ✅ clear value — fixes misleading "search finds it, read fails" ENOENT for all non-master tenants.

**Findings:**
- No critical or important issues.
- `tenantWorkspace()` is correct: slug from context first; `TenantStore.GetTenant` fallback fires only when slug is missing **and** the tenant is non-master; master remains a no-op.
- Passing the tenant root (not the global one) into `resolvePath` strengthens the symlink/boundary guard — a free cross-tenant isolation win.
- Tests are strong: 3 new unit tests (non-master resolves under `tenants/<slug>/`, master still uses base root, slug-missing context falls back to TenantStore), plus integration fixtures moved to tenant-scoped dirs with slug-carrying contexts. The `tenantSlug`/`tenantWorkspaceDir`/`tenantCtxSlug` helper extraction keeps on-disk layout and seed data from drifting.
- Suggestion (non-blocking): if `GetTenant` fails in the fallback path the slug stays empty and the subsequent open fails naturally with a path-context error — acceptable, no change needed.

**Verdict: Approve** — Clean, well-scoped fix for a real multi-tenant regression with good coverage.

*Posted by github-maintain at 2026-08-03T18:30:00Z*

clark-cant (approved):
## Review — fix(vault): resolve vault_read paths under tenant workspace root

**Summary**: Fixes a multi-tenant bug where `vault_read` (wired once at boot with the global workspace root) joins `doc.Path` onto the wrong root. Since `doc.Path` is stored relative to the **tenant** root (`tenants/<slug>/`), every non-master tenant gets ENOENT on read while `vault_search` (DB-only) still returns the document — the misleading "search finds it, read fails" symptom. The fix resolves the tenant-scoped root per request via `TenantLayer` and uses the tenant root as the boundary for the path-traversal check, so it doubles as cross-tenant isolation. Channel/cron runs without a slug in context fall back to a nil-safe `TenantStore` lookup.

**Risk level**: Low — 2 production files touched (`cmd/gateway_vault_wiring.go` +6, `internal/tools/vault_read.go` +34/−7), rest is tests.

**Mandatory gates**:
- Duplicate / prior implementation: clear — `TenantWorkspace`/`TenantLayer`/`ResolveWorkspace` already exist and are reused, not reimplemented. Sibling PR #1501 fixes a different `vault_read` concern (UTF-8 sniff boundary).
- Project standards: docs found — matches existing `config.TenantScopedDir` semantics (master no-op, empty-slug → ID-named dir, path-traversal guard).
- Strategic necessity: clear value — repairs broken multi-tenant vault reads and hardens the boundary check; single-tenant/master behavior unchanged.

**Findings**:
- Suggestion: none blocking. `GetTenant` is only consulted when slug is missing and tenant is non-master — cheap, cached-per-request, no hot path concern.

**Verdict**: Approve — no critical or important issues. Merge blocked only by missing CI checks (no checks reported on the branch yet).

*Posted by github-maintain (ck:review-pr flow) at 2026-08-03T18:58:00Z*

PR #1500, opened 2026-08-03
fix(mcp): stop stripping all-caps business identifiers from tool arguments

justintruong29:
## Why

`isAllCapsPlaceholder()` classifies **any** string of >=3 chars made only of `A-Z` and underscore as an LLM placeholder, and `stripEmptyOptionalArgs()` then removes it from the outbound `CallTool` arguments.

Real business identifiers match that exact shape:

- bank codes: `BKASH`, `NAGAD`, `ROCKET`, `VCB`
- currency codes: `BDT`, `VND`, `USD`, `PKR`, `MMK`
- enum values: `SUCCESS`, `PENDING`, `ALL`

The impact is silent and severe: dropping an optional filter turns a filtered query into an unfiltered one. The MCP server still answers 200 with the full result set, so nothing errors — the model reports totals for the entire dataset as if they were the filtered result, with full confidence.

Observed in production: asking "how many BKASH proxies are active" returned 6,610 (all banks) instead of 1,263. Three different bank codes returned byte-identical results, which also drove the model into a retry loop until the loop guard fired.

Diagnosis was slow because the removal is invisible: the outbound log records `args_len` **before** stripping, and `stripEmptyOptionalArgs` logs nothing at all — while `normalizeArgsForSchema` right below it does emit `mcp.tool.args.coerced`. We confirmed the argument never reaches the server by echoing the raw JSON-RPC arguments from `RequestContext[CallToolRequestParams]` server-side.

## What changed

- `isAllCapsPlaceholder`: require multi-word (contains `_`) or membership in a known single-word placeholder list. This keeps the original intent — the doc comment's own examples (`SHOULD_NOT_BE_HERE`, `DO_NOT_SEND`, `NOT_APPLICABLE`) are all multi-word — while letting business codes through.
- `stripEmptyOptionalArgs`: emit `slog.Warn("mcp.tool.args.stripped")` whenever an argument is dropped, so this class of bug is one grep away instead of a multi-hour investigation.

## Scope

`internal/mcp/bridge_tool.go`.

## Tests

`internal/mcp/bridge_tool_test.go` guards both layers against regression.

### DIFF
--- a/internal/mcp/bridge_tool.go
+++ b/internal/mcp/bridge_tool.go
@@ -370,12 +370,18 @@ func (t *BridgeTool) stripEmptyOptionalArgs(args map[string]any) map[string]any 
 		}
 		if s, ok := v.(string); ok {
 			// Strip known placeholder values (e.g. "optional", "null", "http://example.com").
+			// Logged: dropping an argument silently changes query semantics (a filtered
+			// call becomes unfiltered) and is otherwise invisible in traces.
 			if isPlaceholderValue(s) {
+				slog.Warn("mcp.tool.args.stripped",
+					"tool", t.registeredName, "arg", k, "value", s, "reason", "placeholder")
 				continue
 			}
 			// Type-aware empty string: keep for string-typed params (user may want empty),
 			// strip for non-string params (empty string is never valid for number/boolean/UUID).
 			if s == "" && t.propertyType(k) != "string" {
+				slog.Warn("mcp.tool.args.stripped",
+					"tool", t.registeredName, "arg", k, "reason", "empty_string_non_string_param")
 				continue
 			}
 		}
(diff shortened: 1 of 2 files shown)


clark-cant (commented):
## Review: PR #1500 — fix(mcp): stop stripping all-caps business identifiers from tool arguments

**Summary:** Fixes a production bug where `isAllCapsPlaceholder()` classified any all-caps string ≥3 chars as a placeholder, silently dropping real business identifiers (bank codes, currency codes, enum values) from MCP tool arguments. A filtered query silently became unfiltered, and the model reported wrong totals with full confidence.

**Risk level:** Low — targeted 2-file fix with clear scope, no API changes.

**Mandatory gates:**
- Duplicate / prior implementation: ✅ clear — no prior PR/issue/commit addresses this
- Project standards: ✅ clean Go code, follows existing patterns
- Strategic necessity: ✅ clear value — fixes silent data corruption in production

**Findings:**
- Logic change is sound: multi-word all-caps (`_`-separated) → placeholder; single-word must be in known list
- Known-placeholder list is reasonable and well-documented in code comments
- `slog.Warn` additions for stripped args are good observability — makes this class of bug grep-able
- Tests cover both unit (`isAllCapsPlaceholder`) and integration (`stripEmptyOptionalArgs`) levels
- Test names document the production incident for future maintainers

**Verdict: Approve** — Clean, well-tested fix for a real production bug.

*Posted by github-maintain at 2026-08-03T16:28:00Z*

clark-cant (approved):
LGTM. Clean fix for a real production bug — silent data corruption from all-caps business identifiers being stripped as placeholders. Logic change is sound, tests cover both unit and integration, slog warnings add good observability. No issues found.

PR #1503, opened 2026-08-04
fix(slack): gate require_mention on the event's own text, not injected thread context

### LINKED ISSUE #1502: Slack: require_mention is bypassed for every reply in a thread whose root message mentions the bot
## Bug

With a Slack channel configured as `require_mention: true` and `thread_ttl: 0` ("always require @mention"), the bot still auto-replies to **every** reply in a thread whose **root message** @mentions the bot — including human-to-human messages that never mention it.

## Root cause

In `internal/channels/slack/handlers.go` (`handleMessage`), the order of operations is:

1. For thread replies, `fetchThreadParentContext` fetches the thread root and prepends its **raw text** — mention tags included — to `content`:
   `content = replyCtx + "\n\n" + content`
2. The mention gate then runs **on the combined string**:
   `mentioned := c.isBotMentioned(content)`

Since `isBotMentioned` is a plain `strings.Contains(text, "<@"+botUserID+">")`, the parent's mention tag inside the quoted `[Replying to …]` block satisfies the gate for every reply in that thread, forever — regardless of `require_mention` and `thread_ttl`.

## Reproduction

1. Slack channel instance with `require_mention: true`, `thread_ttl: 0`.
2. User A posts a channel message mentioning the bot: `@bot @UserB please help with X` (starts a thread).
3. User B replies **in the thread without mentioning the bot**: `@UserA do you have more info?`

**Expected:** the bot stays silent (no mention in the reply, thread auto-reply disabled).
**Actual:** the bot replies. Confirmed in production logs: the reply event reaches the agent (`inbound: scheduling message`) with no mention in its text.

## Fix

Evaluate the mention gate on the **event's own text** (`ev.Text`) before any context injection. PR incoming that extracts the gate into a `shouldRespondInGroup(eventText, channelID, threadTS)` helper (thread-participation TTL behavior preserved) with table tests pinning the regression.

alex-coolfy:
Fixes #1502

## Problem

`handleMessage` prepends the thread parent's raw text (mention tags included) to `content` via `fetchThreadParentContext`, then evaluates `isBotMentioned(content)`. In any thread whose root message @mentions the bot, every reply inherits the root's mention and passes the gate — the bot auto-replies to human-to-human thread messages even with `require_mention: true` and `thread_ttl: 0`.

## Fix

- Extract the gate into `shouldRespondInGroup(eventText, channelID, threadTS)` in `handlers_mention.go`, called with `ev.Text` **before** any context injection.
- Thread participation TTL behavior is preserved unchanged: same check order, same auto-reply window, same stale-entry eviction, same debug log.
- `content` still carries the quoted parent context downstream for the agent — only the *gating source* changes.

The `message_changed` path already gates on `ev.Message.Text` and promotes it to `ev.Text` before this point, so it is unaffected. `file_share` messages now gate on the user's caption (`ev.Text`) rather than extracted document text, which is consistent with the same principle.

## Tests

- `TestShouldRespondInGroup` table tests: mention in event text (channel + thread), no mention at channel level, **the regression case** (thread reply without its own mention, `thread_ttl=0`, participation entry present → must NOT respond), fresh participation within TTL, stale participation.
- `TestShouldRespondInGroupEvictsStaleParticipation` pins stale-entry eviction.
- `go build ./...`, `go vet`, `gofmt` and the full `internal/channels/slack` suite pass.


alex-coolfy:
FYI: we are now running this patch in production (built as upstream v3.14.0 + this commit cherry-picked). It deterministically fixed the reported behavior — thread replies without a mention no longer trigger the bot in channels with require_mention: true, while explicit mentions and the thread-participation TTL window keep working. Happy to adjust anything if you'd like changes.

clark-cant (changes requested):
## Review — scope blockernnThe described Slack mention-gate fix is valuable and the production note is useful, but this PR cannot be reviewed or merged as submitted: it changes **1,330 files** (+138,935/-7,005) and is currently **DIRTY**. That is far beyond the stated single-handler fix and makes correctness, regression risk, and provenance impossible to assess safely.nnPlease rebase onto the intended base branch and split this into a minimal PR containing only the Slack gate change plus focused regression tests. Once the diff is narrow and mergeable, it can receive a substantive review.nn**Mandatory gates**n- Duplicate/prior implementation: no merged replacement identified for this scoped fix.n- Project standards: repository conventions cannot be validated across this unrelated oversized diff.n- Strategic necessity: clear for the Slack bug, not justified for the unrelated bulk changes.nn**Verdict: Request changes** — narrow scope and resolve conflicts before review.nn*Posted by /ck:review-pr at 2026-08-16T05:15:51Z*

Issue #1414, opened 2026-07-09
Cannot add Agent (with Ollama provider)


imjackvo:
Provider & Model verified successfully but failed to summon (Actually there was nothing happening once I clicked on Summon button.

<img width="2544" height="786" alt="Image" src="https://github.com/user-attachments/assets/693a5c97-3f70-434b-b562-ac7a2d42672f" />

<img width="2566" height="1668" alt="Image" src="https://github.com/user-attachments/assets/e1839049-6693-4d3b-a96d-e9d75bba6e0f" />

clark-cant:
## Maintainer Triage — github-maintain

Thanks for reporting, @imjackvo. This looks like a UI-level silent failure during agent creation with Ollama provider.

**Issue type:** Bug — Agent creation UI unresponsive
**Severity:** P2-medium — Core workflow broken, but provider/model verification works
**Area:** Desktop UI, agent creation flow

### Initial assessment

From the screenshots:
- Provider `qwen3-coder:30b` (Ollama local) is configured and verified successfully
- Agent creation modal is fully populated with valid data
- Clicking "Summon" button produces no visible response (no error, no toast, no agent creation)

### Likely causes

1. **Frontend silent failure**: The `createAgent` call may be failing but not propagating the error to the UI. The code shows error handling with `toast.error()`, but if the error is undefined or malformed, the toast may not display.

2. **Backend validation failure**: The `/v1/agents` POST endpoint may be rejecting the request (e.g., duplicate agent key "dev", missing required field, provider not registered) but returning a non-standard error format that the frontend doesn't handle.

3. **Ollama provider registration**: Even though the provider is "verified", it may not be fully registered in the agent creation path. The provider verification and agent creation may use different code paths.

4. **Agent limit reached**: The code shows `MAX_AGENTS_LITE = 5`. If you're at the limit, creation may silently fail.

### Diagnostic steps

Please check:

1. **Browser console**: Open DevTools (F12) → Console tab. Look for any JavaScript errors when clicking Summon.

2. **Network tab**: Check if the POST `/v1/agents` request is being sent. What's the response status and body?

3. **GoClaw logs**: Check gateway logs for agent creation attempts:
   ```
   docker compose logs goclaw --tail 100 | grep -i agent
   ```

4. **Try web UI**: Does the same issue occur in the web UI or only in desktop?

5. **Agent key uniqueness**: Try changing the "Agent Key" from "dev" to something unique (e.g., "dev-test-1").

[...]

PR #1062, opened 2026-04-28
fix(cli): use openai_compat instead of openai-compat in provider setup

wucm667:
Fixes #1046

## Problem

The CLI setup wizard (`goclaw setup`) and `providers add` command display `openai-compat` as the provider type, but the HTTP API (`/v1/providers`) and database validation expect `openai_compat` (underscore).

This mismatch causes provider creation to fail during the setup process.

## Solution

Change `openai-compat` to `openai_compat` in the CLI provider type definitions to match the internal constant `store.ProviderOpenAICompat`.

## Changes

- `cmd/setup_provider.go`: Updated provider type mapping and check
- `cmd/providers_cmd.go`: Updated provider type mapping and check

PR #1504, opened 2026-08-05
Allow adding Ollama local provider without API key

bclermont:
## Summary
- Ollama local deployments don't require API keys by default
- Provider add command was mandating an API key for all types, blocking local Ollama setup
- Now Ollama (local) is a distinct option that skips the API key prompt
- Provides localhost:11434 as default base URL for convenience

## Changes
- Added "Ollama (local)" as separate provider type option
- Skip API key prompt for Ollama local
- Added defaultBaseURL case for Ollama
- Only include api_key in request body if non-empty
- Removed unused debug.zip artifact

## Test plan
- Manual test: `goclaw providers add` → select Ollama (local) → no API key prompt
- Verify localhost:11434 is default URL
- Create provider and verify it works

🤖 Generated with Claude Code


clark-cant:
## Maintainer Review — github-maintain bot

**Summary**: Adds Ollama (local) as a distinct provider type in the CLI, skipping the API key prompt and defaulting to localhost:11434. Also conditionally includes api_key in request body only when non-empty.

**Risk level**: Low — small, focused CLI change (+29/-9).

**Mandatory gates**:
- Duplicate / prior implementation: clear — Ollama provider exists but CLI always required an API key. This fills a real gap (see #1414).
- Project standards: clear — follows existing CLI patterns.
- Strategic necessity: clear value — Ollama is a popular local LLM; removing the fake API key friction improves onboarding.

**Findings**:
- **Suggestion**: The diff includes a rename of `openai-compat` → `openai_compat` and adds `atlascloud` — both already present on main (merged via #1062 and earlier). A rebase would clean these up, but they're harmless no-ops on merge.
- **Note**: CI `go` job failed on `go vet` — but the failure is in `internal/http/builtin_tools_tenant_settings_test.go` and `tenant_scope_hotfix_test.go` (missing `Upsert` method on test mock). This is a **pre-existing main branch issue**, not caused by this PR.

**Verdict**: Comment — code is correct and addresses a real need. CI failure is pre-existing on main. Recommend rebase to clean stale diff hunks, then merge.

Issue #688, opened 2026-04-04
Feature: Cryptographically Signed Tool Call Receipts via SpanExporter


tomjwxf:
## Problem

GoClaw's tracing system (`internal/tracing/collector.go`) already produces spans for every tool call, with OTLP export to Jaeger/Grafana Tempo/Datadog. But these spans are observability data -- they tell you what happened, but they can't prove it. If the PostgreSQL store or the OTLP receiver is compromised, the traces can be silently modified.

For multi-tenant deployments where GoClaw manages agents across different organizations, the audit question escalates: *"Can you prove, to a third party, that Agent X executed Tool Y under Policy Z at time T?"*

Related: #611 (langsmith-go tracing) shows demand for richer tracing integrations.

## Proposal

Implement a `ReceiptSpanExporter` that satisfies GoClaw's existing `SpanExporter` interface and adds Ed25519 receipt signing to every span:

```go
// Implements the existing SpanExporter interface -- zero core changes needed
type ReceiptExporter struct {
    signer   ed25519.PrivateKey
    policyID string  // SHA-256 digest of active Cedar policy set
    store    ReceiptStore
}

func (r *ReceiptExporter) ExportSpan(span *Span) error {
    receipt := Receipt{
        Tool:      span.ToolName,
        Decision:  span.Decision,  // allow/deny from PolicyEngine
        AgentID:   span.AgentID,
        TenantID:  span.TenantID,
        Policy:    r.policyID,
        Timestamp: span.StartTime,
    }
    receipt.Signature = ed25519.Sign(r.signer, receipt.Canonical())
    return r.store.Append(receipt)
}
```

### Why SpanExporter (not tool executor middleware)

1. **Zero core changes** -- implements the existing interface, plugs in via config
2. **Already has the data** -- spans contain tool name, input preview, timing, cost, tenant
3. **Parallel to OTLP** -- receipts and OTLP traces coexist, serving different needs (verifiability vs. observability)
4. **License-clean** -- the exporter lives in a separate package (MIT), GoClaw stays CC BY-NC

### What receipts add on top of tracing

| | OTLP Traces (current) | + Receipt Signing |
|---|---|---|
| **Proves** | Something happened | Something happened **and can't be denied** |
| **Trust model** | Trust the trace collector | **Verify the math** (Ed25519) |
| **Tamper detection** | None | Signature breaks on modification |
| **Multi-tenant** | Tenant isolation via labels | + Tenant-scoped receipt chains |
[...]

PR #1505, opened 2026-08-06
feat(tracing): sign exported spans as verifiable receipts

tomjwxf:
Implements #688. Thanks for the triage, and for confirming the `SpanExporter` seam was the right place before I built anything.

Adds a `SpanExporter` that signs exported spans as Ed25519 receipts, chains them, and anchors the sequence with a signed checkpoint, so a third party can verify what a run did without trusting the collector, the database, or the operator.

Standard library crypto only. No new dependencies, and no existing file is modified: it attaches through `SetExporter` exactly the way OTLP does.

## The thing I want reviewed hardest

`Collector.EmitSpan` drops spans when its buffer is full. That is the right call for observability, since tracing must never block a run, and it means **an exporter can only sign what reaches it.**

So `Coverage` counts spans this exporter received, never spans the runtime executed, and every checkpoint carries a note saying so. A signed receipt set is a floor on what happened, not a complete record. I would rather that limitation be stated in the artifact than discovered by someone relying on it.

Quantifying the gap would need a drop counter in the collector. I deliberately did not add one, since the triage valued the zero-core-changes property, but it is a small follow-up if you want the number.

## The checkpoint is not optional, and here is why

A hash chain detects modification, reordering, and deletion from the middle. It cannot detect truncation. Remove the newest receipts and every remaining link still verifies clean, which is exactly the attack worth running, since the inconvenient span is usually the most recent one.

The checkpoint commits to a receipt count and a terminal hash. `VerifyReceipts` returns `Valid: false` when no checkpoint is supplied, because without one it can only say "nothing I was given has been altered", never "nothing is missing". An empty bundle fails too, unless the anchor says the run genuinely produced nothing.

One honest limit, stated in the package docs: **a checkpoint stored beside the receipts it anchors is deletable by whoever deletes the receipts.** To be worth anything the anchor has to reach somewhere the operator cannot quietly rewrite, such as a build log, an object store with retention, or the commit the run produced. `Sink` keeps the receipt and checkpoint destinations separate so that publishing step is easy to add. I did not presume which one GoClaw would want.

## Other design decisions worth arguing with

**Content is committed by digest, not copied.** There is a `redaction.go` in this package, so previews are treated as sensitive. Duplicating that content into a second, separately stored artifact would widen the blast radius of a leak while adding nothing a digest does not already prove. A test asserts a preview value never appears in the signed bytes.

**Ordering comes from an exporter sequence, not a clock.** Ordering evidence by a timestamp invites an older stamp recorded later to make honest history look altered.

**Revocation is not retroactive.** A receipt signed while the key was trusted stays valid, because the alternative silently voids honest history every time an operator rotates. `RevokedAt` exists for the compromise case, where everything from that moment on should fail.

**Signature and chain failures are reported separately.** They call for different responses, and a deleted receipt leaves every remaining signature genuine, so collapsing them into one boolean would hide the distinction that makes the chain worth having.

**A failed sink write does not advance the chain head.** Otherwise a transient outage produces a permanent break indistinguishable from tampering. My own test caught this when I got it wrong the first time.

**`ExportSpans` returns nothing by interface contract**, so every failure is logged and counted rather than propagated. A signing problem must never take down a run that was otherwise fine.

## What a verified receipt establishes

> This span, with exactly these fields, was exported by the holder of this key at this time, and the verified sequence has not been altered since.

## What it does not establish

Returned in every verification report rather than left to a reader to infer:

* Not that these are all the spans the runtime produced.
* Not that a tool call was correct, authorised, or safe.
* Not that receipts signed by a compromised key were truthful.
* Without a checkpoint, not that receipts were not removed from the end.

## Testing

26 tests, most of them adversarial: payload tampering, forged signatures under a claimed `kid`, cross domain signature replay, key validity windows, revocation ordering in both directions, deletion from the middle, reordering, truncation from the end, empty bundles, keyring entries lying about their own key, sink failure, spans arriving after shutdown, and concurrent export.

`go test`, `go vet` and `go test -race` all clean. `gofmt` clean. The rest of `./internal/tracing/...` still passes.

## Disclosure

I opened #688, I maintain protect-mcp, and I author the Acta signed receipts Internet-Draft. This implementation depends on none of them and references none of them anywhere in the code. It is standard library crypto, written to the seam you already had.

## Happy to adjust

Config wiring is not included, since I did not want to guess where GoClaw would want the toggle or how it would supply key material. Say where and I will add it. Equally happy to rename the package, move it, split the PR, or add the collector drop counter as a follow-up.

### DIFF
--- a/internal/tracing/receiptexport/README.md
+++ b/internal/tracing/receiptexport/README.md
@@ -1,0 +1,124 @@
+# receiptexport
+
+Signs exported spans as Ed25519 receipts so a third party can verify what a run
+did without trusting the collector, the database, or the operator.
+
+Implements `tracing.SpanExporter`, so it attaches the same way OTLP does and
+changes nothing in the collector.
+
+## Why this is not just a trace
+
+A trace says what the runtime emitted. A receipt says what a named key attested
+to, in what order, and whether that record has been altered since. The
+difference matters the moment a run crosses an organisational boundary, or
+someone has to explain a quiet failure after the fact.
+
+## Usage
+
+```go
+key, err := receiptexport.GenerateSigningKey()
+if err != nil {
+    return err
+}
+
+sink, err := receiptexport.NewFileSink(
+    "/var/lib/goclaw/receipts.jsonl",
+    "/var/lib/goclaw/checkpoint.json",
+)
+if err != nil {
+    return err
+}
+
+exporter, err := receiptexport.New(receiptexport.Config{
+    Key:             key,
+    Sink:            sink,
+    SpanTypes:       []string{"tool_call"}, // omit to sign everything
+    CheckpointEvery: 100,
+})
+if err != nil {
+    return err
+}
+
+collector.SetExporter(exporter)
+```
+
+Publish only the public half for verifiers:
+
+```go
+keyring := receiptexport.NewKeyring(key.PublicEntry())
+```
+
+Verify with nothing but the evidence and that keyring:
+
+```go
+receipts, checkpoint, err := receiptexport.LoadBundle(receiptPath, checkpointPath)
+report, err := receiptexport.VerifyReceipts(receipts, checkpoint, keyring)
+```
+
+## The checkpoint is not optional
+
+A hash chain detects modification, reordering, and deletion from the middle. It
+cannot detect truncation: remove the newest receipts and every remaining link
+still verifies clean.
+
+The checkpoint closes that by committing to a receipt count and a terminal
+hash. `VerifyReceipts` reports `Valid: false` when no checkpoint is supplied,
+because without one it can only say "nothing I was given has been altered",
+never "nothing is missing".
+
+**A checkpoint stored beside the receipts it anchors is deletable by whoever
+deletes the receipts.** To be worth anything it has to reach somewhere the
+operator cannot quietly rewrite: a build log, an object store with retention, or
+the commit the run produced. `Sink` keeps receipt and checkpoint destinations
+separate so that publishing step is easy to add.
+
+## Coverage, and what it does not claim
+
+`Collector.EmitSpan` drops spans when its buffer is full. That is correct for
+observability, since tracing must never block a run, and it means an exporter
+can only sign what it receives.
+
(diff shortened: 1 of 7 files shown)


clark-cant (approved):
## Review: feat(tracing): sign exported spans as verifiable receipts

**Verdict: Approve** ✅

### Summary
Clean, well-scoped feature PR implementing #688. Adds a `SpanExporter` that signs exported spans as Ed25519 receipts with hash chaining and checkpoint anchoring — allowing third-party verification without trusting the collector, database, or operator.

### Mandatory Gates
- **Duplicate/prior implementation**: Clear — no existing implementation found; issue #688 was open and awaiting this work
- **Project standards**: Compliant — standard library crypto only, no new dependencies, attaches through existing `SetExporter` seam, zero modifications to existing files
- **Strategic necessity**: Clear value — verifiable audit trail for agent actions is a meaningful trust/auditability feature

### Code Quality
- **Architecture**: Excellent separation of concerns — `Sink` interface decouples storage from signing, `FileSink` is the simplest implementation, checkpoint/receipt paths are separate by design
- **Correctness**: Chain head only advances after successful write (prevents transient failures from looking like tampering), mutex-protected concurrent access, proper shutdown semantics
- **Security**: Ed25519 with proper signing domain separation, content committed by digest (not copied), redaction-aware design
- **Testing**: 26 tests covering adversarial scenarios (tampering, forgery, truncation, reordering, sink failure, concurrent export, post-shutdown arrivals)
- **Documentation**: README clearly states what receipts establish vs. what they don't, coverage limitations are honest

### Minor Notes
1. **Base branch**: PR targets `main` — confirm this is intentional. Most feature PRs target `dev`. Since this is a self-contained new package with no existing code modifications, it may be fine, but worth verifying against branch policy.
2. **Config wiring**: Author explicitly deferred config wiring (toggle, key material supply) — this is the right call to keep the PR scoped. Follow-up issue for config integration would be appropriate.
3. **Collector drop counter**: Author noted this as optional follow-up — worth tracking if operators need to quantify the gap between runtime spans and signed spans.

### Risk Level
**Low** — additive feature, no existing behavior modified, clean package boundary.

---

PR #1514, opened 2026-08-15
fix(channels): archive group messages before the pending buffer drops them

minhdang03:
`channel_pending_messages` is a buffer, not an archive. Rows leave it on two paths, both plain deletes:

- the bot being mentioned hands the whole buffer to the agent, then `Clear()` wipes the key;
- LLM compaction replaces old rows with a summary.

For group capture that buffer holds the only stored copy of the raw text, so a single mention or one compaction pass permanently destroyed every message nothing had read yet. Observed in production: a compaction run erased ~4.5 days of a Zalo group's history, and a later mention erased the rest along with the summary that replaced it.

## Change

- New table `channel_message_archive` (migration 97; SQLite schema 60).
- `DeleteByKey`, `Compact` and `DeleteStale` copy rows into the archive **inside the same transaction** as the delete, tagged with `archive_reason` (`consumed`, `compacted`, `stale`).
- Archived rows keep their original `id`, so a replayed delete is a no-op.
- New `ListArchivedByKey(channel, key, since, limit)` for consumers that need full history; tenant-scoped like every other store read.

Wiring lives in the store layer, so every channel (Telegram, Discord, Slack, Feishu, WhatsApp, Zalo) and the HTTP compact endpoint are covered with no call-site changes.

## Tests

- 4 SQLite unit tests (`internal/store/sqlitestore/pending_message_archive_test.go`)
- 4 PG integration tests (`tests/integration/pending_message_archive_test.go`), run against a clean pgvector instance migrating 1 → 97

Covers clear, compaction, idempotency across compact-then-clear, the `since` filter, and tenant scoping.

`go build` and `go vet` clean under both the default and `sqliteonly` tags. The 6 pre-existing Ollama failures in `internal/http` reproduce on a clean `origin/dev` checkout and are unrelated.

### DIFF
--- a/docs/06-store-data-model.md
+++ b/docs/06-store-data-model.md
@@ -419,12 +419,26 @@ Offline message queue for group chats. Buffers messages when the bot is not acti
 |--------|---------|
 | `AppendBatch(msgs)` | Insert multiple messages in one query |
 | `ListByKey(channelName, historyKey)` | Retrieve buffered messages for a group |
-| `DeleteByKey(channelName, historyKey)` | Clear messages after processing |
-| `Compact(deleteIDs, summary)` | Atomically delete old messages + insert summary |
-| `DeleteStale(olderThan)` | Prune messages older than duration |
+| `DeleteByKey(channelName, historyKey)` | Clear messages after processing (archives first) |
+| `Compact(deleteIDs, summary)` | Atomically archive + delete old messages, insert summary |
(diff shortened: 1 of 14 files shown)


clark-cant (approved):
Approved: archive-before-delete is transactionally implemented for PostgreSQL and SQLite, tenant-scoped reads are covered, migration/schema parity is present, duplicate gate is clear, and all required CI checks pass. No critical or important findings.

PR #1515, opened 2026-08-15
fix(pipeline): stop delivering the same generated image twice

minhdang03:
## Symptom

Every image-edit reply in a Zalo DM arrived as **two identical attachments**. Confirmed on disk — same bytes, two files:

```
md5 f0d73143f6bc4c48dd68497dd154a898  generated/2026-08-15/ca-nha-mac-ao-spiderman-sinh-nhat_20260815-122109_291214.png
md5 f0d73143f6bc4c48dd68497dd154a898  generated/2026-08-14/ca-nha-mac-ao-spiderman-sinh-nhat.jpg
```

Separate inodes, 1 link each, 2365147 bytes both. The last four image replies in that thread all carried `media_refs` of length 2 with the same shape: the file `create_image` actually wrote (`slug_timestamp_rand.png`) plus the path the agent passed to `send_file`.

## Cause

`FinalizeStage.processMedia` deduplicated by path:

```go
seen := make(map[string]bool, len(media))
for _, m := range media {
    if !seen[m.Path] { ... }
}
```

Two paths, identical bytes — both survive. Separately, `processMedia` ran at step 3 while persisted assistant images were appended to `MediaResults` at step 3d, so that producer never passed through dedupe at all.

## Change

- Fold persisted assistant images into `MediaResults` **before** dedupe.
- Deduplicate once over the union: path first, then SHA-256 for local files.
- Build the assistant message from the deduplicated set, so session history and outbound delivery carry the same attachments.
- Media that cannot be read (remote, already cleaned up) is kept as-is.

## Tests

Four cases in `internal/pipeline/finalize_stage_media_dedupe_test.go`: identical image under two paths ships once, distinct images both ship, a repeated path collapses, unreadable media survives. The first reproduces production exactly — without the fix it reports `MediaResults = 2 entries [.../poster.jpg .../poster_20260815-122109_291214.png]`.

Existing `TestFinalizeStage_DeduplicatesMediaByPath` and `TestFinalizeStage_DeduplicatesMedia_WithRealFile` still pass. `go build`/`go vet` clean under both the default and `sqliteonly` tags.

### DIFF
--- a/internal/pipeline/finalize_stage.go
+++ b/internal/pipeline/finalize_stage.go
@@ -2,6 +2,9 @@ package pipeline
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
Summary: Consolidates generated assistant-image and tool media before a single path/SHA-256 dedupe pass, keeping persisted session media aligned with outbound delivery.nnRisk level: Low.nnMandatory gates:n- Duplicate / prior implementation: clear; adjacent prior media fixes do not implement this cross-path content dedupe.n- Project standards: complied with existing pipeline patterns and focused test coverage.n- Strategic necessity: clear reliability and UX value; prevents duplicate attachment delivery.nnFindings: none. CI is green (go, web, release-versioning).nnVerdict: Approvenn*Posted by /ck:review-pr at 2026-08-15T14:45:00Z*

PR #1522, opened 2026-08-23
Update MiniMax music generation handling

octo-patch:
Reason: Add the requested music generation support.

## Summary

- Update MiniMax music generation handling.
- Preserve provider-specific request and response behavior.

## Validation

- GOTOOLCHAIN=local GOPROXY=https://goproxy.cn,direct go test ./internal/audio/minimax
- GOTOOLCHAIN=local GOPROXY=https://goproxy.cn,direct go test ./internal/audio ./internal/http

### DIFF
--- a/internal/audio/minimax/music.go
+++ b/internal/audio/minimax/music.go
@@ -3,6 +3,7 @@ package minimax
 import (
 	"bytes"
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


clark-cant (changes requested):
## Review summary

This focused MiniMax update moves the defaults to music-3.0 and adds response decoding coverage. The scope is small, but the asynchronous response path is not complete.

### Important — handle in-progress responses instead of treating them as terminal failures

`data.status == 1` is now returned as `MiniMax music generation is still in progress`, but the provider does not retain a task/job identifier or poll the provider before failing. For an asynchronous generation request this makes a normal accepted/in-progress response an unavoidable user-visible failure; the new test locks that failure in as expected behavior.

Please implement the documented completion flow (capture the task identifier and poll/retry until success/cancellation/context deadline), or confirm with an API fixture that this endpoint is synchronous and cannot return status 1 for a valid request. Add coverage for the eventual-success path.

**Verdict: Request changes** — the async lifecycle must be handled before this can reliably serve music generation.

*Posted by github-maintain at 2026-08-24T08:24:34Z*

PR #1528, opened 2026-08-25
fix(teams): dispatch team tasks created by detached child runs

yatul:
Fixes #1527.

### Problem

Team tasks are not dispatched inline. `executeCreate` hands them to the turn's `PendingTeamDispatch`, and the post-turn drain assigns and dispatches them once the turn ends. That same tracker also owns the per-`(team, chat)` create lock taken by `team_tasks` `list`/`search`.

`executeAsyncMode` (delegate) and `SpawnWithReceipt` (spawn) detach with `context.WithoutCancel(ctx)`, which preserves context *values* — including the caller's tracker — while the run continues after the caller's turn has already ended and drained. So a team lead reached through async `delegate` adds every task it creates to a tracker nobody will drain again, and takes a create lock nobody will release.

Observed on `v3.15.0-beta.196`, in two independent incidents. The caller's turn always closed before the lead created anything:

```
15:11:34  v3.run.completed agent=foxy          <- caller turn ends, drain fires (empty)
15:12:05  tool call agent=brain tool=team_tasks   (create -> coder, pending)
15:12:35  tool call agent=brain tool=team_tasks   (create -> reviewer, blocked)
15:12:42  v3.run.completed agent=brain
```

The tasks then sat `pending` indefinitely. `task_ticker` does not cover this — it recovers `in_progress` tasks with expired locks, marks pending tasks stale after 2h and asks the lead to retry by hand, and fixes orphaned blocked tasks, but it never dispatches a pending task. (The "ticker will retry" comment in `dispatchTaskToAgent` does not match that behaviour.) Separately, the leaked mutex made the next `list`/`search` for the same `(team, chat)` hang for the rest of the process lifetime.

### Change

Give each detached child run its own tracker through the existing `InjectTeamDispatch` helper, drained when that run ends.

The synchronous paths are deliberately untouched: `executeSyncMode` and `RunSync` execute inside the caller's turn, where the caller's tracker is the correct owner.

`postTurn` is wired into `DelegateTool` and `SubagentManager` the same way it already is for `chatMethods` / `server` / `wakeH`.

### Test

`TestAsyncSpawnDispatchesTeamTasksCreatedAfterParentTurnEnded` reproduces the ordering exactly: the parent's tracker is drained before the detached child starts, the child then registers a task the way `executeCreate` does, and the test asserts the task reaches the `PostTurnProcessor`, that the child got its own tracker, and that nothing leaked into the parent's.

[...]

Issue #1527, opened 2026-08-25
Team tasks created by a delegated lead are never dispatched; team_tasks list then deadlocks


yatul:
### Summary

When a team lead agent is reached through `delegate` (async, the default), every team task it creates stays `pending` forever and is never dispatched, so the team never starts work. As a side effect the per-`(team, chat)` create lock is leaked, and the next `team_tasks` `list`/`search` for that pair blocks for the remaining lifetime of the process.

Both symptoms have the same cause: a detached child run inherits the caller's `PendingTeamDispatch`, but it executes *after* the caller's turn has already drained it.

### Reproduction

1. Agent A (personal assistant, `foxy`) is the one the user talks to. Agent B (`brain`) is the lead of a team with members `coder` / `researcher` / `reviewer`.
2. Ask A for something that A routes to B via `delegate` (default `mode: "async"`).
3. B plans, calls `team_tasks(action="create", assignee="coder", ...)`, replies "pipeline started".
4. Nothing happens. No `team_tasks.dispatch: sent task to agent`, no member session, task stays `pending`.
5. Later, `team_tasks(action="list")` in the same chat never returns.

### Evidence

Two independent incidents on `v3.15.0-beta.196`, one over Telegram and one over the web dashboard. In both the lead created its tasks *after* the caller's turn had finished:

```
15:11:34  v3.run.completed agent=foxy          <- caller turn ends, drain fires (empty)
15:11:44  tool call agent=brain tool=team_tasks   (search)
15:12:05  tool call agent=brain tool=team_tasks   (create -> coder, pending)
15:12:35  tool call agent=brain tool=team_tasks   (create -> reviewer, blocked)
15:12:42  v3.run.completed agent=brain
```

The other incident, same shape: caller finished `15:38:53`, creates landed `15:40:36`, `15:41:32`, `15:42:02`, `15:42:16`.

Trace confirms the tasks were created correctly, just never dispatched:

```
Task created: ... (task_number=1, status=pending,  assignee=Coder)
Task created: ... (task_number=2, status=blocked,  assignee=Reviewer)
```

Two hours later the ticker did its thing, and only then did the work start — via a manual retry by the lead, not via dispatch:

```
17:45:06  task_ticker: marked stale count=2
17:45:20  tool call agent=brain tool=team_tasks    (retry)
17:45:20  team_tasks.dispatch: sent task to agent agent_key=researcher
17:45:20  team_tasks.dispatch: sent task to agent agent_key=coder
[...]

PR #1530, opened 2026-08-25
fix(teams): scope team tasks to the delegation origin, not the delivery channel

### LINKED ISSUE #1529: Team task results never reach the caller when the lead is reached via delegate: notifications routed to the internal "delegate" channel are dropped
Sibling of #1527 — same underlying assumption, that a delegation context is a routable channel. #1527 is about tasks never leaving `pending`; this one is about tasks that *do* run, whose results never make it back. The two do not overlap in code and need separate fixes.

### Summary

A team task created by a lead that was reached through `delegate` records `"delegate"` as its routing channel. `"delegate"` is an internal delivery channel with no registered channel handler, so every outbound message the lead later produces about that task — completion summaries, blocker escalations, stale notifications — is dropped by the outbound dispatcher and the caller never learns anything.

The delegatee's *first* answer still arrives, because it travels back as the delegation result rather than through a channel. Everything the lead says after the delegation has closed is lost.

### Reproduction

1. Agent A (personal assistant) delegates to agent B, the lead of a team.
2. B creates team tasks and answers "started" — A receives this.
3. Members run, complete, or get blocked. B is woken by the teammate announce and composes a real answer.
4. That answer is dropped: `unknown channel for outbound message channel=delegate`. A is never told anything and keeps waiting.

### Evidence

One session, 16 dropped outbound messages on `channel=delegate` (plus 2 on `channel=http`):

```
16:11:39  INFO teammate announce: batch processed  session=delegate:019f4566:brain:e7ee9913-...
16:11:39  WARN unknown channel for outbound message channel=delegate
```

The lead had produced a genuinely useful answer that nobody ever saw — from the trace of that run:

> Reviewer is blocked: tetris.html does not exist, the coder's generation failed. Restarted generation (running), re-pointed the review at it — it will start on its own once the file appears.

Routing of the escalation itself shows the same pseudo-channel:

```
inbound: scheduling message (main lane) channel=delegate chat_id=system peer_kind=direct
  agent=brain session=agent:brain:delegate:direct:system user_id=system
```

With no push path left, the caller's only option is to pull the result with `delegate(action="get", delegation_id=...)`. That is a single point of failure and it is not robust: in this session the caller's model emitted a malformed tool call five times in a row (the whole argument list collapsed into `action`, and the UUID truncated to 27 of 36 characters), hit the loop guard, and gave up. That part is a model-quality problem, not a gateway bug — but it is only reachable because the push path is gone.

### Root cause

`internal/tools/team_tasks_create.go`:

```go
Channel: ToolChannelFromCtx(ctx),
ChatID:  chatID,                    // chatID := ToolChatIDFromCtx(ctx)
```

Inside a delegated run those return the delegation's internal delivery channel, set in `cmd/gateway_delegate.go`:

```go
// buildAgentLinkRunRequest preserves the origin's authorization-bearing
// identity while keeping delegation on its internal delivery channel.
Channel:          "delegate",
...
WorkspaceChannel: req.Channel,   // the real origin channel, deliberately preserved
WorkspaceChatID:  req.ChatID,
```

So the origin is not lost — it is carried explicitly, and `internal/tools/context_keys.go` already exposes it under a section literally named "Workspace scope propagation (delegation origin)":

```go
func WorkspaceChannelFromCtx(ctx context.Context) string
func WorkspaceChatIDFromCtx(ctx context.Context) string
```

`team_tasks_create.go` simply does not consult it, so the task is stamped with the delivery channel instead of the origin, and `internal/channels/dispatch.go` drops everything addressed to it:

```go
channel, exists := m.channels[msg.Channel]
if !exists {
    slog.Warn("unknown channel for outbound message", "channel", msg.Channel)
    return
}
```

Note that `team_tasks_create.go` already stores the origin *session* in metadata for exactly this class of problem (`TaskMetaOriginSession`, "Store origin session key so deferred dispatches route announces correctly") — the channel and chat were just missed.

### Impact

Any team whose lead is reached through `delegate` is write-only: work happens, files land in the team workspace, but no completion, no failure and no blocker escalation ever reaches the user. From the caller's side the run looks like it silently stalled.

### Suggested fix

When recording a task, prefer the delegation origin over the delivery channel — `WorkspaceChannelFromCtx` / `WorkspaceChatIDFromCtx` when set, falling back to `ToolChannelFromCtx` / `ToolChatIDFromCtx` as today. That keeps non-delegated flows byte-identical and makes delegated ones route home.

Possibly related, same session, worth a look while in there: member sessions are keyed `agent:<member>:team:<team-id>:system` and the escalation carries `user_id=system`, i.e. the user scope also degrades to `system` for tasks created by a delegated lead.

### Environment

- `v3.15.0-beta.196`; also verified against `dev` @ `fc8a35e` (`v3.15.0-beta.197`) — unchanged there.
- PostgreSQL store, Kubernetes, Telegram + web channels.

Happy to send a PR for this one too.

yatul:
Fixes #1529. Sibling of #1527 / #1528 — same underlying assumption, no code overlap. The two are independent and can land in either order.

### Problem

`buildAgentLinkRunRequest` is explicit that a delegated run is delivered on an internal channel while its real origin is kept aside:

```go
// preserves the origin's authorization-bearing identity while keeping
// delegation on its internal delivery channel.
Channel:          "delegate",
WorkspaceChannel: req.Channel,
WorkspaceChatID:  req.ChatID,
```

The team tools never consulted that origin, so a task created by a lead reached through `delegate` was stamped with the delivery channel. Nothing is registered for `"delegate"`, so `internal/channels/dispatch.go` dropped every notification about that task — completion, failure, blocker escalation, `ask_user`:

```
16:11:39  INFO teammate announce: batch processed  session=delegate:019f4566:brain:e7ee9913-...
16:11:39  WARN unknown channel for outbound message channel=delegate
```

The delegatee's *first* answer still arrived, because it travels back as the delegation result rather than through a channel. Everything the lead said afterwards was lost — one session produced 16 such drops, including an escalation the user needed:

> Reviewer is blocked: tetris.html does not exist, the coder's generation failed. Restarted generation (running), re-pointed the review at it.

### Change

Add `OriginChannelFromCtx` / `OriginChatIDFromCtx` beside the existing workspace-scope helpers, and use them where team scoping and notification routing are decided: task records, list/search scoping, dispatch fallbacks, event payloads, escalation tasks, `ask_user`, leader notifications. Both resolve to the identity when no delegation origin is present, so non-delegated flows are untouched.

Create and read had to move together — the create lock key and the `list` chat filter must agree, or a lead would stop seeing its own board.

**Deliberately not changed:** the channel used in authorization decisions — `checkTeamAccess`, `requireLead`, and the approve/reject lead bypass. Those ask "how did this call arrive", not "where should the answer go"; widening them is a separate question and a separate risk.

### Test

Two subtests around `executeCreate`: one shapes the context the way `buildAgentLinkRunRequest` does (delivery channel `"delegate"`, origin preserved separately) and asserts the task records the origin; the other asserts a non-delegated lead still records its own channel and chat. The first fails on `dev`:

```
task.Channel = "delegate", want "telegram" — notifications on the delivery channel are dropped
task.ChatID = "system", want "313683273"
```

`go test ./internal/tools/ ./cmd/... ./internal/gateway/... ./internal/http/... ./internal/agent/...` is green.

### Verified in a live deployment

Built with this change (on top of #1528, so the tasks actually dispatch) and deployed to Kubernetes. Same delegated-lead flow:

```
16:39:49  v3.run.completed agent=brain
16:39:49  team_tasks.dispatch: sent task to agent agent_key=coder
16:41:03  v3.run.completed agent=coder
16:41:03  team_tasks.dispatch: sent task to agent agent_key=reviewer
16:41:04  run_id=teammate-announce-brain-1  session_key=agent:019f4566-...:http-313683273-5fa7002d
```

That last line is the point: the completion announce is now delivered into the **caller's** session instead of a delegation session that closed minutes earlier. Drops on `channel=delegate` went from 16 in the previous session to 2, and both remaining ones come from tasks created *before* the change — their bad channel is already persisted, so the fix is forward-looking only. Existing stuck tasks keep routing nowhere until they are recreated; worth a note in release notes, or a migration if you would rather backfill from `TaskMetaOriginSession`.

### DIFF
--- a/internal/tools/context_keys.go
+++ b/internal/tools/context_keys.go
@@ -660,6 +660,27 @@ func WorkspaceChatIDFromCtx(ctx context.Context) string {
 	return ""
 }
 
+// OriginChannelFromCtx returns the channel a run ultimately originated from.
+// A delegated run is delivered on the internal "delegate" channel while the
+// real origin is preserved separately (see buildAgentLinkRunRequest), so team
+// scoping and notification routing must follow the origin — addressing them to
+// the delivery channel makes the outbound dispatcher drop them. Identity for
+// non-delegated runs, where the workspace channel is unset.
+func OriginChannelFromCtx(ctx context.Context) string {
+	if c := WorkspaceChannelFromCtx(ctx); c != "" {
+		return c
+	}
+	return ToolChannelFromCtx(ctx)
+}
+
+// OriginChatIDFromCtx is the chat counterpart of OriginChannelFromCtx.
+func OriginChatIDFromCtx(ctx context.Context) string {
+	if c := WorkspaceChatIDFromCtx(ctx); c != "" {
+		return c
+	}
+	return ToolChatIDFromCtx(ctx)
+}
+
 // --- Pending team task dispatch (post-turn processing) ---
 
 const ctxPendingDispatch toolContextKey = "tool_pending_team_dispatch"
--- a/internal/tools/team_event_helpers.go
+++ b/internal/tools/team_event_helpers.go
@@ -118,8 +118,8 @@ func WithProgress(percent int, step string) TaskEventOption {
 func WithContextInfo(ctx context.Context) TaskEventOption {
 	return func(p *protocol.TeamTaskEventPayload) {
 		p.UserID = store.UserIDFromContext(ctx)
-		p.Channel = ToolChannelFromCtx(ctx)
-		p.ChatID = ToolChatIDFromCtx(ctx)
+		p.Channel = OriginChannelFromCtx(ctx)
+		p.ChatID = OriginChatIDFromCtx(ctx)
 		p.PeerKind = ToolPeerKindFromCtx(ctx)
 		p.LocalKey = ToolLocalKeyFromCtx(ctx)
 	}
--- a/internal/tools/team_task_origin_channel_test.go
+++ b/internal/tools/team_task_origin_channel_test.go
@@ -1,0 +1,93 @@
+package tools
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/nextlevelbuilder/goclaw/internal/store"
+)
+
+// A lead reached through `delegate` runs on the internal "delegate" delivery
+// channel, while the real origin is preserved separately. A task must record
+// the origin: notifications addressed to the delivery channel have no
+// registered handler and are dropped by the outbound dispatcher, so the caller
+// never hears about completions, failures or blocker escalations.
+func TestCreateRecordsDelegationOriginNotDeliveryChannel(t *testing.T) {
+	mb, tool, _, _, ctx := newTestTeamSetup()
+
+	// Shape the context the way a delegated run is built (buildAgentLinkRunRequest):
+	// delivery channel is "delegate", origin preserved as the workspace channel/chat.
+	ctx = WithToolChannel(ctx, "delegate")
+	ctx = WithToolChatID(ctx, "system")
+	ctx = WithWorkspaceChannel(ctx, "telegram")
+	ctx = WithWorkspaceChatID(ctx, "313683273")
+
+	ptd := NewPendingTeamDispatch()
+	ptd.MarkListed()
+	ctx = WithPendingTeamDispatch(ctx, ptd)
+
+	result := tool.Execute(ctx, map[string]any{
+		"action":      "create",
+		"subject":     "Origin routing",
+		"description": "Task created by a delegated lead",
+		"assignee":    "member-agent",
+	})
+	if result.IsError {
+		t.Fatalf("unexpected error: %s", result.ForLLM)
+	}
+	if !strings.Contains(result.ForLLM, "Task created") {
+		t.Fatalf("expected 'Task created', got: %s", result.ForLLM)
+	}
+
+	mb.taskStore.mu.Lock()
+	var task *store.TeamTaskData
+	for _, v := range mb.taskStore.tasks {
+		task = v
+	}
+	mb.taskStore.mu.Unlock()
+	if task == nil {
+		t.Fatal("no task was created")
+	}
+	if task.Channel != "telegram" {
+		t.Errorf("task.Channel = %q, want %q — notifications on the delivery channel are dropped", task.Channel, "telegram")
+	}
+	if task.ChatID != "313683273" {
+		t.Errorf("task.ChatID = %q, want %q", task.ChatID, "313683273")
+	}
+}
+
+// Without a delegation origin the resolution is the identity: a task keeps the
+// channel and chat of the run that created it.
+func TestCreateKeepsOwnChannelWithoutDelegationOrigin(t *testing.T) {
+	mb, tool, _, _, ctx := newTestTeamSetup()
+
+	ptd := NewPendingTeamDispatch()
+	ptd.MarkListed()
+	ctx = WithPendingTeamDispatch(ctx, ptd)
+
+	result := tool.Execute(ctx, map[string]any{
+		"action":      "create",
+		"subject":     "Direct routing",
+		"description": "Task created by a lead talking to the user directly",
+		"assignee":    "member-agent",
+	})
+	if result.IsError {
+		t.Fatalf("unexpected error: %s", result.ForLLM)
+	}
+
+	mb.taskStore.mu.Lock()
+	var task *store.TeamTaskData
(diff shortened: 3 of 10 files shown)


yatul:
Addressed the merge gates from the #1529 triage — both named tests are in, and each fails when the fix is reverted.

**`TestDelegatedLeadCompletionNotifiesOrigin`** — completes a task raised in a delegated context and asserts the completion event is addressed to the caller's origin. Reverting `team_event_helpers.go` to `ToolChannelFromCtx` fails it with:

```
completion event addressed to delegate/system, want telegram/313683273 —
the delivery channel has no registered handler and the notification is dropped
```

**`TestCompletionNotificationStaysWithinItsOwnOrigin`** — the guard you asked for: two tasks from different origins on one board, complete one, assert no completion notification carries the other origin.

Together with the two tests already in the PR (create persists the origin; a non-delegated run is unchanged) that covers persistence, notification routing, isolation, and the identity case.

`go test ./internal/tools/ ./cmd/... ./internal/gateway/... ./internal/http/... ./internal/agent/...` is green.

clark-cant (approved):
## Review: fix(teams): scope team tasks to the delegation origin, not the delivery channel

**Verdict: Approve** ✅

**Risk level:** Low — focused, consistent change across 10 files with comprehensive tests.

**Mandatory gates:**
- Duplicate / prior implementation: clear — no overlap with existing work. Sibling PR #1537 addresses a different concern (team workspace for delegated lead).
- Project standards: clear — follows established context-key helper pattern (WorkspaceChannelFromCtx/WorkspaceChatIDFromCtx).
- Strategic necessity: clear value — fixes 16 notification drops in a single session, including missed blocker escalations.

**Findings:** None. The change is clean:
- New `OriginChannelFromCtx`/`OriginChatIDFromCtx` helpers correctly prefer workspace origin over delivery channel, with identity fallback for non-delegated runs.
- All call sites consistently updated (create, list, search, followup, attach, dispatch, escalation, event helpers, validation).
- Authorization paths deliberately untouched — correct security boundary awareness.
- 4 tests cover: delegation origin persistence, non-delegated identity, completion notification routing, and cross-origin isolation.
- PR body is excellent: problem statement, fix rationale, test strategy, live deployment verification with before/after evidence.

*Posted by github-maintain at 2026-09-01T13:42:00Z*

yatul:
Force-pushed to strip build output that should never have been in this branch.

`internal/webui/dist` is not tracked upstream — zero files on `dev` and in every release; only `internal/webui/*.go` is. `.gitignore` covers `ui/web/dist/`, but not the directory the UI build copies into, so a `git add -A` swept 411 built assets into the fix commit.

The diff was 421 files for a ten-line behavioural change. That is on me, and it made this PR unreviewable at a glance — quite possibly why it has sat while its siblings (#1528, #1533) were triaged and merged.

Now:

```
internal/tools/context_keys.go                   21 +
internal/tools/team_event_helpers.go              4 +-
internal/tools/team_task_origin_channel_test.go 204 +
internal/tools/team_tasks_create.go               6 +-
internal/tools/team_tasks_followup.go             4 +-
internal/tools/team_tasks_mutations.go            2 +-
internal/tools/team_tasks_read.go                 8 +-
internal/tools/team_tool_dispatch.go              4 +-
internal/tools/team_tool_helpers.go               4 +-
internal/tools/team_tool_validation.go            4 +-
10 files changed, 243 insertions(+), 18 deletions(-)
```

Both commits keep their original messages, authorship and dates; only the file set of the first changed. Verified: diffing the old head against the new one while excluding `internal/webui/dist` produces no output, so nothing else moved.

Nothing else about the PR changed — the fix and `TestDelegatedLeadCompletionNotifiesOrigin` / `TestCompletionNotificationStaysWithinItsOwnOrigin` are as reviewed in the triage on this issue. `go test ./cmd/... ./internal/tools/ ./internal/agent/... ./internal/gateway/... ./internal/http/...` is green on `4b9337f0`.

Unrelated, but perhaps worth a line in `.gitignore` upstream: `internal/webui/dist/` is a build artefact and easy to commit by accident this way.

yatul:
Rebased onto current `dev` (no conflicts, same two commits). The earlier red `go` job was the teammate-streaming test race fixed in #1556, which this branch now includes; `go test ./cmd/ ./internal/tools/ ./internal/agent/` passes locally.

PR #1533, opened 2026-08-25
fix(teams): stream teammate runs so long generations keep the connection alive

### LINKED ISSUE #1532: Team member runs are never streamed, so long generations die on "http2: timeout awaiting response headers"
### Summary

`handleTeammateMessage` pins `Stream: false`, so every team member run — coder, reviewer, researcher — issues a non-streamed provider call. There is no setting that changes this.

On a slow reasoning model that means the connection stays silent for the entire generation while the provider buffers the full response, and the request eventually dies:

```
iter 0 think: llm call: litellm: request failed:
  Post "https://litellm.yatul.ru/v1/chat/completions": http2: timeout awaiting response headers
```

The same model, same provider, same prompt size is fine on an ordinary channel run, because those stream by default.

### Reproduction

1. Give a team member a task whose answer takes minutes to generate — e.g. "write a complete single-file HTML game", with the agent's `max_tokens` raised enough to actually fit it.
2. The member run makes one non-streamed call and hangs.
3. After ~15 minutes the run fails with `timeout awaiting response headers`, `tool_call_count=0`, `total_output_tokens=0`. Nothing was produced and nothing indicates why to the user.

### Evidence

Before, on the same task and model (`openrouter/stealth/ox-alpha` via a LiteLLM proxy):

```
17:35:35 → 17:50:36   one LLM span, ~15 minutes
status=error  tools=0  out_tok=0
iter 0 think: llm call: ... http2: timeout awaiting response headers
```

After flipping this run to streamed, the identical task on the identical model:

```
coder: completed  iterations=18  tools=22  out_tok=22432  duration=570s
snake.html — 27 KB, written in 7 append chunks, well-formed
```

Three consecutive failures before, first-try success after.

Note the interaction with `max_tokens`: at the default 8192 the same task failed differently — `llm response truncated before content after compaction`. Raising the cap so the answer fits is what exposes this timeout. The two limits hide each other, which makes the failure hard to attribute.

### Root cause

`cmd/gateway_consumer_handlers.go`, in `handleTeammateMessage`:

```go
outCh := deps.Sched.Schedule(schedCtx, scheduler.LaneTeam, agent.RunRequest{
    ...
    Stream:          false,
```

Compare with the other entry points:

| path | `Stream` |
|---|---|
| ordinary channels (`gateway_consumer_normal.go`) | from channel config — `Streaming *bool // default true` |
| WS chat (`gateway/methods/chat.go`) | from the client's request |
| **team members (`handleTeammateMessage`)** | **hardcoded `false`** |
| announces, cron, subagent announces | `false` |

`internal/providers/defaults.go` sets `ResponseHeaderTimeout: 300 * time.Second`, with a comment noting it was already raised from 180s because reasoning models "emit nothing at all while thinking". A non-streamed member run is exactly that case, with no upper bound on how long the model reasons.

### Why streaming is safe on this path

The obvious worry is that a partially streamed answer leaks to the caller or gets processed incomplete. It cannot, for three independent reasons:

1. `emitChunk` only publishes `ChatEventThinking` / `ChatEventChunk` run events — it never publishes an outbound message.
2. `handleTeammateMessage` never registers the run with the channel manager, and `HandleAgentEvent` starts with `m.runs.Load(runID); if !ok { return }`. The chunks are dropped on the first line; there is nothing to deliver them to.
3. The task result is unchanged: `ChatStream` returns the complete response once the stream ends, and `resolveTeamTaskOutcome` reads `outcome.Result`, not chunks.

So streaming here buys connection liveness — response headers arriving immediately — and changes nothing about what the caller receives or when the answer is considered final. The only visible difference is that a dashboard subscriber can watch the member work.

### Environment

- `v3.15.0-beta.196`, verified against `dev` @ `fc8a35e` (`v3.15.0-beta.197`) — unchanged there.
- PostgreSQL store, Kubernetes, model reached through a LiteLLM proxy.

Adjacent to #1527 and #1529 (same subsystem, delegated team leads) but independent of both — different failure, no code overlap.

PR follows.

yatul:
Fixes #1532.

### Problem

`handleTeammateMessage` pinned `Stream: false`, so every team member run — coder, reviewer, researcher — made a non-streamed provider call, with no setting to change it. On a slow reasoning model the connection then stays silent for the whole generation while the provider buffers the full response, and `ResponseHeaderTimeout` kills it:

```
iter 0 think: llm call: litellm: request failed:
  Post ".../v1/chat/completions": http2: timeout awaiting response headers
```

Measured on one task (a complete single-file HTML game, `openrouter/stealth/ox-alpha` behind a LiteLLM proxy):

```
before   one LLM span, ~15 min   status=error  tools=0   out_tok=0      (3 attempts, all identical)
after    completed               iterations=18 tools=22  out_tok=22432  duration=570s
```

The same model over the same provider was always fine on ordinary channel runs, which stream by default.

### Change

One line: `Stream: true`, plus a comment recording why — because "why is a background run streamed to nobody" is a fair question to ask of this code later.

Streaming on this path is for connection liveness, not delivery, and cannot leak a partial answer:

- `emitChunk` only publishes `ChatEventThinking` / `ChatEventChunk` run events; it never publishes an outbound message.
- `handleTeammateMessage` never registers the run with the channel manager, and `HandleAgentEvent` opens with `m.runs.Load(runID); if !ok { return }` — the chunks are dropped immediately.
- The task result is untouched: `ChatStream` returns the complete response once the stream ends, and `resolveTeamTaskOutcome` reads `outcome.Result`.

The only visible difference is that a dashboard subscriber can watch a member work.

### Alternatives considered

- **Raising `ResponseHeaderTimeout`** — treats the symptom, and the comment already in `internal/providers/defaults.go` explains why that value should not keep growing: it is the only thing bounding a silently dead connection.
- **Making it configurable** — a knob implies a reason to turn it off, and on this path there is none: nothing consumes the chunks either way. Happy to add one if you disagree.

### Test

No new unit test: the change is a field value on a scheduler request, and asserting it would only restate the assignment. What matters is behavioural and is covered by the before/after above on a live deployment. `go test ./cmd/... ./internal/channels/... ./internal/agent/...` is green.

### Note

The failure is easy to misattribute because it hides behind the per-agent `max_tokens` default of 8192: below that cap the same task fails earlier and differently, with `llm response truncated before content after compaction`. Only after the cap is raised enough for the answer to fit does the timeout surface. Worth keeping in mind when triaging similar reports.

### DIFF
--- a/cmd/gateway_consumer_handlers.go
+++ b/cmd/gateway_consumer_handlers.go
@@ -276,7 +276,17 @@ func handleTeammateMessage(
 		SenderID:        teammateSenderID, // real user who triggered the teammate dispatch (#915)
 		Role:            teammateRole,     // RBAC role for admin bypass during teammate turn (#915)
 		RunID:           fmt.Sprintf("teammate-%s-%s", msg.Metadata[tools.MetaFromAgent], msg.Metadata[tools.MetaToAgent]),
-		Stream:          false,
+		// Streamed for connection liveness, not for delivery. A teammate run is
+		// never registered with the channel manager, so HandleAgentEvent drops its
+		// chunks on the first line and nothing is delivered incrementally; the task
+		// result still comes from the final RunResult. What streaming buys is the
+		// response headers arriving immediately: a non-streamed request to a slow
+		// reasoning model holds a silent connection for the whole generation, and
+		// ResponseHeaderTimeout kills it — observed as
+		// `http2: timeout awaiting response headers` on a member asked to produce a
+		// large file, while the same model over the same provider was fine on a
+		// streamed channel run.
+		Stream:          true,
 		TeamTaskID:      msg.Metadata[tools.MetaTeamTaskID],
 		TeamWorkspace:   msg.Metadata[tools.MetaTeamWorkspace],
 		LeaderAgentID:   msg.Metadata[tools.MetaLeaderAgentID],
--- a/cmd/gateway_consumer_teammate_stream_test.go
+++ b/cmd/gateway_consumer_teammate_stream_test.go
@@ -1,0 +1,83 @@
+package cmd
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/nextlevelbuilder/goclaw/internal/agent"
+	"github.com/nextlevelbuilder/goclaw/internal/bus"
+	"github.com/nextlevelbuilder/goclaw/internal/channels"
+	"github.com/nextlevelbuilder/goclaw/internal/config"
+	"github.com/nextlevelbuilder/goclaw/internal/scheduler"
+	"github.com/nextlevelbuilder/goclaw/internal/tools"
+)
+
+// A teammate run must ask the provider to stream. Without it a slow reasoning
+// model holds a silent connection for the whole generation and
+// ResponseHeaderTimeout kills the request before a single token is produced.
+//
+// The same test pins the other half of the contract: streaming here is for
+// connection liveness only. The run is deliberately never registered with the
+// channel manager, so HandleAgentEvent drops its chunks and nothing reaches a
+// user incrementally — the task result keeps coming from the final RunResult.
+func TestHandleTeammateMessageSchedulesStreamedRun(t *testing.T) {
+	var gotReq agent.RunRequest
+	ran := make(chan struct{})
+
+	sched := scheduler.NewScheduler(
+		scheduler.DefaultLanes(),
+		scheduler.QueueConfig{
+			Mode:          scheduler.QueueModeQueue,
+			Cap:           1,
+			Drop:          scheduler.DropOld,
+			MaxConcurrent: 1,
+		},
+		func(_ context.Context, req agent.RunRequest) (*agent.RunResult, error) {
+			gotReq = req
+			close(ran)
+			return &agent.RunResult{Content: "member deliverable"}, nil
+		},
+	)
+	defer sched.Stop()
+
+	channelMgr := channels.NewManager(nil)
+	deps := &ConsumerDeps{
+		Cfg:        &config.Config{},
+		Sched:      sched,
+		ChannelMgr: channelMgr,
+	}
+
+	msg := bus.InboundMessage{
+		Channel:  tools.ChannelSystem,
+		SenderID: "teammate:dashboard",
+		AgentID:  "coder",
+		Content:  "[Assigned task #1 (id: 00000000-0000-0000-0000-000000000001)]: build something",
+		Metadata: map[string]string{
+			tools.MetaOriginChannel: "telegram",
+			tools.MetaOriginChatID:  "12345",
+			tools.MetaFromAgent:     "brain",
+			tools.MetaToAgent:       "coder",
+		},
+	}
+
+	if !handleTeammateMessage(context.Background(), msg, deps) {
+		t.Fatal("handleTeammateMessage() = false, want true for a teammate: message on the system channel")
+	}
+
+	select {
+	case <-ran:
+	case <-time.After(5 * time.Second):
+		t.Fatal("teammate run was never scheduled")
+	}
+
+	if !gotReq.Stream {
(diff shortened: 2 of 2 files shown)


clark-cant (changes requested):
## Maintainer reviewnn**Summary:** This is a small, focused change that switches teammate runs to streaming to avoid provider response-header timeouts. The rationale is well documented and #1532 provides credible production evidence.nn**Mandatory gates**n- Duplicate / prior implementation: clear; search found only this PR and #1532.n- Project standards: no conflicting standards identified in the changed code path.n- Strategic necessity: clear reliability value for long-running team tasks.nn**Important — add a regression test before merge.** The behavioral contract is more than a field assignment: teammate runs must request streaming, while chunks remain undelivered and the final task result path stays intact. Please add or extend the scheduler/consumer test around `handleTeammateMessage` to assert `RunRequest.Stream == true` and verify the existing final-result behavior is preserved. This protects against a future refactor reverting the liveness fix or accidentally routing chunks to a channel.nn**Verdict:** Request changes — narrow test coverage is required; no other blocking finding in the 1-file diff.nn*Posted by github-maintain cron at 2026-08-25T21:57:28Z*

yatul:
Thanks — fair call, and my "a test would only restate the assignment" line was wrong. The contract has two halves that a refactor can break independently, and only one of them is the field.

Added `TestHandleTeammateMessageSchedulesStreamedRun` (`cmd/gateway_consumer_teammate_stream_test.go`). It drives `handleTeammateMessage` through a real scheduler with a capturing `RunFunc` and asserts:

1. the scheduled `RunRequest.Stream == true` — reverting it to `false` fails the test with the reason spelled out in the failure message;
2. the run is **not** registered with the channel manager (`InterimDeliverySnapshot(runID)` is empty), so a future change that starts routing teammate chunks to a channel trips the same test.

The third property you mentioned — a streamed call still yielding the complete final response — is already covered in `internal/agent/loop_pipeline_callbacks_test.go` (`TestMakeCallLLM_StreamsFinalThinkingWhenNoThinkingChunkArrives` and the surrounding `ChatStream` providers), so I did not duplicate it. Happy to add a dedicated assertion here too if you would rather have it local to the teammate path.

`go test ./cmd/... ./internal/channels/... ./internal/agent/...` is green.

yatul:
Ready for re-review — the requested change landed in `e7e3f25` on 2026-08-26.

`TestHandleTeammateMessageSchedulesStreamedRun` (`cmd/gateway_consumer_teammate_stream_test.go`) drives `handleTeammateMessage` through a real scheduler and pins both halves of the contract you named:

- the scheduled `RunRequest.Stream == true` — reverting it to `false` fails the test;
- the run is not registered with the channel manager (`InterimDeliverySnapshot(runID)` empty), so a later change that starts routing teammate chunks to a channel trips the same test.

The third property, a streamed call still returning the complete final response, is already covered in `internal/agent/loop_pipeline_callbacks_test.go`; say the word if you would rather have it duplicated locally to the teammate path.

I cannot re-request review from here (outside contributor), hence the comment. PR is mergeable, `go test ./cmd/... ./internal/channels/... ./internal/agent/...` green.

clark-cant (approved):
## Re-review after author addressed requested changes

**Summary:** Author pushed the requested regression test in commit `e7e3f25` (2026-08-26). `TestHandleTeammateMessageSchedulesStreamedRun` pins both halves of the contract:

1. `gotReq.Stream == true` — reverting the liveness fix fails the test with a clear reason.
2. `InterimDeliverySnapshot(runID)` is empty — a future refactor that accidentally routes chunks to a channel trips the same test.

The third property (streamed call still returning the complete final response) is already covered by existing `ChatStream` provider tests in `internal/agent/loop_pipeline_callbacks_test.go`.

**Diff review:** Clean. 2 files, +94/-1. The one-line production change (`Stream: true`) is well-commented explaining why streaming is for connection liveness, not delivery. The test is well-structured with a real scheduler and capturing RunFunc.

**Verdict: Approve.** All requested changes addressed. No new findings.

PR #1534, opened 2026-08-26
fix(skills): five defects in slash activation, grants, subagent tool policy, preview and import

mor-duongmh:
Five defects in the skill subsystem, found while porting a large external skill collection onto GoClaw. They are unrelated to each other except by area, so each is its own commit and each can be reverted alone.

Two are security fixes. One of those — subagent tool policy — lets an agent reach files it is otherwise denied, simply by spawning.

## 1. A spawned subagent did not inherit the parent's tool policy

`buildSubagentToolsRegistry` clones the parent registry, then overwrites `exec`, `read_file`, `write_file` and `list_files` with freshly constructed tools. A fresh tool carries none of the hardening the gateway applies at startup (`cmd/gateway_setup.go:213-290`): exec path denials and their exemptions, the shell deny-group toggles, the command keyword allowlist, and the read/write/list deny prefixes covering `config.json`, the internal databases and `delegate/`.

So spawning a subagent widened what an agent could reach. Verified by disabling the new call and running the tests: the subagent's `exec` read `config.json` out of the denied data dir, `read_file` did the same, and `write_file` overwrote it.

The fix copies the policy from the **live parent instances** rather than re-deriving it, so there is one source of truth and a deny path reloaded later through config pub/sub reaches subagents without a second wiring site that can drift. Allow-prefixes are inherited alongside the denials — they are what make the denied roots usable at all, since the skills store sits under the denied data dir.

There is precedent immediately below the new call: `SetSecureCLIStore` patches one other facet of this same class.

## 2. Slash commands bypassed skill grants

The inline `<available_skills>` block has always been filtered by visibility and agent grants, but slash activation read the loader's full skill list. `/<slug>` could activate a managed skill the agent was never granted, and the not-found suggestions disclosed that such skills existed.

**Scope.** The allow list comes from a query over the `skills` table (`internal/store/pg/skills_grants.go:410`), so filesystem-tier skills — the workspace, `.agents`, `~/.agents` and `$GOCLAW_SKILLS_DIR` directories of the five-tier loader — have no row in it. Filtering *every* skill against the list made those tiers unreachable by slash, which would have traded a security hole for a functional regression.

The gate therefore applies to the managed tier only — which is exactly the predicate `skill_search` already uses (`internal/tools/skill_search.go:186-196`: *"Filesystem skills (source != managed) are always allowed"*). Slash now behaves the same way search does, rather than inventing a third rule. Builtins are seeded with `is_system` and returned unconditionally, so they stay reachable either way.

## 3. Preview and runtime disagreed about inline vs search mode

The system-prompt preview exists to show the prompt an agent actually gets, but it decided between inlining skills and search mode on its own terms: it counted tokens with the fallback counter over the fully rendered XML — tags and `<location>` paths included, at roughly runes/2 — while the prompt builder estimated name+description at chars/4. On the same eighteen skills that read **3087 against 944**, so the preview reported search mode for an agent that was running inline.

Both paths now call `shouldInlineSkills`. The estimate keeps the prompt builder's rule, so the decision still costs no rendering.

## 4. Skill import discarded every reference file

Export archives the whole skill directory, but import recognised only `metadata.json`, `SKILL.md` and `grants.jsonl`. The switch had no `default`, so every other file was dropped without a log line. A skill whose `SKILL.md` cites `references/*.md` arrived without them, the import reported success, and the skill failed at its first read.

Import now writes the rest of the tree, with three guards:

- **Reserved names.** `sanitizeRelPath` collapses `./SKILL.md` to `SKILL.md`, and that name does not match the switch (which compares the raw archive path), so a crafted entry landed in the auxiliary set and overwrote the `SKILL.md` that `GuardSkillContent` had already scanned — the scan passed on one file while a different one reached disk. Auxiliary paths resolving to a reserved name are now rejected and logged.
- **Directory entries.** A tar directory entry has a trailing separator and no content; written as a file it took the name a real directory needed and everything beneath was dropped. Whether that happened depended on map iteration order, so the failure was intermittent.
- **Count cap**, logged when it trims, so a truncated import is visible rather than silent.

Path traversal was already handled by `sanitizeRelPath`; there is a test that asserts nothing is written above the skill directory.

## 5. Slash commands broke on a newline

The tokenizer split the skill name off with `strings.Cut(after, " ")` — a literal space. A user who typed the command and pressed Enter before the rest of the message sent `/ck:git\nreview the diff`, which parsed to the target `ck:git\nreview` and matched nothing. The reply was "skill not found" followed by near-matches that **included the skill they had just named**, because the similarity fallback searched the mangled string and still landed beside it.

Multi-line messages are ordinary in Slack and Telegram, so this was reachable in normal use. `cutFirstField` splits on the first run of whitespace; `hasFieldPrefix` replaces the matching loop's `value+" "` prefix check and decodes the following rune properly, so a multi-byte skill name is not truncated. That also stops a plain string prefix from matching: a command for `frontend-design-extra` no longer resolves to `frontend-design`.

## Testing

Five new test files. Every guard was verified by **disabling the fix and confirming the test goes red** — including one round that exposed two of three subagent assertions passing for the wrong reason (the denied files did not exist, so `cat` failed on absence rather than on policy) and one inline-decision case that never reached the token branch because the count check short-circuited first. Both were corrected; `22de521f` is that correction.

- `go build ./...` and `go build -tags sqliteonly ./...`
- `go vet` on the touched packages
- `go test` on `agent`, `http`, `tools`, `cmd`, `skills`
- `go test -race` on the touched paths

## Review record

The branch was put through an adversarial review before this PR: one reviewer per fix, then independent skeptics instructed to refute each finding, with anything uncertain defaulting to refuted. Twenty findings were raised and four survived refutation. All four are fixed here:

| Fix | Finding | Addressed in |
|---|---|---|
| slash grants | gate stranded four loader tiers | `b47b800e` |
| import | `./SKILL.md` overwrote the guard-scanned file | `6e6e7b5a` |
| import | tar directory entries destroyed their own subtree | `6e6e7b5a` |
| preview | test table never reached the token branch | `22de521f` |

Subagent policy inheritance and the tokenizer drew no confirmed findings (0 of 6 and 0 of 3 survived).

## Known follow-ups, not in this PR

- `use_skill` is a marker that logs an activation and tells the agent to open the file itself; the read then goes through `read_file`, which is path-allowed to the skills directories (`cmd/gateway_tools_wiring.go`). An agent that already knows a slug can therefore read an ungranted managed skill's body. Pre-existing, and wider than any of these five — enforcement belongs at the read, not at the marker.
- The preview endpoint now walks the skills tree twice per request (`FilterSkills` and `BuildSummary` each call `ListSkills`, which takes a write lock). The runtime path already did this; a `BuildSummaryFrom([]skills.Info)` renderer would fix both.
- The token estimate counts description **bytes** while `BuildSummary` truncates at 200 **runes**, so a non-ASCII skill catalogue inlines over the intended budget. Pre-existing; changing it shifts runtime verdicts for zh/vi catalogues and deserves its own change.
- The channel handlers prepend reply context and media tags to the message before it reaches the agent, so a slash command sent as a thread reply or with an attachment is no longer at the start of the string and does not fire. Present in Slack, Discord, Telegram, Feishu and WhatsApp; confirmed in Slack.

## Surface parity

Web UI: N/A. CLI: N/A. No wire-contract change — the two widened Go interfaces (`SkillsLoader`, `SkillPreviewBuilder`) are internal, and the preview response shape is unchanged.

## One behaviour change worth calling out

An agent whose accessible-skill list is empty can no longer activate a managed skill by slash; before, it could. That is the point of the fix, but a deployment inadvertently relying on the gap will notice. There is no way to close the hole and keep the old behaviour.

### DIFF
--- a/cmd/gateway_agents.go
+++ b/cmd/gateway_agents.go
@@ -231,19 +231,34 @@ func buildSubagentToolsRegistry(
 ) (*tools.Registry, *tools.ExecTool) {
 	reg := parentReg.Clone()
 	var execTool *tools.ExecTool
+	var readTool *tools.ReadFileTool
+	var writeTool *tools.WriteFileTool
+	var listTool *tools.ListFilesTool
 	if sandboxMgr != nil {
-		reg.Register(tools.NewSandboxedReadFileTool(workspace, restrict, sandboxMgr))
-		reg.Register(tools.NewSandboxedWriteFileTool(workspace, restrict, sandboxMgr))
-		reg.Register(tools.NewSandboxedListFilesTool(workspace, restrict, sandboxMgr))
+		readTool = tools.NewSandboxedReadFileTool(workspace, restrict, sandboxMgr)
+		writeTool = tools.NewSandboxedWriteFileTool(workspace, restrict, sandboxMgr)
+		listTool = tools.NewSandboxedListFilesTool(workspace, restrict, sandboxMgr)
 		execTool = tools.NewSandboxedExecTool(workspace, restrict, sandboxMgr)
-		reg.Register(execTool)
 	} else {
-		reg.Register(tools.NewReadFileTool(workspace, restrict))
-		reg.Register(tools.NewWriteFileTool(workspace, restrict))
-		reg.Register(tools.NewListFilesTool(workspace, restrict))
+		readTool = tools.NewReadFileTool(workspace, restrict)
+		writeTool = tools.NewWriteFileTool(workspace, restrict)
+		listTool = tools.NewListFilesTool(workspace, restrict)
 		execTool = tools.NewExecTool(workspace, restrict)
-		reg.Register(execTool)
 	}
+
+	// These four tools are built fresh, so they start with none of the hardening the
+	// gateway applied to the parent's instances at startup: exec path denials and their
+	// exemptions, shell deny-group toggles, the command keyword allowlist, and the
+	// read/write/list deny prefixes covering config.json, the databases, and delegate/.
+	// Without this, spawning a subagent widened reach — the parent could not touch the
+	// data dir, the subagent could. Inherit from the live parent instances so there is
+	// one source of truth and a later config reload cannot leave subagents behind.
+	inheritParentPathPolicy(parentReg, readTool, writeTool, listTool, execTool)
+
+	reg.Register(readTool)
+	reg.Register(writeTool)
+	reg.Register(listTool)
+	reg.Register(execTool)
 	// Red Team F3: subagent ExecTool must enforce the secure-CLI gate
 	// (and env scrub on fall-through) — without this, a parent agent
 	// can spawn a subagent to bypass the gate via host-inherited env.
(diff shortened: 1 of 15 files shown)


mor-phongdt (approved):

clark-cant:
🤖 **github-maintain triage (cron-safe, metadata-first)**

This PR addresses five real defects in the skill subsystem — two with security impact (subagent tool policy inheritance, slash grants bypass). The description is thorough, the commit structure is clean (one per defect, independently revertable), and the adversarial review record is impressive.

**Current state:**
- 1 approval from @mor-phongdt ✅
- Mergeable: MERGEABLE ✅
- **CI checks: none reported** ⚠️
- mergeStateStatus: UNSTABLE

**Blocker for merge:** No CI checks are reported on this branch. Given the security-sensitive nature of the changes (subagent policy inheritance, path deny lists, skill grants gating), automated verification (go build, go vet, go test, go test -race) must be confirmed green before merge. The PR body describes running these locally — we need them in CI.

**Next steps:**
1. Confirm CI pipeline is configured and running for this branch
2. Once checks are green, this PR is merge-ready

*Posted by github-maintain automation — 2026-08-26T16:25Z*

clark-cant (approved):
## github-maintain review (cron-safe)

**Verdict: Approve**

**Summary:** Five well-scoped defects in the skill system — slash activation grant bypass, literal-space command splitting, subagent tool policy widening, preview/runtime inline-decision disagreement, and import losing reference files or overwriting guarded ones. Each fix is focused, carries regression tests that fail without it, and touches nothing outside its stated scope.

**Risk level:** Medium (security-adjacent: subagent policy inheritance is an authorization boundary fix, import path sanitization is a traversal guard).

**Mandatory gates:**
- Duplicate / prior implementation: clear — no overlapping PR or merged fix found for any of the five defects.
- Project standards: docs found — Go project conventions, test table style, commit message quality all consistent with repo norms.
- Strategic necessity: clear value — closes a slash-command grant bypass, a subagent privilege widening, and an import path traversal; fixes a real UX failure mode (multi-line slash commands); eliminates a preview/runtime disagreement that undermines the preview endpoint's purpose.

**Findings:**
- None at Critical or Important level.
- Suggestion: the import aux-file cap (50) is reasonable; consider surfacing the cap value in a structured log field rather than the count alone, so operators can alert on truncated imports without parsing the message text.

**CI:** No checks reported on the branch. mergeStateStatus=UNSTABLE with empty rollup — merge deferred until CI populates.

*Posted by github-maintain cron-safe automation — 2026-08-31T21:12Z*

PR #1537, opened 2026-08-27
fix(teams): grant a delegated lead read access to its own team

### LINKED ISSUE #1535: A delegated team lead cannot return the team's files: send_file and message are refused, and the artifact outputs dir is unreachable
Sibling of #1527 and #1529 — same theme, that a delegated team lead is a second-class citizen, but a different failure with no code overlap.

### Summary

A team lead reached through `delegate` cannot hand the team's output back to the user by any route. `send_file` and `message` both refuse outright inside a delegation artifact run, and the sanctioned alternative — staging the file into the delegation `outputs/` directory — is unreachable because the lead can neither read the team workspace nor shell out to copy from it.

The work completes, the file sits in the team workspace, and the user gets nothing.

### Evidence

One user request, all three routes tried by the lead in a single run:

```
tool call  agent=brain tool=send_file args_len=161
tool error agent=brain tool=send_file error="delegation files are published only after the delegated run completes"
tool call  agent=brain tool=send_file args_len=146
tool error agent=brain tool=send_file error="delegation files are published only after the delegated run completes"

tool error agent=brain tool=read_file error="access denied: path outside workspace"
WARN security.path_escape path=/app/workspace/teams/<team-id>/system/review-....md

tool error agent=brain tool=exec error="delegated artifact exec requires an active sandbox manager
                                        and sandbox key; host execution is not allowed"
```

The contrast is decisive: **the same agent, the same file, addressed directly instead of through `delegate`, sends it without trouble.** Only the delegation wrapper makes it impossible.

### Root cause

Two independent blocks, either one sufficient.

**1. Blanket refusal.** `internal/tools/send_file.go:83`, and verbatim in `internal/tools/message.go:122`:

```go
if IsDelegationArtifactRun(ctx) {
    return ErrorResult("delegation files are published only after the delegated run completes")
}
```

The intent is clear and reasonable: a delegated run publishes through its artifact exchange, not by pushing files sideways. But it forecloses the case where the file the lead wants to return is not a delegation artifact at all — it is the team's deliverable, produced by a member in the shared team workspace.

**2. The lead cannot stage the file either.** Reading and listing already honour the team workspace — `internal/tools/filesystem.go:188`, `filesystem_list.go:105`, `send_file.go:97` all resolve paths through `allowedWithTeamWorkspace(ctx, ...)`. That allowance comes from `ToolTeamWorkspaceFromCtx`, which is populated from `RunRequest.TeamWorkspace` (`internal/agent/loop_context.go:251`).

`handleTeammateMessage` sets that field for team **members**. `buildAgentLinkRunRequest` — the delegated path — does not set it at all. So a delegated lead runs without its own team's workspace in context, and every read of it is denied. With `exec` also barred in artifact runs, there is no way to copy the file into `outputs/`.

### Impact

Any team whose lead is reached through `delegate` — the natural shape when a user-facing assistant fans work out to a specialist team — can produce files but never deliver them. From the user's side the run looks like it silently refused; the agent typically reports success, having no way to tell that the tool was blocked rather than the file missing.

### Suggested direction

I would rather agree the shape before sending a patch, since the refusal in (1) is deliberate.

Two candidates:

- **Narrow the refusal.** Let `send_file` proceed in a delegation artifact run when the resolved path lies inside the team workspace, keeping the block for the delegation exchange itself. Smallest change; leaves the exchange hermetic.
- **Give the delegated lead its team workspace.** Populate `TeamWorkspace` for delegated leads the way `handleTeammateMessage` does for members, so reads succeed and the lead can stage into `outputs/` through the normal artifact flow. More faithful to the existing design, but needs a rule for a lead who leads several teams.

Happy to implement either — say which you prefer and I will send it with tests.

### Environment

- `v3.15.0-beta.196` plus local patches for #1527/#1529; the code paths above are unchanged on `dev` @ `fc8a35e`.
- PostgreSQL store, Kubernetes, Telegram + web channels.

yatul:
Fixes #1535. Implements the direction from your triage — give the delegated lead its own team context rather than loosening `send_file`/`message` — but through a different mechanism than my first attempt.

**Stacked on #1530.** The team workspace path is derived from the origin chat, which is exactly what #1530 corrects; based on `dev` alone the lead would resolve a different directory than its own tasks use. Review only the last commit, or merge #1530 first and this becomes a three-file diff.

### The mechanism

My first attempt set `RunRequest.TeamWorkspace` on the delegated run. That is rejected outright, and deliberately — `internal/agent/loop_context.go:233`:

```go
if isArtifactDelegation {
    if req.TeamWorkspace != "" ||
        !validateDelegationArtifactWorkspace(req.DelegationID, req.DelegateInputsPath, req.DelegateOutputsPath) {
        return contextSetupResult{}, fmt.Errorf("invalid delegation artifact workspace")
    }
```

The invariant is right: that field does not only grant read access, it also becomes `ToolWorkspace`, which would displace the exchange outputs directory. Deployed, every delegation died at setup. This version does not go near the field.

The separation the lead needs already exists in the code. A lead addressed **directly** resolves its team at `loop_context.go:285-321` and gets

```go
ctx = tools.WithToolTeamWorkspace(ctx, wsDir)
ctx = tools.WithToolTeamRoot(ctx, teamRoot)
```

with no `WithToolWorkspace` call — a read allowance, not an override. Only the `!isArtifactDelegation` gate keeps a delegated lead out of it. So the grant is made inside the artifact branch instead, from the same inputs. The guard above stays exactly as strict as it was, and no new field is added to `RunRequest`.

### Both paths are needed

The workspace alone covers only the lead's own chat leaf. The deliverable in the report sits at `teams/<id>/system/review-....md` — written by a member under a different chat scope. `buildAllowedPrefixes` adds the team root for reads and not for writes (`internal/tools/filesystem.go:366-370`), which is precisely the asymmetry this case wants: the lead reads the team's output, per-chat write isolation is untouched.

### What is deliberately not granted

`ToolTeamID` is not set. It switches on the workspace interceptor's write validation, file-change broadcast and task attachment (`internal/tools/workspace_interceptor.go:36,114,203`) — none of which a delegated run should trigger, and none of which reading needs.

Hermeticity is unchanged: `send_file` and `message` still refuse inside an artifact run, and publication still happens through delegation outputs on completion. There is a test pinning that.

### The multi-team rule

As requested, ambiguity is not resolved silently: an agent leading more than one active team gets **nothing**. `store.GetTeamForAgent` would answer in one call, but it is `ORDER BY (lead_agent_id = $1) DESC LIMIT 1` — it would quietly pick a team, which is the choice this must not make on its own.

### Tests

| property | test |
|---|---|
| the guard is crossed, outputs not displaced | `TestInjectContext_DelegatedLeadReadsTeamWithoutDisplacingOutputs` |
| an agent leading no team gains nothing | `TestInjectContext_DelegatedNonLeadGainsNothing` |
| resolution rule: exactly one active led team | `SingleLedTeamResolves`, `MultipleLedTeamsResolveNothing`, `NonLeadResolvesNothing`, `InactiveTeamIgnored` |
| shared vs isolated path shape | `SharedTeamCollapsesToRoot` |
| no cross-team access | `SeparateLeadsGetSeparateWorkspaces` |
| store failure or no store grants nothing | `StoreFailureResolvesNothing`, `NoTeamStoreResolvesNothing` |
| team root widens reads, not writes | `TestTeamRootDoesNotWidenWrites` |
| direct sending still blocked | `TestSendFileStaysBlockedInDelegationArtifactRun` |

The first one is the one that matters: my previous attempt had eight green unit tests and still broke every delegation, because not one of them crossed the guard in `injectContext`. This one drives `injectContext` itself and asserts both halves — the run is not refused, and `ToolWorkspace` is still the exchange outputs directory. Reverting the change fails it.

`go test ./cmd/... ./internal/tools/ ./internal/agent/... ./internal/gateway/... ./internal/http/...` is green.

### Verified running, not just under test

Given how the first attempt failed, unit tests alone are not evidence. Built and deployed to a live cluster:

1. the delegation reaches the lead — the previous `invalid delegation artifact workspace` is gone;
2. the lead reads `teams/<id>/system/review-....md`, the path from the report, in a chat scope that is not its own — reachable only through the team root;
3. the content comes back verbatim to the caller;
4. separately, the lead reads a team file and writes it into its own `outputs/` — the sanctioned publication route the report said was unreachable — byte count matching the source.

### Note on the branch

Force-pushed twice: once to replace the first approach, and once to strip 411 built `internal/webui/dist` assets that rode in from #1530's commit. See the note on #1530. Nothing else moved; the diff is now the three files this change actually touches, plus #1530's.

### DIFF
--- a/internal/agent/delegated_lead_team.go
+++ b/internal/agent/delegated_lead_team.go
@@ -1,0 +1,86 @@
+package agent
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/google/uuid"
+
+	"github.com/nextlevelbuilder/goclaw/internal/store"
+	"github.com/nextlevelbuilder/goclaw/internal/tools"
+)
+
+// delegatedLeadTeamRead is the read-only view of its own team granted to a team
+// lead reached through delegate: workspace is the directory that team's tasks
+// resolve to, root is the team directory above it — the one that makes files
+// written by members under a different chat scope reachable.
+//
+// Both are allowances only. Neither may become the active workspace: in a
+// delegation artifact run that stays the exchange outputs directory, and
+// displacing it breaks the exchange.
+type delegatedLeadTeamRead struct {
+	workspace string
+	root      string
+}
+
+func (r delegatedLeadTeamRead) ok() bool { return r.workspace != "" && r.root != "" }
+
+// resolveDelegatedLeadTeamRead returns the team paths a delegated lead may read,
+// or the zero value when no unambiguous team answers.
+//
+// A team member run carries its team workspace in dispatch metadata and a lead
+// addressed directly resolves one below; a delegated lead had neither, so it
+// could not read its own team's files nor stage them into the delegation outputs
+// directory — the team's deliverables were unreachable to the only agent able to
+// hand them back (#1535).
+//
+// The paths mirror what injectContext resolves for a directly addressed lead and
+// what team_tasks_create stores on every task the lead creates; those must agree
+// or the lead is handed a sibling directory. l.dataDir is already tenant-scoped
+// (resolver.go: config.TenantDataDir), so TenantLayer must NOT be reapplied here.
+//
+// Ambiguity is deliberately not resolved: an agent leading several active teams
+// gets nothing rather than a guess, since the delegation does not say which team
+// it concerns. store.GetTeamForAgent would answer in one call, but it orders by
+// lead-ness and takes LIMIT 1 — it would silently pick one team, which is exactly
+// the choice this must not make on its own.
+func (l *Loop) resolveDelegatedLeadTeamRead(ctx context.Context, req *RunRequest) delegatedLeadTeamRead {
+	if l.teamStore == nil || l.agentUUID == uuid.Nil || l.dataDir == "" {
+		return delegatedLeadTeamRead{}
+	}
+	teams, err := l.teamStore.ListTeams(ctx)
+	if err != nil {
+		slog.Warn("delegate: cannot resolve lead team read access",
+			"agent_id", l.agentUUID, "error", err)
+		return delegatedLeadTeamRead{}
+	}
+	var lead *store.TeamData
+	for i := range teams {
+		if teams[i].LeadAgentID != l.agentUUID || teams[i].Status != store.TeamStatusActive {
+			continue
+		}
+		if lead != nil {
+			slog.Debug("delegate: agent leads several active teams, team read access not granted",
+				"agent_id", l.agentUUID)
+			return delegatedLeadTeamRead{}
+		}
+		lead = &teams[i]
+	}
+	if lead == nil {
+		return delegatedLeadTeamRead{}
+	}
+
+	// Origin chat, not the delivery channel: team_tasks_create derives the same
+	// path from OriginChatIDFromCtx, which prefers the workspace chat ID.
+	chatID := req.WorkspaceChatID
+	if chatID == "" {
+		chatID = req.ChatID
+	}
+	return delegatedLeadTeamRead{
+		workspace: tools.ResolveWorkspace(l.dataDir,
+			tools.TeamLayer(lead.ID),
+			tools.UserChatLayer(chatID, tools.IsSharedWorkspace(lead.Settings)),
+		),
+		root: tools.ResolveWorkspace(l.dataDir, tools.TeamLayer(lead.ID)),
+	}
+}
--- a/internal/agent/delegated_lead_team_test.go
+++ b/internal/agent/delegated_lead_team_test.go
@@ -1,0 +1,210 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/google/uuid"
+
+	"github.com/nextlevelbuilder/goclaw/internal/store"
+	"github.com/nextlevelbuilder/goclaw/internal/tools"
+)
+
+// mockTeamStoreLead serves ListTeams and nothing else; the embedded interface is
+// nil on purpose so an unexpected store call fails loudly instead of silently.
+type mockTeamStoreLead struct {
+	store.TeamStore
+	teams []store.TeamData
+	err   error
+}
+
+func (m *mockTeamStoreLead) ListTeams(context.Context) ([]store.TeamData, error) {
+	return m.teams, m.err
+}
+
+func leadTeam(id, lead uuid.UUID, status string, settings string) store.TeamData {
+	td := store.TeamData{
+		BaseModel:   store.BaseModel{ID: id},
+		LeadAgentID: lead,
+		Status:      status,
+	}
+	if settings != "" {
+		td.Settings = json.RawMessage(settings)
+	}
+	return td
+}
+
+// newLeadTestLoop mirrors newArtifactTestLoop but wires a team store and pins the
+// agent identity, so the resolver has something to resolve against.
+func newLeadTestLoop(root string, agentID uuid.UUID, teams store.TeamStore) *Loop {
+	return NewLoop(LoopConfig{
+		ID:        "brain",
+		AgentUUID: agentID,
+		TenantID:  store.MasterTenantID,
+		Workspace: filepath.Join(root, "agents", "brain"),
+		DataDir:   root,
+		Sessions:  &nopSessionStore{},
+		TeamStore: teams,
+	})
+}
+
+func TestResolveDelegatedLeadTeamRead(t *testing.T) {
+	root := t.TempDir()
+	lead := uuid.New()
+	other := uuid.New()
+	teamA := uuid.New()
+	teamB := uuid.New()
+	req := &RunRequest{ChatID: "chat-1", WorkspaceChatID: "chat-1"}
+
+	t.Run("SingleLedTeamResolves", func(t *testing.T) {
+		s := &mockTeamStoreLead{teams: []store.TeamData{
+			leadTeam(teamA, lead, store.TeamStatusActive, ""),
+			leadTeam(teamB, other, store.TeamStatusActive, ""),
+		}}
+		got := newLeadTestLoop(root, lead, s).resolveDelegatedLeadTeamRead(context.Background(), req)
+		if !got.ok() {
+			t.Fatalf("nothing resolved for an agent leading exactly one team: %#v", got)
+		}
+		if !strings.Contains(got.workspace, teamA.String()) {
+			t.Errorf("workspace %q does not belong to the led team %s", got.workspace, teamA)
+		}
+		if strings.Contains(got.workspace, teamB.String()) || strings.Contains(got.root, teamB.String()) {
+			t.Errorf("resolved paths leak into an unrelated team: %#v", got)
+		}
+		// Isolated is the default: the workspace is a chat-scoped leaf below the
+		// root, and the root is what makes a peer's chat scope reachable at all.
+		if got.workspace != filepath.Join(got.root, "chat-1") {
+			t.Errorf("workspace %q is not the chat leaf of root %q", got.workspace, got.root)
+		}
+	})
+
+	t.Run("SharedTeamCollapsesToRoot", func(t *testing.T) {
+		s := &mockTeamStoreLead{teams: []store.TeamData{
+			leadTeam(teamA, lead, store.TeamStatusActive, `{"workspace_scope":"shared"}`),
+		}}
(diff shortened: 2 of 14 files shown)


yatul:
Converting to draft — **this approach is wrong and I found out by deploying it.** Flagging before anyone spends review time on it.

Setting `RunRequest.TeamWorkspace` on a delegated run is explicitly rejected, and deliberately so — `internal/agent/loop_context.go:233`:

```go
if isArtifactDelegation {
    if req.TeamWorkspace != "" ||
        !validateDelegationArtifactWorkspace(req.DelegationID, req.DelegateInputsPath, req.DelegateOutputsPath) {
        return contextSetupResult{}, fmt.Errorf("invalid delegation artifact workspace")
    }
```

and the team-workspace application below it is gated on `!isArtifactDelegation` twice over. In a live cluster every delegation then died at setup:

```
WARN delegate.async.failed to=brain error="setup context: inject context: invalid delegation artifact workspace"
```

The invariant makes sense: `req.TeamWorkspace` does not just grant read access, it also becomes `ToolWorkspace`, which would displace the artifact outputs directory and break the exchange. My unit tests passed precisely because they exercised the resolver and the allow-list, never this guard — my fault for not covering the wiring end to end.

The idea still seems right; the mechanism has to change. What the lead needs is the *read allowance* (`WithToolTeamWorkspace`, feeding `allowedWithTeamWorkspace`) **without** the workspace override — so it cannot travel through `req.TeamWorkspace` as long as that field carries both meanings.

Two shapes, and since the guard is deliberate I would rather you pick:

1. **A separate field**, e.g. `RunRequest.TeamReadWorkspace`, applied inside the artifact branch as a read allowance only. Leaves the existing invariant and its guard untouched.
2. **Split the meaning of `TeamWorkspace`** — keep rejecting it as a workspace override in artifact runs, but let the artifact branch adopt it for reads. Fewer fields, but weakens a guard that currently says something simple.

I lean towards (1). Happy to push it once you say which.

Sorry for the noise — should have caught this before opening.

clark-cant (approved):
## Review: fix(teams): give a delegated lead its own team workspace

**Summary:** Resolves #1535 by populating the delegated lead's TeamWorkspace from its own team context, restoring read access to team files that was always available to team members but missing from the agent-link delegation path. Stacked on #1530 (already merged).

**Risk level:** Low — focused Go changes (+450/-16 across 14 source files) with comprehensive regression tests. The 400+ webui dist files are an unrelated UI rebuild bundled in.

**Mandatory gates:**
- Duplicate / prior implementation: clear — no overlapping PR found; #1530 (prerequisite) already merged
- Project standards: aligned — uses existing patterns (ListTeams walk, allowedWithTeamWorkspace validation, OriginChannelFromCtx helpers)
- Strategic necessity: clear value — completes the delegated-lead capability chain (dispatch → routing → workspace access)

**Findings:**
- **Suggestion:** The PR bundles ~400 webui dist asset files unrelated to the Go fix. Consider separating UI rebuilds into dedicated commits/PRs for cleaner history and easier bisect.
- **Suggestion:** The multi-team rule (agent leading >1 team gets no workspace) is correct defensive behavior. Worth documenting this constraint in team architecture docs.

**Verdict:** Approve — well-designed narrow fix with proper test coverage (8 test cases covering resolution, isolation, hermeticity). No Critical or Important findings.

*Posted by /ck:review-pr at 2026-08-31T20:44:00Z*

yatul:
Ready for review — reworked and out of draft.

I put this in draft on 27 August because the first approach was wrong: it set `RunRequest.TeamWorkspace` on the delegated run, which `injectContext` rejects by design, and every delegation died at setup. I asked which of two shapes you preferred — a new `TeamReadWorkspace` field, or splitting the meaning of `TeamWorkspace`.

Neither turned out to be necessary, so I did not wait for the answer. The read-allowance-without-override separation already exists in the code: a lead addressed directly gets `WithToolTeamWorkspace` + `WithToolTeamRoot` and no `WithToolWorkspace` (`loop_context.go:285-321`). Only the `!isArtifactDelegation` gate excluded a delegated lead. The grant now happens inside the artifact branch from the same inputs, the guard on `req.TeamWorkspace` is untouched, and `RunRequest` gains no field. The PR body has the details.

Two things worth your attention:

**The team root is not optional.** The workspace alone covers only the lead's own chat leaf, and the deliverable in the report lives under another member's chat scope. `buildAllowedPrefixes` already adds the root for reads and not for writes, so this grants exactly the asymmetry the case needs.

**`ToolTeamID` is deliberately left unset**, since it would switch on the workspace interceptor's write validation and broadcast, which a delegated run has no business triggering.

I also verified it running, not only under test — deployed and drove a real delegation end to end, including the lead reading the exact path from the report and staging a file into its delegation outputs. My previous attempt passed eight unit tests and still broke everything, so I no longer treat green tests as evidence here. There is now a test that crosses the guard those eight missed.

yatul:
Flagging that the approval on this PR is stale, rather than quietly merging on it.

The review was submitted on 31 August against `019cea10`. The branch head is now `91376643`, and the approval carried across the force-push. Three details in the review text confirm it describes the earlier state and not the current one:

- *"populating the delegated lead's TeamWorkspace from its own team context"* — that is the first approach, which I withdrew. Setting `RunRequest.TeamWorkspace` on a delegated run is rejected by `injectContext` by design, because that field also becomes `ToolWorkspace` and would displace the exchange outputs directory. Deployed, it killed every delegation at setup.
- *"The 400+ webui dist files are an unrelated UI rebuild bundled in"* — those were build output that rode in from #1530's commit. Both branches were force-pushed to strip them; the diff is now 14 files, all source.
- *"Stacked on #1530 (already merged)"* — #1530 is still open. This PR still depends on it.

What is on the branch now takes a different route: the read-allowance-without-override separation already exists in `loop_context.go:285-321`, where a directly addressed lead gets `WithToolTeamWorkspace` + `WithToolTeamRoot` and no `WithToolWorkspace`. Only the `!isArtifactDelegation` gate excluded a delegated lead, so the grant is made inside the artifact branch from the same inputs. The guard on `req.TeamWorkspace` is untouched and `RunRequest` gains no field. Full reasoning is in the PR body, which I rewrote at the same time; I have also corrected the PR title, which still described the withdrawn approach.

Re-review should be cheap — the whole change is three source files plus tests, and the one that matters is `TestInjectContext_DelegatedLeadReadsTeamWithoutDisplacingOutputs`, which drives `injectContext` through the guard the first attempt tripped over.

I would rather this went in on a review of the code that is actually here.

yatul:
Pushed `2c63ea8c` — one test, no production change.

`resolveDelegatedLeadTeamRead` relies on `l.dataDir` arriving already tenant-scoped from `resolver.go` (`config.TenantDataDir`), which is why it must not apply `TenantLayer` itself. That was carried only by a comment — here and in the sibling branch of `injectContext` — and nothing failed if the layer were added back.

The failure it guards against is a quiet one. The path stays plausible and simply gains a second tenant segment, `tenants/<slug>/tenants/<slug>/teams/<id>`, so the lead is handed a directory its own tasks never write to; reads then come back empty rather than denied, which reads as "no files" rather than as a bug.

I checked the assertion has teeth by reintroducing the layer — the subtest fails with `tenant segment appears 2 times`. It is prompted by #1546 in this same area, where a contract that lived outside the unit under test went unpinned and the feature shipped green while doing nothing in production.

While you are here: the approval on this PR is still recorded against `019cea10`, the approach I withdrew, and the head is now `2c63ea8c`. The review text describes populating `RunRequest.TeamWorkspace`, 400+ bundled `webui/dist` files, and #1530 as already merged — none of which match what is on the branch. Details are in my comment of 4 September. A fresh look at the current code would be welcome before this is merged.

yatul:
Rebased onto current `dev` on top of the rebased #1530 (no conflicts, same two commits). The red `go` job was `TestHandleTeammateMessageSchedulesStreamedRun` panicking with `close of closed channel` — the race fixed in #1556, now included; `go test ./cmd/ ./internal/tools/ ./internal/agent/` passes locally.

clark-cant (approved):
## Re-review: delegated lead team workspacenn**Verdict: Approve.** Re-reviewed at the current head `6ce4f03322bfa95be101e8e4e828a44820034be4`: the prior approval was stale, but the branch is now CLEAN and all CI checks pass. The stacked prerequisite #1530 is already merged.nnThe implementation preserves artifact-run hermeticity and the active exchange workspace, grants only team-root read access to an unambiguous active lead team, and carries origin routing through task creation, dispatch, reads, and events. Regression coverage addresses multi-team ambiguity, read-vs-write boundaries, delegation artifact safety, and notification origin isolation.nn*Posted by github-maintain cron-safe automation*

PR #1540, opened 2026-08-29
fix(providers): carry reasoning_content on assistant turns without thinking

maitanloi:
## Problem

DeepSeek rejects a request whose assistant history omits `reasoning_content` once the conversation is in thinking mode:

```
HTTP 400: {"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error"}}
```

A thinking-mode turn that only emits tool calls produces no reasoning of its own, so `Message.Thinking` is empty and `openai_request.go` dropped the wire field. The next request then mixes assistant messages with and without `reasoning_content`, and upstream fails. Only `kimi_coding` had an empty-string fallback.

Observed sequence (agent on `deepseek-v4-pro`, telegram channel):

| # | Span | Result |
|---|---|---|
| 1 | llm_call #1 | OK — returned reasoning + `edit` tool call |
| 2 | tool `edit` | failed, `old_string not found in file` |
| 3 | llm_call #2 | OK — bare `read_file` tool call, **no reasoning** |
| 4 | tool `read_file` | OK |
| 5 | llm_call #3 | **HTTP 400**, whole run failed |

The payload sent on call #3 carried one assistant message with `reasoning_content` and one without.

The first replay always succeeds, so the failure only appears from the third iteration onward — it reads as intermittent.

## Fix

Detect thinking mode from the conversation itself: when any assistant message carries reasoning, every replayed assistant message must carry the field, empty string included.

This is narrower than matching on model name, which would also inject the key for non-thinking models such as `deepseek-chat`, and it generalizes to any provider with the same constraint rather than hardcoding a second provider type next to `kimi_coding`.

## Tests

Two tests in `openai_extra_headers_test.go` rebuild the exact message sequence above and assert the second assistant turn carries `reasoning_content: ""`, plus a negative control for a `deepseek-chat` conversation that never entered thinking mode.

The pre-existing negative control `TestNonKimi_ReasoningContentNotAddedWhenEmpty` passes unchanged — its conversation carries no reasoning anywhere, so the new condition leaves it alone.

## Verification

- `go build ./...`
- `go vet ./internal/providers/`
- `go test ./internal/providers/` — all pass
- Image rebuilt from this branch and running against a live gateway (schema 97)

## Surface parity

- API contract: N/A — no request/response shape change; the field is provider-wire only
- Web UI: N/A — no user-facing surface
- CLI/runtime package: N/A — no command or flag change

### DIFF
--- a/internal/providers/openai_extra_headers_test.go
+++ b/internal/providers/openai_extra_headers_test.go
@@ -201,3 +201,56 @@ func TestNonKimi_ReasoningContentNotAddedWhenEmpty(t *testing.T) {
 		t.Error("non-kimi providers must not inject empty reasoning_content; key should be absent")
 	}
 }
+
+// TestDeepSeek_ReasoningContentPresentOnLaterToolCallTurns reproduces upstream
+// "The `reasoning_content` in the thinking mode must be passed back to the API."
+// — once a DeepSeek turn has emitted reasoning, a later assistant tool-call turn
+// that emitted none must still carry the key, or the next request is rejected.
+// Trace: 01a04c34-a74b-7be7-973e-2cb66ea68f3d.
+func TestDeepSeek_ReasoningContentPresentOnLaterToolCallTurns(t *testing.T) {
+	p := NewOpenAIProvider("deepseek", "sk", "https://api.deepseek.com/v1", "deepseek-v4-pro").
+		WithProviderType("deepseek")
+
+	body := p.buildRequestBody("deepseek-v4-pro", ChatRequest{
+		Messages: []Message{
+			{Role: "user", Content: "thêm khoản thu 300k"},
+			{Role: "assistant", Thinking: "the user wants an income entry", ToolCalls: []ToolCall{{ID: "call_1", Name: "edit", Arguments: map[string]any{}}}},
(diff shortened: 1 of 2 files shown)


clark-cant (approved):
## Maintainer reviewnn**Verdict: Approve.**nnThis is a focused provider-wire compatibility fix for DeepSeek thinking-mode replay. The implementation gates empty `reasoning_content` on observed assistant reasoning in the conversation, retains the existing Kimi-specific behavior, and adds both a reproducer and negative control.nn**Mandatory gates**n- Duplicate / prior implementation: no overlapping open remediation found in bounded metadata review.n- Project standards: existing provider serialization and table-test conventions followed.n- Strategic necessity: clear reliability value; prevents a reproducible HTTP 400 during multi-tool turns.nn**Verification / merge gate**nNo Critical or Important findings in the two-file +70/-2 diff. GitHub currently reports no check rollup and `mergeStateStatus=UNSTABLE`, so an immediate merge is not safe in this run. Re-run or restore CI / mergeability, then merge this reviewed SHA.nn*Posted by github-maintain cron-safe automation — 2026-08-29T10:14:57Z*

Issue #1525, opened 2026-08-24
spawn 的 list/wait/cancel 缺少 session 隔离，存在跨会话泄露和误杀


fanchen777:
问题

全局唯一 SubagentManager（cmd/gateway.go:307）的内存 map 混放所有 agent/用户/会话的子任务，list/wait 只按父 agent key 过滤，不看 session。且 cancel id="all"（subagent_control.go:49）遍历整个 map，跨 agent 全杀。而 OriginSessionKey 在 spawn 时已记录（subagent_spawn.go:89），却没被用于过滤。

风险

chat B 的 list 能看到 chat A 的任务（含任务文本前 60 字符）；wait 会被 A 的任务阻塞
cancel id="all" 误杀全进程所有运行中的子任务
open agent 一个 agent 服务多用户：用户 B 可读到用户 A 的任务文本，跨用户信息泄露

修复建议

list/wait/cancel 按 OriginSessionKey 过滤（字段已存在，改动仅涉及 subagent_control.go 和 subagent_spawn_tool_actions.go）。

clark-cant:
[...]

PR #1546, opened 2026-09-01
fix(delegate): add action=list so a lost delegation ID is recoverable

### LINKED ISSUE #1545: delegate has no way to list your own delegations, so one mistyped UUID makes a completed result permanently unreachable
Sibling of #1527, #1529, #1532 and #1535 — same area, but this one is about recovery rather than a broken path.

### Summary

`delegate` persists every delegation durably, but the only way to read a result back is `action: "get"` with the exact delegation UUID. There is no way to list your own delegations. The UUID reaches the caller once, as text in a tool result, and the model has to carry it forward by hand. When it mistypes one character the result becomes permanently unreachable — even though it is sitting in the database, complete.

### Evidence

One production session, a lead delegating to another agent. Two distinct corruption modes, both from the calling model transcribing the id:

```
created  b1bc9f93-3169-4680-92ce-d64c6e6a7d0a
get      b1bc9f93-3169-4680-92ce-d64c6e6a3d0a   -> "delegation result not found"   (x3)
```

One character: `7d0a` became `3d0a`. And earlier in the same session:

```
created  8087e514-904e-4c37-8c18-30d5f19c0b9b
created  3fd46117-8339-45c3-9a3c-8ddf99c83ea1
get      3fd46117-8339-45c3-9a3c-8ddf99c0b9b   -> "delegation_id must be a valid UUID"
```

The tail of the *first* delegation spliced onto the prefix of the second.

After three failed `get` calls the loop guard fires ("called 3 times with the same arguments and identical results"), the agent gives up and spawns a subagent to try to recover the work by itself. From the user's side the delegation simply never answers.

The delegated work had in fact completed and written its deliverable to the team workspace, where it still sits. Nothing was lost except the handle.

### Why this is not only "use a stronger model"

It is fair to say a stronger model would mistype less. Two things make it a design gap regardless:

**The caller is structurally the cheap one.** Only the delegating side ever transcribes the id; the delegatee never does. The delegating side is the user-facing agent carrying the whole conversation — exactly the one an operator runs on a smaller model for cost. So the transcription burden lands where transcription is least reliable, by construction.

**The sibling mechanism already has the affordance.** `spawn` — the other async child-run mechanism, with the same durable tracking — supports `list`, `wait` and `cancel` (`internal/tools/subagent_spawn_tool.go:90-96`). `delegate` supports only `delegate` and `get` (`internal/tools/delegate_tool.go:229`). There is no stated reason for the asymmetry, and it is the difference between a recoverable slip and an unrecoverable one.

### Suggested direction

Add `action: "list"` to `delegate`, returning the caller's own delegations with id, target agent, status and creation time, newest first — enough to re-acquire a handle and call `get`.

**Please do not model it on `spawn`'s current implementation.** #1525 (P1-high, `area:security`) reports precisely that `spawn`'s `list`/`wait`/`cancel` filter only by parent agent key and ignore session, so one chat can enumerate another chat's tasks, and on an open agent one user can read another user's task text. A `list` for delegations must be scoped from the start:

- tenant, as `get` already is;
- the calling agent, as `get` already is;
- **and the origin session**, which `get` does not check today.

The field is already there: `createDelegateCompletion` persists `SessionKey` on the task (`internal/tools/delegate_completion_ledger.go:32-52`), so scoping costs a predicate, not a schema change.

Worth noting while here: `executeGetCompletion` resolves by tenant and calling agent only (`delegate_completion_ledger.go:220`), not by session. That is the same class of leak as #1525, though far weaker in practice — the caller must already know an unguessable UUID. If `list` lands scoped by session, `get` should arguably match it, or `list` hands out ids that `get` then honours across the boundary `list` was careful to draw.

I am happy to implement this once you say whether you want the session predicate on `get` as well, since that is a behaviour change to an existing action rather than a new one.

### Environment

- `v3.15.0-beta.203` plus local patches for #1529/#1532/#1535; the code paths above are unchanged on `dev`.
- PostgreSQL store, Kubernetes. Calling agent on a 31B-class model; the delegatee on a larger one.

yatul:
Fixes #1545. Adds `action: "list"` to `delegate`, scoped to the chat the delegation was raised in.

Not stacked on anything — this applies to `dev` on its own.

### The problem

A delegation result is addressable only by the UUID returned once, in a tool result, which the calling model then has to carry forward by hand. One mistyped character orphans a completed, durably stored result: `get` answers `delegation result not found`, and there is nothing else to ask. Two real failures from one production session, same caller:

```
created  b1bc9f93-3169-4680-92ce-d64c6e6a7d0a
get      b1bc9f93-3169-4680-92ce-d64c6e6a3d0a   -> delegation result not found   (x3)
```

and, earlier, the tail of the previous delegation spliced onto the next one's prefix. After three failures the loop guard fires and the agent gives up. The delegated work had completed and written its deliverable; only the handle was lost.

`spawn` — the sibling async mechanism over the same table — has had `list`/`wait`/`cancel` all along. `delegate` had `delegate`/`get`.

### Scope: tenant + calling agent + origin chat

Tenant and agent as `get` already resolves; the third predicate is `OriginChatID`.

**The chat, not the session, because a delegation belongs to the conversation it was raised in.** Work started in a team chat should stay visible in that team chat and not surface in someone's DM with the same agent. That is deliberate: the history should be available where it began, to the people who were there. A session is an implementation detail of one stretch of a conversation, not of the conversation.

The failure mode in the issue then falls out for free. Deferring long work, clearing the session and context, and coming back to ask for status is ordinary use — the clean context is the point of the reset. A `SessionKey` predicate would return an empty list exactly then, which is when the handle is most likely already gone. With the chat, a reset changes the session key and nothing else.

**Not `spawn`'s current scope.** #1525 (P1-high, `area:security`) reports that `spawn`'s `list`/`wait`/`cancel` filter on the parent agent key alone and ignore the session, so one chat enumerates another chat's tasks — and on an open agent, one user reads another user's task text. The chat predicate is what keeps that from being reintroduced here.

`OriginUserID` is deliberately not added on top. In a direct chat the chat ID is already per person. In a group, a member asking after work the group watched being handed out is reading their own conversation's history. Happy to add it if you would rather have it stricter — it is a predicate, not a redesign.

### `get` is unchanged, deliberately

#1525 is an **enumeration** defect: no prior knowledge required, and task text is disclosed. `get` is **access through an unguessable handle** — there is no way to walk the space and nothing is disclosed without one.

A session predicate on `get` would also break fetching a result by an ID kept across a reset or from another chat with the same agent, which is the exact failure this PR exists to remove. `list` draws a boundary around discovery; `get` stays a capability check.

### Implementation notes

- `ToolChatIDFromCtx` is used deliberately, being the same accessor `createDelegateCompletion` writes from. `OriginChatIDFromCtx` prefers the workspace chat ID and would disagree with the stored value on delegated runs.
- No chat in context is a refusal, not a fallback to listing everything.
- No schema change: the origin fields are already persisted. `ListByParent` is filtered in Go behind a cap of 20, which suits handle recovery; a dedicated store predicate is the obvious next step if this ever needs to page.

### Tests

| property | test |
|---|---|
| only this chat's delegations, and spawns sharing the table are excluded | `TestDelegateListIsScopedToItsChat` |
| a session reset does not hide the delegation | `TestDelegateListSurvivesSessionReset` |
| no chat means refuse, without querying the store | `TestDelegateListRefusesWithoutChat` |
| the cap holds | `TestDelegateListStopsAtItsLimit` |

The fake store leaves `ListBySession` embedded and nil, so a refactor back to session scoping panics rather than passing quietly.

`go test ./cmd/... ./internal/tools/ ./internal/agent/... ./internal/gateway/... ./internal/http/...` is green.

### Update, 7 September

The first version of this PR did not work: it read through `ListByParent`, whose SQL excludes delegations, so `list` always returned an empty list. Fixed in `f4288050`, which adds `ListDelegationsByChat` to `SubagentTaskStore` and both implementations — there was no query in the store capable of returning a delegation at all, only `Get` by ID. Details and the live verification are in the comments below.

### DIFF
--- a/cmd/gateway_subagent_announce_queue_test.go
+++ b/cmd/gateway_subagent_announce_queue_test.go
@@ -25,6 +25,9 @@ func (*recordingGatewayTaskStore) Get(context.Context, uuid.UUID, uuid.UUID) (*s
 func (*recordingGatewayTaskStore) UpdateStatus(context.Context, uuid.UUID, uuid.UUID, string, *string, int, int64, int64) error {
 	return nil
 }
+func (*recordingGatewayTaskStore) ListDelegationsByChat(context.Context, uuid.UUID, string) ([]store.SubagentTaskData, error) {
+	return nil, nil
+}
 func (*recordingGatewayTaskStore) ListByParent(context.Context, uuid.UUID, string) ([]store.SubagentTaskData, error) {
 	return nil, nil
 }
--- a/internal/store/pg/subagent_tasks.go
+++ b/internal/store/pg/subagent_tasks.go
@@ -204,6 +204,37 @@ func (s *PGSubagentTaskStore) ListBySession(
 	return collectTasks(rows)
 }
 
+// ListDelegationsByChat returns delegations raised in one chat (tenant-scoped).
+// The inverse of the predicate ListByParent and ListBySession carry: those serve
+// spawn and filter delegations out, so without this there is no way to list a
+// delegation at all — only Get by ID reaches one.
+func (s *PGSubagentTaskStore) ListDelegationsByChat(
+	ctx context.Context, rootAgentID uuid.UUID, chatID string,
+) ([]store.SubagentTaskData, error) {
+	tid, err := requireTenantID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if rootAgentID == uuid.Nil {
+		return nil, store.ErrSubagentRootAgentIDRequired
+	}
+	if chatID == "" {
+		return nil, nil
+	}
+
+	q := fmt.Sprintf(`SELECT %s FROM subagent_tasks
+		WHERE tenant_id = $1 AND root_agent_id = $2 AND origin_chat_id = $3
+		AND COALESCE(metadata->>'completion_kind', 'subagent') = 'delegate'
+		ORDER BY created_at DESC LIMIT 50`, subagentTaskSelectCols)
+	rows, err := s.db.QueryContext(ctx, q, tid, rootAgentID, chatID)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	return collectTasks(rows)
+}
+
 // Archive marks a bounded batch of old terminal tasks as archived.
 func (s *PGSubagentTaskStore) Archive(
 	ctx context.Context, rootAgentID uuid.UUID, olderThan time.Duration, limit int,
--- a/internal/store/sqlitestore/subagent-tasks.go
+++ b/internal/store/sqlitestore/subagent-tasks.go
@@ -221,6 +221,36 @@ func (s *SQLiteSubagentTaskStore) ListBySession(
 	return collectTasks(rows)
 }
 
+// ListDelegationsByChat returns delegations raised in one chat (tenant-scoped).
+// The inverse of the predicate ListByParent and ListBySession carry: those serve
+// spawn and filter delegations out, so without this there is no way to list a
+// delegation at all — only Get by ID reaches one.
+func (s *SQLiteSubagentTaskStore) ListDelegationsByChat(
+	ctx context.Context, rootAgentID uuid.UUID, chatID string,
+) ([]store.SubagentTaskData, error) {
+	tid, err := requireTenantID(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if rootAgentID == uuid.Nil {
+		return nil, store.ErrSubagentRootAgentIDRequired
+	}
+	if chatID == "" {
+		return nil, nil
+	}
+
+	q := fmt.Sprintf(`SELECT %s FROM subagent_tasks
+		WHERE tenant_id = ? AND root_agent_id = ? AND origin_chat_id = ?
+		AND COALESCE(json_extract(metadata, '$.completion_kind'), 'subagent') = 'delegate'
+		ORDER BY created_at DESC LIMIT 50`, subagentTaskSelectCols)
+	rows, err := s.db.QueryContext(ctx, q, tid, rootAgentID, chatID)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	return collectTasks(rows)
+}
+
 // Archive marks a bounded batch of old terminal tasks as archived.
 func (s *SQLiteSubagentTaskStore) Archive(
 	ctx context.Context, rootAgentID uuid.UUID, olderThan time.Duration, limit int,
--- a/internal/store/sqlitestore/subagent-tasks_test.go
+++ b/internal/store/sqlitestore/subagent-tasks_test.go
@@ -584,3 +584,111 @@ func assertSQLiteArchivedCount(
 		t.Fatalf("archived tasks for %s/%s = %d, want %d", tenantID, rootAgentID, got, want)
 	}
 }
+
+// createSQLiteDelegation persists a delegation row: completion_kind=delegate and
+// an origin chat, which is what ListDelegationsByChat selects on.
+func createSQLiteDelegation(
+	t *testing.T,
+	taskStore *SQLiteSubagentTaskStore,
+	ctx context.Context,
+	rootAgentID uuid.UUID,
+	rootAgentKey, chatID string,
+) uuid.UUID {
+	t.Helper()
+
+	id := uuid.Must(uuid.NewV7())
+	chat := chatID
+	session := "session-" + chatID
+	task := &store.SubagentTaskData{
+		RootAgentID:    rootAgentID,
+		ParentAgentKey: rootAgentKey,
+		SessionKey:     &session,
+		OriginChatID:   &chat,
+		Subject:        "Delegate to brain",
+		Description:    "verify delegation listing",
+		Status:         "completed",
+		Depth:          1,
+		Metadata:       map[string]any{"completion_kind": "delegate"},
+	}
+	task.ID = id
+	if err := taskStore.Create(ctx, task); err != nil {
+		t.Fatalf("Create(%s): %v", id, err)
+	}
+	return id
+}
+
+// ListDelegationsByChat is the only listing that returns delegations at all:
+// ListByParent and ListBySession both carry "completion_kind <> 'delegate'" to
+// serve spawn. This pins that complementarity against the real SQL — a unit test
+// over a fake store cannot see it, and the first version of the delegate list
+// action shipped green and returned nothing in production because of exactly
+// that gap.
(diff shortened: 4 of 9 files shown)


clark-cant (approved):
## Summary

Adds `action: "list"` to the `delegate` tool, scoped to tenant + agent + origin chat, so a caller that lost a delegation UUID can recover it instead of orphaning a completed result.

**Risk level**: Low — small, well-scoped change with 4 tests pinning the contract.

## Mandatory Gates

- **Duplicate / prior implementation**: clear — no prior PR or merged work implements this.
- **Project standards**: docs found — follows established `executeGetCompletion` patterns, consistent error handling, proper context propagation.
- **Strategic necessity**: clear value — directly fixes #1545 (confirmed P2 bug). Real production failure where a mistyped UUID orphans durably stored work with no recovery path.

## Findings

No critical or important issues. The design is sound:

- Chat scoping (not session) is the right call — survives session resets, keeps chats apart (addresses #1525's enumeration concern), and keeps history where it began.
- `ToolChatIDFromCtx` is correctly chosen over `OriginChatIDFromCtx` to match what `createDelegateCompletion` writes.
- No-chat refusal (not fallback) is correct — listing everything would cross the boundary this predicate draws.
- Cap of 20 suits handle recovery; `ListByParent` reuse is pragmatic.
- Tests are excellent — the nil embedded `SubagentTaskStore` trick to catch regressions back to session scoping is particularly sharp.

**Verdict**: Approve.

*Posted by /ck:review-pr at 2026-09-01T14:42:00Z*

clark-cant (approved):
## Review Summary

**Verdict: Approve** — clean, well-scoped fix for a real production problem.

### Mandatory Gates

- **Duplicate/prior implementation**: Clear — no prior `list` action exists for delegate; spawn has it but delegate did not.
- **Project standards**: Follows existing `executeGetCompletion` patterns, proper Go conventions, good error handling.
- **Strategic necessity**: Clear value — lost delegation UUIDs orphan completed work with no recovery path (#1545).

### Analysis

The design choice to scope by **origin chat** (not session) is correct and well-argued:
- Survives session resets (the exact failure case)
- Keeps chats apart (avoids reintroducing #1525's enumeration defect)
- Group chats see their own history, which is appropriate

The test suite is excellent:
- `TestDelegateListIsScopedToItsChat` — verifies chat boundary + spawn exclusion
- `TestDelegateListSurvivesSessionReset` — tests the exact failure case
- `TestDelegateListRefusesWithoutChat` — verifies fail-closed behavior (store not queried)
- `TestDelegateListStopsAtItsLimit` — verifies cap

The fake store's embedded `ListBySession` that panics on use is a clever regression guard against accidentally refactoring back to session scoping.

### CI Status

The `go` test failure appears to be a pre-existing issue on `dev` — same failure pattern on PRs #1530 and #1537 from the same period. This PR's changes are isolated to `internal/tools/` and do not introduce the failure.

### Merge Note

Merge is currently blocked by CI. Once the pre-existing `dev` test issue is resolved, this PR should merge cleanly.

---
*Posted by github-maintain automation at $(date -u +%Y-%m-%dT%H:%M:%SZ)*

yatul:
Please do not merge on the two approvals above — the version they reviewed did not work. Pushed a fix; the PR is now 9 files instead of 3.

### What was wrong

`action: "list"` always returned an empty list. It read through `SubagentTaskStore.ListByParent`, whose SQL carries

```sql
AND COALESCE(metadata->>'completion_kind', 'subagent') <> 'delegate'
```

`ListBySession` carries the same clause. Both serve `spawn` and filter delegations out deliberately, so **no listing in the store could return a delegation** — only `Get` by ID reaches one. The feature was a no-op from the first commit.

That also answers the question this PR opened with more completely than I did. `spawn` has `list`/`wait`/`cancel` and `delegate` had nothing not merely because the action was missing from the tool, but because there is no query underneath it. My original note that "a dedicated store predicate is the obvious next step if this ever needs to page" was wrong: it was required from the start.

### How it was found, and why the tests missed it

Deployed to a live cluster and drove it. An async delegation was created; `get` returned it `completed` with its result; `list` alongside reported `{"count":0}`.

The unit tests were green throughout, because the fake store returned whatever rows the fixture supplied and never reproduced the predicate that does the damage. The test mocked away the exact thing that breaks.

### The fix

`ListDelegationsByChat(ctx, rootAgentID, chatID)` added to the interface and to both implementations, with the inverse predicate plus `origin_chat_id`. The tool points at it, and the chat scope and delegate-only selection now live in the query rather than in a Go filter over whatever the store happened to return.

Tests moved to where the contract actually lives:

- **`internal/store/sqlitestore` against real SQL** — pins that the new method returns the chat's delegation, that a `spawn` in the same chat is not one, that another tenant's identically named chat stays invisible, and that `ListByParent`/`ListBySession` still do **not** return delegations. That last assertion is the one that would have caught this, and it documents the complementarity instead of leaving it assumed.
- **the tool's fake now panics** if `ListByParent` or `ListBySession` is called, so a regression to either fails loudly rather than quietly listing nothing.

Both store implementations are updated; `go test ./cmd/... ./internal/tools/ ./internal/agent/... ./internal/gateway/... ./internal/http/... ./internal/store/...` is green, as is `-tags sqlite ./internal/store/sqlitestore/`.

### Verified running

Same probe as the one that exposed the bug, on the fixed build:

```json
{"count":8,"delegations":[
  {"agent":"brain","status":"running","created_at":"2026-09-07T16:27:42Z", ...},
  {"agent":"brain","status":"completed","created_at":"2026-09-07T16:11:05Z", ...},
  {"agent":"brain","status":"completed","created_at":"2026-08-27T08:11:54Z", ...},
  ...
]}
```

Worth noting what those older rows demonstrate: they were created on 25–27 August, in sessions that no longer exist, and `list` still returns them. That is the argument for scoping by origin chat rather than session, shown in production rather than asserted in a test.

One behavioural note for the record: only async delegations appear. Sync mode returns its result inline and writes no ledger row, so there is no handle to lose and nothing to list — consistent with `get`, whose parameter is already documented as the ID returned by async mode.

clark-cant (approved):
## Re-reviewnn**Verdict: Approve.** The author corrected the prior no-op implementation by adding `ListDelegationsByChat` in both PostgreSQL and SQLite stores, with tenant + root-agent + origin-chat scoping and a delegate-only predicate in SQL.nnThe focused SQL-backed tests now cover the originally missed contract: the new listing returns only this chat's delegation, excludes spawns and same-named chats in other tenants, while the spawn-oriented listings keep excluding delegations. Tool tests also preserve fail-closed no-chat behavior, session-reset recovery, and the 20-item cap.nnRequired CI is green and `mergeStateStatus=CLEAN`; the linked P2 bug #1545 has clear value and no duplicate implementation was found.nn*Posted by github-maintain cron-safe automation*

PR #1547, opened 2026-09-01
fix(tools): use_skill inlines content when read_file isn't granted

### LINKED ISSUE #1477: use_skill only returns activation notice — tool-less agents (no fs group) cannot read skill body/references and loop
For a locked-down Q&A agent (`tools_config: {"profile": "minimal", "allow": ["skill_search", "use_skill"], "deny": ["exec", "group:fs", "group:runtime", "group:sessions", "spawn"]}`), granted skills are effectively unusable:

- `use_skill` returns an activation notice (`skill.activated` logged) but not the SKILL.md body.
- The agent has no read/fs tool, so it cannot follow up by reading the skill files.
- Result: the model re-calls `use_skill` with identical args until the loop guard kills the run:

```
WARN msg="tool loop warning" tool=use_skill message="use_skill has been called 4 times with the same arguments and identical results..."
WARN msg="tool loop critical" tool=use_skill message="CRITICAL: use_skill has been called 5 times..."
ERROR ... content_preview="I keep getting the same activation message. Let me try reading the skill file directly."
```

Request: when the calling agent lacks filesystem tools (or always), `use_skill` should inline the SKILL.md body (and ideally let the agent fetch a specific bundled reference file) in the tool result, so no-tool agents can consume skills. Docs describe progressive disclosure ("SKILL.md body loaded when the skill triggers"), but that load path seems to assume fs access.

Workaround we're using: putting the whole knowledge base in the agent's SOUL.md context file (subject to the 24k total context-file budget).

Observed on v3.14.x self-hosted.

isaacgao4396:
Fixes #1477

## Problem
`use_skill` is a deliberate no-op that just prints an activation message,
relying on the caller to follow up with `read_file` to get the actual skill
body. For agents without `read_file` granted, that follow-up is impossible,
so they loop.

## Fix
`UseSkillTool` now inlines the skill body directly when `read_file` isn't in
the agent's resolved tool set for the current iteration (surfaced via a new
context helper in `internal/store/context.go`, following the existing
`WithAgentID`-style pattern). Agents that do have `read_file` see the exact
same two-step behavior as before — unchanged, byte for byte.

## How this was verified
- `go build ./...` and `go build -tags sqliteonly ./...`
- `go vet` on the touched packages
- `go test -race` on the touched packages
- New unit tests in `internal/tools/use_skill_test.go` covering the inline
  path, the unchanged-behavior regression path, and the nil-allowlist edge
  case
- `test-integration`, `test-invariants`, `test-contracts`, `test-scenarios`
  all pass locally against a `pgvector/pgvector:pg18` container
- Full `go test ./...` shows no new failures introduced by this change

### DIFF
--- a/cmd/gateway_setup.go
+++ b/cmd/gateway_setup.go
@@ -579,7 +579,7 @@ func setupSkillsSystem(
 	skillsLoader := skills.NewLoader(workspace, globalSkillsDir, builtinSkillsDir)
 	skillSearchTool := tools.NewSkillSearchTool(skillsLoader)
 	toolsReg.Register(skillSearchTool)
-	toolsReg.Register(tools.NewUseSkillTool())
+	toolsReg.Register(tools.NewUseSkillTool(skillsLoader))
 	slog.Info("skill_search tool registered", "skills", len(skillsLoader.ListSkills(context.Background())))
 
 	// Wire skills-store directory into filesystem loader so agents
--- a/internal/agent/usage_events_test.go
+++ b/internal/agent/usage_events_test.go
@@ -100,7 +100,7 @@ func TestRecordToolUsageEvent_RuntimeAliasUsesCanonicalName(t *testing.T) {
 func TestRecordToolUsageEvent_UseSkillCountsSkillName(t *testing.T) {
 	storeSpy := newFakeUsageEventStore()
 	registry := tools.NewRegistry()
-	registry.Register(tools.NewUseSkillTool())
+	registry.Register(tools.NewUseSkillTool(nil)) // Execute() not exercised by this test — nil loader is safe
 	loop := &Loop{registry: registry, usageEvents: storeSpy, agentUUID: uuid.New(), tenantID: uuid.New()}
 	ctx := tracing.WithTraceID(store.WithTenantID(t.Context(), loop.tenantID), uuid.New())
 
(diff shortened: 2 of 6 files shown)


clark-cant (approved):
## Summary

Clean fix for #1477: `use_skill` now inlines skill content when `read_file` isn't in the agent's resolved tool set, preventing the loop that tool-less agents hit.

## Risk level: Low

Focused change (+179/-7 across 6 files), well-tested, follows existing patterns.

## Mandatory gates

- **Duplicate / prior implementation:** clear — no prior fix for #1477, no overlapping PRs
- **Project standards:** follows existing context helper pattern (`WithAgentID`-style in `internal/store/context.go`)
- **Strategic necessity:** clear value — fixes confirmed bug where tool-less agents loop indefinitely

## Code review

**Architecture:** Clean separation. The `AvailableToolNamesKey` context helper in `internal/store/context.go` follows the established pattern and is well-documented (nil = "no restriction known", not "no tools available"). Injection in `tool_stage.go` covers both sequential and parallel dispatch paths.

**Correctness:** The inline path correctly checks `available != nil && !available["read_file"]` — only inlines when we can positively confirm `read_file` is missing. The nil-allowlist case preserves original two-step behavior (regression-tested).

**Testing:** 5 tests covering all paths:
- Inline when `read_file` missing (the fix)
- Regression: unchanged behavior when `read_file` present
- Nil allowlist: unchanged behavior (no false positives)
- Skill not found: proper error result
- Empty name: unchanged pre-existing behavior

**Minor observations (non-blocking):**
- `NewUseSkillTool(nil)` in `usage_events_test.go` is safe because `Execute()` isn't exercised there — documented inline
- The loader dependency is minimal and well-scoped

## Verdict: Approve

No critical or important issues. Ready to merge once CI passes.

PR #1548, opened 2026-09-01
fix(anthropic): stack overflow guard + adaptive thinking format + Fable 5 support

JFernandoAmorim2005:
## Summary

3 fixes/additions to the Anthropic provider, developed against our internal fork while running goclaw as internal AI agent infrastructure (self-hosted, non-commercial-redistribution use per license clarification with @nextlevelbuilder — thanks for the quick response!).

- **Bug fix**: `ResolveForwardCompat` could parse a datestamp (e.g. `20260501`) as a minor version when a model ID lacks the `-N-` separator between major version and date (e.g. `claude-opus-4-20260501`), decrementing it in an unbounded recursive loop and crashing the gateway with a stack overflow. Guarded — real minor versions are always < 100.
- **Bug fix**: two related bugs in the `content_block_stop` handler for `RawAssistantContent` thinking-block passback — `ThinkingSignature` was only set after the stream ended (so tool-use passback always omitted it, causing a 400 on the next turn), and `stripThinking=true` produced an empty `thinking` field (also a 400). Both fixed by accumulating per-block signature/content directly.
- **Feature**: Opus 4.6/4.7/4.8, Sonnet 4.6, and the Fable family use the newer "adaptive" extended-thinking format (`thinking:{type:adaptive}` + `output_config.effort`) rather than the older `{type:enabled,budget_tokens}` format, which returns HTTP 400 on these models. Added `claude-fable-5` to the model registry.
- Small fix: removed `TodoRead`/`NotebookRead` from `--disallowedTools` in the Claude CLI provider — these tools don't exist in the current Claude CLI and caused the headless invocation to abort.

All changes are scoped to `internal/providers/` (Anthropic + Claude CLI providers only). Verified against current `main` — the stack overflow is independently reproducible on a fresh checkout (regression test included, reliably crashes within seconds without the guard).

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/providers/...` — all existing tests pass, no regressions
- [x] New regression test for the stack overflow (crashes without the fix, passes with it)
- [x] New tests for adaptive-thinking format selection (Opus 4.6/4.8, Sonnet 4.6, Fable 5) and legacy-format fallback (Sonnet 4.5)
- [x] Extended existing `SkipsTemperatureForClaude46` test to cover `claude-fable-5`


clark-cant (changes requested):
## Maintainer reviewnn**Verdict: request changes.** The stated Anthropic-only fix is reasonable, but this PR is not scoped to that work: it changes 49 files (+3050/-166) across gateway cron/chat wiring, compaction, Feishu, Codex, MCP credential storage, vault enrichment, tracing receipt export, and UI. Those unrelated changes cannot be safely reviewed or merged as part of an Anthropic provider fix.nnPlease split this into a focused PR containing only the Anthropic/Claude CLI changes and their directly related tests: forward-compat guard, adaptive-thinking selection, stream thinking-signature handling, Fable registry entry, and obsolete Claude CLI disallowed-tool cleanup. Submit the other features and refactors as separate PRs with their own scopes and verification.nnThe current branch is also `DIRTY` with no check rollup, so it is not merge-ready.nn*Posted by github-maintain cron-safe automation*

PR #969, opened 2026-04-20
feat(packages): Phase 2b — apk update flow + pkg-helper v2 protocol (#900)

mrgoonie:
## Summary

- Alpine `apk` update flow (checker + executor routed via privileged pkg-helper IPC)
- **BREAKING**: pkg-helper Unix socket protocol v2 — adds `code`/`data` response fields + 3 new actions (`upgrade`, `update-index`, `list-outdated`)
- Edition gating: `SupportsApk` preset (Standard=true, Lite=false) + runtime `/etc/alpine-release` check (`IsAlpineRuntime` with sync.Once cache)
- 3-branch gateway wiring (Alpine+SupportsApk → register / non-Alpine+SupportsApk → SetAvailability("apk",false) / Lite → skip)
- Frontend: source pill, i18n EN/VI/ZH, bugfix in `update-all-modal.tsx` (hardcoded `github:` → dynamic `${source}:`)

## Breaking change

pkg-helper v2 wire protocol. Atomic rebuild at container/desktop boundary — no two-operator rolling upgrade supported. Gateway + pkg-helper must bump together.

## Red-team catches (pre-code)

4 blocking issues fixed in plan before Phase 1 started:
- C-1: executor self-deadlock (non-reentrant `PackageLocker` chan) → documented NO-lock contract
- C-2: no valid availability-map host → added public `UpdateRegistry.SetAvailability`
- H-1: deadline removal → renewable 10-min deadline (set+renew per Scan)
- H-2: zero-value edition field → explicit `SupportsApk`/`SupportsPipNpm` in both presets

## Validation

- **Tester**: 97/97 passing (37.9s, no race warnings, no hangs)
- **Code-reviewer**: APPROVE (0 critical/high/medium, 3 low cosmetic)
- Both builds clean: `go build ./...` + `go build -tags sqliteonly ./...`

## Stacking

Stacked on #944 (Phase 2a pip+npm). Merge order: main ← #944 ← this PR. After #944 merges, rebase onto `feat/packages-update-flow`.

## Test plan

- [x] Unit tests (apk_update_checker/executor + pkg_update_helpers + edition + update_registry)
- [x] Integration tests (`tests/integration/packages_apk_test.go`, `apk_e2e` build tag) — 7 tests incl. arg injection, concurrent install/upgrade, helper unavailable
- [x] Drift-guards: `TestSetAvailability_ExportedWrapper`, `TestSupportsApk`, `TestEditionPresets_ApkField`
- [x] Frontend lints clean
[...]

PR #1549, opened 2026-09-03
fix(skills): retry pip installs without --break-system-packages when unsupported

### LINKED ISSUE #956: Bug: skill dependency checks/install break on macOS because GoClaw resolves the wrong Python/pip runtime
## Summary

On macOS, GoClaw can report core skills as `Missing deps` even when the required Python packages are already installed in the user's active Homebrew Python environment.

This happens because skill dependency install/check paths rely on plain `python3` / `pip3` resolution from the service runtime PATH, which may differ from the interactive shell PATH.

In practice on macOS:
- interactive shell used Homebrew Python/pip:
  - `/opt/homebrew/bin/python3`
  - `/opt/homebrew/bin/pip3`
- GoClaw service resolved Command Line Tools Python/pip instead:
  - `/usr/bin/python3`
  - `/usr/bin/pip3`

That mismatch causes both false `Missing deps` reports and failed installs.

## Symptoms

Core skills repeatedly showed missing dependencies, for example:
- `docx` → `pip:defusedxml, pip:lxml`
- `pdf` → `pip:pypdf, pip:pdf2image, pip:Pillow, pip:pdfplumber`
- `pptx` → `pip:lxml, pip:defusedxml, pip:Pillow`
- `skill-creator` → `pip:anthropic`
- `xlsx` → `pip:lxml, pip:openpyxl, pip:defusedxml`

Observed install failure:

```text
skills: dep install failed
no such option: --break-system-packages
```

## Root cause

There are two related portability issues:

### 1) Hardcoded `--break-system-packages`

`internal/skills/dep_installer.go` invokes pip install with:

```bash
pip3 install --no-cache-dir --break-system-packages <pkg>
```

That breaks on environments where the resolved `pip3` does not support the flag.

### 2) Installer/checker/runtime do not share one Python/pip resolver

Different code paths use plain `python3` / `pip3` from PATH instead of resolving a single canonical runtime.

So GoClaw can end up in this state:
- install/check service path uses `/usr/bin/python3` and `/usr/bin/pip3`
- user manually installs packages into `/opt/homebrew/bin/python3` / `/opt/homebrew/bin/pip3`
- imports succeed in shell but GoClaw still reports `Missing deps`

## Why this matters

This makes skill dependency status unreliable on macOS and any environment with multiple Python installations.

A clean release can still look broken even when the dependencies are already present in the expected user Python environment.

### A. Resolve Python/pip consistently

Introduce shared helpers for runtime selection, e.g.:
- `preferredPython3()`
- `preferredPip3()`

Then use them consistently in:
- dependency installer
- dependency checker
- runtime check
- package lister

On macOS, prefer Homebrew Python first when present:
- `/opt/homebrew/bin/python3`
- `/opt/homebrew/bin/pip3`

with fallback to PATH lookup.

### B. Only add `--break-system-packages` when supported

Before adding the flag, probe:

```bash
pip3 install --help
```

and only include `--break-system-packages` if the selected pip advertises support for it.

## Expected behavior

If the selected runtime can already import the required packages, GoClaw should not report those deps as missing.

If GoClaw needs to install packages, it should use the same Python/pip runtime that the checker uses.

## Environment where reproduced

- macOS Apple Silicon
- GoClaw `v3.9.1` (also observed on `v3.9.0` line)
- service managed by launchd
- Homebrew Python installed
- Command Line Tools Python also present

## Useful evidence

Interactive shell runtime:

```bash
which pip3
/opt/homebrew/bin/pip3
```

Service/runtime PATH ended up resolving `/usr/bin/pip3` / `/usr/bin/python3` in skill paths, while direct imports in Homebrew Python succeeded for:
- `defusedxml`
- `lxml`
- `pypdf`
- `pdf2image`
- `Pillow`
- `pdfplumber`
- `anthropic`
- `openpyxl`

## Operator workaround

There is currently no first-class operator command to compact/fix/recheck a skill session/runtime in a way that resolves this mismatch. On affected systems, the practical workaround is patching GoClaw to use one canonical Python/pip resolver and rebuilding locally.

fcy222fcy:
## Summary

pip >= 23.0 introduced the PEP 668 `--break-system-packages` flag. On older pip builds — e.g. the macOS Command Line Tools Python that GoClaw's managed service resolves on macOS — the flag is rejected outright, and skill dependency installs fail with:

```text
skills: dep install failed
no such option: --break-system-packages
```

This is issue #956.

## Root cause

`--break-system-packages` was appended unconditionally in three pip code paths, and pip `< 23.0` does not know the option at all. Additionally, the pip **`list`** subcommand rejects the flag on **every** pip version (verified against pip 24.0), so `pip3 list --outdated --format json --break-system-packages` failed unconditionally too — silently breaking the pip update checker on all supported environments.

## Changes

- **`internal/skills/dep_installer.go`** — `InstallSingleDep` / `InstallDeps` now route pip installs through a new `pipRunInstall` helper that retries once *without* the flag when pip reports `no such option: --break-system-packages`.
- **`internal/skills/pip_update_executor.go`** — the `pip3 install --upgrade` path applies the same one-shot retry.
- **`internal/skills/pip_update_checker.go`** — dropped `--break-system-packages` from `pip3 list --outdated` entirely: the command is read-only, PEP 668 never applies to it, and no pip accepts the flag there.
- **`internal/skills/pip_flags.go`** (new) — `pipRejectsBSPFlag` / `dropPipBSPFlag` / `pipRunInstall` helpers.
- **`internal/skills/pip_flags_test.go`** (new) — unit tests for the helpers plus integration tests that reproduce the exact legacy-pip failure from #956 and assert the retry sends a flag-free install.

## Verification

- `go test ./internal/skills/` — all pass (incl. pre-existing `TestPipExecutor_*`)
- `go build ./...` — OK
- `go vet ./internal/skills/` — clean
- Bug-first TDD: with the retry disabled, the new legacy-pip tests fail exactly as reported; with the fix they pass.

Fixes #956

### DIFF
--- a/internal/skills/dep_installer.go
+++ b/internal/skills/dep_installer.go
@@ -106,8 +106,9 @@ func InstallSingleDep(ctx context.Context, dep string) (bool, string) {
 			}
 			defer release()
 		}
-		cmd := exec.CommandContext(ctx, "pip3", "install", "--no-cache-dir", "--break-system-packages", pkg)
-		out, err := cmd.CombinedOutput()
+		// pipRunInstall retries without the PEP 668 flag when pip rejects it
+		// (pip < 23.0, see issue #956).
+		out, err := pipRunInstall(ctx, []string{"install", "--no-cache-dir", pipBreakSystemPackagesFlag, pkg})
 		if err != nil {
 			msg := fmt.Sprintf("%s: %v", strings.TrimSpace(string(out)), err)
 			slog.Error("skills: dep install failed", "dep", dep, "error", msg)
@@ -183,8 +184,9 @@ func InstallDeps(ctx context.Context, manifest *SkillManifest, missing []string)
 		slog.Info("skills: installing pip packages", "pkgs", pipPkgs)
 		var successful []string
 		for _, pkg := range pipPkgs {
-			cmd := exec.CommandContext(ctx, "pip3", "install", "--no-cache-dir", "--break-system-packages", pkg)
-			if out, err := cmd.CombinedOutput(); err != nil {
+			// pipRunInstall retries without the PEP 668 flag when pip rejects it.
+			out, err := pipRunInstall(ctx, []string{"install", "--no-cache-dir", pipBreakSystemPackagesFlag, pkg})
+			if err != nil {
 				result.Errors = append(result.Errors, fmt.Sprintf("pip %s: %s (%v)", pkg, strings.TrimSpace(string(out)), err))
 				if hint := pipBuildFailHint(pkg, string(out)); hint != "" {
 					slog.Warn("skills: dep install hint", "pkg", pkg, "hint", hint)
--- a/internal/skills/pip_flags.go
+++ b/internal/skills/pip_flags.go
@@ -1,0 +1,49 @@
+package skills
+
+import (
+	"bytes"
+	"context"
+	"os/exec"
+)
+
+// pipBreakSystemPackagesFlag is the PEP 668 opt-out flag accepted by pip >= 23.0.
+// Older pip builds (and the `list` subcommand on every pip) reject it with
+// "no such option: --break-system-packages", which breaks skill dependency
+// installs on macOS Command Line Tools Python and other legacy environments.
+const pipBreakSystemPackagesFlag = "--break-system-packages"
+
+// pipRejectsBSPFlag reports whether pip's combined output indicates the
+// --break-system-packages flag was rejected as an unknown option. pip < 23.0
+// predates PEP 668 and emits exactly this error (see issue #956).
+func pipRejectsBSPFlag(combinedOut []byte) bool {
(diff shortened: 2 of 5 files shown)


clark-cant (approved):
## Maintainer reviewnn**Summary:** Fixes #956 by retrying pip installs once without `--break-system-packages` only when legacy pip rejects that flag, and removes the invalid flag from read-only `pip list`.nn**Risk:** Low.nn**Mandatory gates:**n- Duplicate/prior implementation: clear; #969 is adjacent package-work, not a duplicate.n- Project standards: aligned with existing skills package patterns and focused regression tests.n- Strategic necessity: clear reliability fix for macOS/legacy pip users.nn**Verdict: Approve.** CI is green and the PR is merge-ready.nn*Posted by github-maintain cron-safe automation at 2026-09-03T06:14:31Z*

PR #1553, opened 2026-09-06
fix(providers): iterate tool-call accumulators by map, not index

aikiata:
## Problem
`OpenAIProvider.ChatStream` parses the accumulated tool-call arguments with a positional loop:

```go
for i := 0; i < len(accumulators); i++ {
    acc := accumulators[i]
    ...
}
```

This assumes every OpenAI-compatible provider emits contiguous, zero-based `tc.Index` values in streamed `delta.tool_calls`. That assumption doesn't hold for all providers — I hit it against a self-hosted OpenAI-compatible gateway that emits non-contiguous/non-zero-based indices. When that happens, `accumulators[i]` misses a populated map key and returns a nil `*toolCallAccumulator`, and the very next line dereferences `acc.rawArgs` on that nil pointer:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x91af07]
goroutine ... [running]:
github.com/nextlevelbuilder/goclaw/internal/providers.(*OpenAIProvider).ChatStream(...)
	/src/internal/providers/openai_chat.go:202 +0x16e7
...
created by .../internal/gateway/methods.(*ChatMethods).dispatchChatSends
```

The panic kills the in-flight chat-run goroutine. It's recovered at a higher level (the gateway logs `tracing: safety-net finalizing orphan trace`), so the process stays up, but the user's in-progress message goes unanswered with no visible error.

## Fix
Range over the `accumulators` map directly instead of indexing by position, and skip nil entries defensively:

```go
for _, acc := range accumulators {
    if acc == nil {
        continue
    }
    ...
}
```

## Verification
- Built a full backend image from this branch (`CGO_ENABLED=0 go build -tags embedui`, same as `make build-full`) — compiles clean.
- Deployed it live against the agent/provider combo that was reliably panicking (multi-tool-call turns: `use_skill` → `read_file` → `exec` x3).
- Before the fix: 2 panics logged within ~10 minutes of runtime, each on the first multi-tool-call turn for that agent.
- After the fix: the same agent/provider ran multiple consecutive multi-tool-call turns end to end, each completing with `v3.run.completed`, zero panics, zero safety-net orphan-trace log lines, restart count stayed at 0.

### DIFF
--- a/internal/providers/openai_chat.go
+++ b/internal/providers/openai_chat.go
@@ -195,9 +195,15 @@ func (p *OpenAIProvider) ChatStream(ctx context.Context, req ChatRequest, onChun
 		return result, fmt.Errorf("%s: stream read error: %w", p.name, err)
 	}
 
-	// Parse accumulated tool call arguments
-	for i := 0; i < len(accumulators); i++ {
-		acc := accumulators[i]
+	// Parse accumulated tool call arguments.
+	// Iterate the map directly (not 0..len(accumulators)-1): providers are not
+	// guaranteed to emit contiguous zero-based tc.Index values in delta.ToolCalls,
+	// so indexing by position can miss populated slots and hit a nil accumulator,
+	// causing a nil-pointer panic (observed with the point-p1/9router provider).
+	for _, acc := range accumulators {
+		if acc == nil {
+			continue
+		}
 		args := make(map[string]any)
 		if err := json.Unmarshal([]byte(acc.rawArgs), &args); err != nil && acc.rawArgs != "" {
 			slog.Warn("openai_stream: failed to parse tool call arguments",


clark-cant (approved):
Approved. Clear, correct fix for a nil-pointer panic in OpenAI streaming tool-call accumulation.

**Root cause:** Positional indexing (`for i := 0; i < len(accumulators); i++`) assumes contiguous zero-based `tc.Index` values from all OpenAI-compatible providers. Providers emitting non-contiguous indices cause `accumulators[i]` to return nil, and the subsequent `acc.rawArgs` dereference panics.

**Fix:** Range over the map directly with nil guard. Minimal, targeted, no behavioral change for conforming providers.

**Verification:** Author confirmed live deployment — 2 panics/10min before fix, zero panics across multiple multi-tool-call turns after fix.

**Risk:** Low — single file, +9/-3, defensive nil check added. No API/protocol changes.

**Note:** CI is not reporting on this branch due to the ongoing go CI infrastructure failure (since 2026-09-01). Approval is based on code review correctness. Merge should proceed once CI infrastructure recovers or maintainer overrides.

clark-cant (approved):
Reviewed and approved. This is a correct fix for a real production panic in `OpenAIProvider.ChatStream` when OpenAI-compatible providers emit non-contiguous or non-zero-based `tc.Index` values in streamed `delta.tool_calls`.

**Assessment:**
- **Correctness:** ✅ Ranging over the `accumulators` map directly instead of positional indexing is the right fix. The nil guard is appropriate defensive coding.
- **Scope:** ✅ Minimal change (+9/-3, 1 file) — no protocol changes, no API surface changes.
- **Verification:** Author confirmed with full backend build + live deployment against the exact agent/provider combo that was panicking. Zero panics after fix across multiple multi-tool-call turns.
- **Risk:** Low. This fixes a nil pointer dereference that was silently dropping user messages (recovered by safety-net but user-visible as unanswered messages).

**Note:** No CI checks have been reported on this PR. Approving based on code review — maintainer should verify CI is green before merge.

PR #1555, opened 2026-09-08
fix(providers): skip temperature for base gpt-5 and gpt-5-chat models

### LINKED ISSUE #1460: OpenAI provider: base "gpt-5" and "gpt-5-chat" send temperature → HTTP 400 from OpenAI-compatible gateways
## Summary

`skipTemp` in `internal/providers/openai_request.go:240` only covers `gpt-5-mini` / `gpt-5-nano` (plus o1/o3/o4). It does **not** cover base **`gpt-5`** or **`gpt-5-chat`**. When an agent runs one of those models through an OpenAI‑compatible provider, GoClaw sends `temperature` (default `0.7` from `config.DefaultTemperature`) and the upstream rejects it:

```
HTTP 400: {"error":{"code":"unsupported_value",
  "message":"Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.",
  "param":"temperature","type":"invalid_request_error"}}
```

Because `inbound: suppressed error for external channel` hides the error on channel integrations (Zalo/Telegram/etc.), the agent goes **silently dead** — no reply, no user-visible error.

## Environment

- Branch: `dev` (build reports `goclaw fork-dev`, protocol 3)
- Provider type: `openai_compat` (reproduced against an OpenAI-compatible gateway serving `gpt-5` family)
- Model configured on the agent: `gpt-5` (also fails with `gpt-5-chat`)

## Repro

1. Create an agent on an `openai_compat` provider with `model = gpt-5` (or `gpt-5-chat`).
2. Send any message.
3. Every turn fails with the HTTP 400 above at `iter 0 think: llm call`.

## Root cause

`internal/providers/openai_request.go:240`
```go
skipTemp := strings.HasPrefix(capabilityModel, "gpt-5-mini") ||
            strings.HasPrefix(capabilityModel, "gpt-5-nano") ||
            strings.HasPrefix(capabilityModel, "o1") ||
            strings.HasPrefix(capabilityModel, "o3") ||
            strings.HasPrefix(capabilityModel, "o4")
```
Base `gpt-5` and `gpt-5-chat` fall through → `temperature` is sent → 400.

## Proposed fix

Gate the whole `gpt-5*` family except the flagship models that *do* accept temperature (matches the existing comment on line 238 about gpt-5.1 / gpt-5.4 / gpt-5.5):

```go
isGPT5Flagship := strings.HasPrefix(capabilityModel, "gpt-5.1") ||
                  strings.HasPrefix(capabilityModel, "gpt-5.4") ||
                  strings.HasPrefix(capabilityModel, "gpt-5.5")
skipTemp := (strings.HasPrefix(capabilityModel, "gpt-5") && !isGPT5Flagship) ||
            strings.HasPrefix(capabilityModel, "o1") ||
            strings.HasPrefix(capabilityModel, "o3") ||
            strings.HasPrefix(capabilityModel, "o4")
```

## Optional hardening

Consider a per-agent `temperature` override (UI field) so operators can pin `temperature = 1` for any model without a code change — there is currently no way to override `config.DefaultTemperature = 0.7` per agent.

fcy222fcy:
## What

`skipTemp` in `buildRequestBody` only covered `gpt-5-mini` / `gpt-5-nano` and the o-series, so base **`gpt-5`** and **`gpt-5-chat`** fell through and received the default `temperature: 0.7`, which OpenAI-compatible gateways reject:

HTTP 400: Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.

Because the inbound error is suppressed on channel integrations, the agent goes silently dead with no user-visible error.

## Change

Extend the locked-temperature gate in `internal/providers/openai_request.go`:

- exact match `gpt-5`
- prefix match `gpt-5-chat` (covers `gpt-5-chat-latest`)

The gpt-5.X flagship variants (`gpt-5.1`, `gpt-5.4`, `gpt-5.5`) keep their temperature — only the base `gpt-5` and `gpt-5-chat` aliases are locked to the default, per the maintainer triage note on the issue.

## Tests

- Moved `gpt-5` from the kept list to the skipped list in `TestBuildRequestBody_TemperatureKeptForNonReasoningModels` (test expectation updated per triage).
- Added regression cases: `gpt-5`, `gpt-5-chat`, `gpt-5-chat-latest`, `openai/gpt-5`, `openai/gpt-5-chat-latest` in `TestBuildRequestBody_TemperatureSkippedForReasoningModels`.
- Verified bug-first: new cases fail against `dev` without the fix, pass with it.
- Full package: `go test ./internal/providers/` green, `go build ./...` and `go vet ./internal/providers/` clean.

Fixes #1460

### DIFF
--- a/internal/providers/openai_request.go
+++ b/internal/providers/openai_request.go
@@ -265,8 +265,14 @@ func (p *OpenAIProvider) buildRequestBody(model string, req ChatRequest, stream 
 		// Certain model families don't support custom temperature (locked to default).
 		// This is a model-level constraint, not provider-specific — applies to both OpenAI and Azure.
 		// Note: gpt-5.X flagship models (gpt-5.1, gpt-5.4, gpt-5.5) DO support temperature;
-		// only the mini/nano reasoning variants reject it.
-		skipTemp := strings.HasPrefix(capabilityModel, "gpt-5-mini") || strings.HasPrefix(capabilityModel, "gpt-5-nano") || strings.HasPrefix(capabilityModel, "o1") || strings.HasPrefix(capabilityModel, "o3") || strings.HasPrefix(capabilityModel, "o4")
+		// the locked ones are the base gpt-5, gpt-5-chat, and the mini/nano reasoning variants.
+		skipTemp := capabilityModel == "gpt-5" ||
+			strings.HasPrefix(capabilityModel, "gpt-5-chat") ||
+			strings.HasPrefix(capabilityModel, "gpt-5-mini") ||
+			strings.HasPrefix(capabilityModel, "gpt-5-nano") ||
+			strings.HasPrefix(capabilityModel, "o1") ||
+			strings.HasPrefix(capabilityModel, "o3") ||
+			strings.HasPrefix(capabilityModel, "o4")
 		// Kimi Coding rejects any temperature override — `invalid temperature: only
 		// 1 is allowed for this model`. Skip sending so the upstream applies its
 		// own default (1). Matches the model-locked behavior of o1/o3/o4.
--- a/internal/providers/openai_test.go
+++ b/internal/providers/openai_test.go
@@ -59,6 +59,9 @@ func TestBuildRequestBody_TemperatureSkippedForReasoningModels(t *testing.T) {
 	// These model families don't support custom temperature (locked to default).
 	// This is a model-level constraint, not provider-specific.
 	models := []string{
+		"gpt-5",
+		"gpt-5-chat",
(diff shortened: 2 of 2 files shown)


clark-cant (approved):
Approved: this narrowly fixes #1460 by omitting temperature for exact gpt-5 and gpt-5-chat aliases while retaining explicit coverage for supported gpt-5.x variants. Mandatory duplicate/prior-art and strategic-value gates are clear; all CI checks are green.

PR #1562, opened 2026-09-10
feat(image): add GPT Image 2.5 and default to flare

thotam:
## Summary

Adds GPT Image 2.5 (released by OpenAI on 2026-09-08) to the `create_image` tool and makes `gpt-image-2.5-flare` the default image model, replacing `gpt-image-2`. GPT Image 2.5 ships as **two** model IDs rather than one, so the dropdown now offers four choices.

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev`.

## Details

**Why flare is the default:** OpenAI reports flare delivers higher quality than `gpt-image-2` at roughly half the latency, which suits `create_image`'s one-shot call. Pick `sunburst` when editing precision matters most. Both accept text and image input and run on `/images/generations`, `/images/edits` and the Responses API, which is the path the native `image_generation` tool already uses.

**No breakage for existing configs:** `gpt-image-2` and `gpt-image-1.5` stay on the whitelist. Anyone with `image_model: "gpt-image-2"` saved in a provider chain keeps the exact behavior they have; only users who never picked a model move to flare.

**Backend** (`internal/providers/native_image.go`)
- `DefaultImageModel` becomes `gpt-image-2.5-flare`; the whitelist grows to four models.
- The rejection message is now built from the whitelist itself instead of a hand-written string that drifts whenever a model is added.
- `codex_build.go` uses the `DefaultImageModel` constant instead of hardcoding `"gpt-image-2"`, so the inline chat path cannot drift from the shared default.
- `create_image.go` adds both new models to the `isEditModel` branch so reference images route to `/images/edits`.

**Web UI** (`ui/web/src/pages/builtin-tools/`)
- The Image model dropdown lists four options, defaulting to flare.
- `ParamField` gains an optional `labelKey`, letting a field read its label from i18n instead of showing English inside a translated UI. It passes a `defaultValue`, so a missing key degrades to the literal rather than breaking, and fields without a `labelKey` behave exactly as before.

**i18n** - keys added to all five catalogs (`en/vi/zh/ko/ru`). `ko` was previously missing the entire `mediaChain.imageModel*` group.

## Cross-Surface Parity

- **Gateway server:** changed (whitelist, default, edit routing).
- **API contract:** N/A, no struct change. `image_model` lives in the media chain's free-form `params` map; only the backend whitelist widened.
- **Web UI:** changed (dropdown + i18n).
- **Desktop UI:** N/A, `ui/desktop/frontend/src/components/tools/sortable-provider-card.tsx` renders no param form and passes `params` through as `Record<string, unknown>`. The new default still reaches desktop via `DefaultImageModel`.
- **CLI/runtime package:** N/A, `cmd/` exposes no image-model flag or command.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] SQL queries use parameterized `$1, $2` (no string concat) - N/A, no SQL touched
- [x] New user-facing strings added to all 3 locales (en/vi/zh) - added to all five
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration) - N/A, no migration

## Test Plan

**Automated**
- New `internal/providers/native_image_test.go`: seven cases for `ValidateImageModel`, including a case asserting bare `gpt-image-2.5` is rejected because it is not a real model ID, plus a guard test that `DefaultImageModel` always stays inside the whitelist.
- Existing tests that hardcoded `"gpt-image-2"` as the default now reference the `DefaultImageModel` constant, so the next default change needs no test edits.
- `go test -race ./...`, `pnpm test` (355 tests / 55 files), `pnpm lint` (0 errors).

**Manual**
Ran the gateway against SQLite with `-tags sqlite`, opened Dashboard → Built-in Tools → Create Image → Provider Chain and selected a ChatGPT OAuth provider. Confirmed the dropdown lists four models with flare selected by default, and that switching the UI language translates the labels. Generated a real image through this path to confirm the new model works end to end.

### DIFF
--- a/internal/providers/codex_build.go
+++ b/internal/providers/codex_build.go
@@ -118,7 +118,7 @@ func (p *CodexProvider) buildRequestBody(req ChatRequest, stream bool) map[strin
 				tools = append(tools, map[string]any{
 					"type":           "image_generation",
 					"action":         "generate",
-					"model":          "gpt-image-2",
+					"model":          DefaultImageModel,
 					"output_format":  "png",
 					"partial_images": 1,
 				})
--- a/internal/providers/codex_native_image_test.go
+++ b/internal/providers/codex_native_image_test.go
@@ -53,7 +53,7 @@ func mockImageServer(t *testing.T, captured *[]byte) *httptest.Server {
 // asserts each required field is present and well-formed.
 //
 // Sub-cases:
-//   - Default (empty ImageModel) → tools[0].model == "gpt-image-2"
+//   - Default (empty ImageModel) → tools[0].model == DefaultImageModel
 //   - Legacy (ImageModel: "gpt-image-1.5") → tools[0].model == "gpt-image-1.5"
 //   - Rejected (ImageModel: "dall-e-3") → GenerateImage returns error containing "unsupported image model"
 func TestCodexGenerateImage_BuildsNativeRequest(t *testing.T) {
@@ -163,7 +163,7 @@ func TestCodexGenerateImage_BuildsNativeRequest(t *testing.T) {
 }
 
 // TestCodexGenerateImage_ImageModelDefault verifies that an empty ImageModel results
-// in the default gpt-image-2 model in the outbound tools[0].model field.
+// in DefaultImageModel in the outbound tools[0].model field.
 func TestCodexGenerateImage_ImageModelDefault(t *testing.T) {
 	var captured []byte
 	server := mockImageServer(t, &captured)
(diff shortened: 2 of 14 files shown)


clark-cant (approved):
## Maintainer reviewnn**Verdict: Approve.**nn- Duplicate/prior implementation: clear (no overlapping PRs or issues found).n- Project standards: compliant with AGENTS.md; backend, web UI, i18n, and test coverage move together.n- Strategic necessity: clear user-facing model availability and consistent default behavior.n- Risk: low; the full CI rollup is green and merge state is CLEAN.nnReviewed model validation/defaulting, native edit routing, UI option parity, all locale keys, and focused regression coverage. No blocking findings.nn*Posted by github-maintain cron-safe automation*

PR #1563, opened 2026-09-11
fix(tools): charge native media a flat unit instead of its base64 bytes

thotam:
## Summary
`read_video` with a `url` always failed under an agent budget, and `read_audio` / `read_document` only worked for very small files, because out-of-band media was priced by base64-encoding the payload and counting it as text at 0.956 tokens per raw byte. Media is now priced by what a provider actually bills for it: measured duration for video and audio, measured page count for PDFs. Revised after @clark-cant's review; details below.

## Type
- [ ] Feature
- [x] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch
`dev`. Rebased onto `39f66b3a`, single commit.

## What changed since the review

The previous revision charged a flat `InlineMediaUnit` per payload. That was rejected as systematic under-counting that removes pre-send enforcement, and the rejection was correct. This revision drops it entirely.

**Every number is now either a published per-unit rate or a published provider limit.** Nothing is derived from a byte size. An intermediate revision of this branch did try a bitrate floor, and it does not survive contact with real media: a 120-second, 20,627-byte clip priced at 526 tokens against a true 31,560, and a one-hour static-image H.264 file fits in 58,999 bytes. There is no bitrate floor to find.

`internal/mediabudget` prices one payload:

| Input | Source of truth | Rate |
|---|---|---|
| Video | `ffprobe` duration | 263 tokens/second |
| Audio | `ffprobe` duration | 32 tokens/second |
| PDF | `pdfinfo` page count | 258 tokens/page |
| PDF, page count unreadable | provider limit of 1000 pages | 258,000 tokens |

**Video and audio that cannot be measured are refused.** This is not a new policy: `reserveToolLLMUsageUnverifiableMedia` already refused every URL on the Gemini streamed path. This narrows that refusal to what genuinely cannot be measured instead of removing it. PDF differs because its page count is cheaply measurable and its ceiling is small enough to stay usable.

## Measuring a remote video without downloading it

Two ranged GETs, 512 KB from the head and 512 KB from the tail, written into a sparse temp file sized to the declared total and handed to `ffprobe`.

The tail is not optional. Head-only probing under-reports `mpeg` by 98% and `ogg` by 79%, because every container that stores its index at the end keeps it there. And a plain prefix makes `ffprobe` under-report a 30-second WAV as 0.74 seconds, since it clamps to the bytes it can see; sizing the temp file to the real total fixes that.

Verified end to end on an 11,673,105-byte MP4 served by nginx: 1 MB of ranged reads yields `duration=40.000000`, identical to `ffprobe` reading the whole URL, for a charge of 10,520 tokens. The same file cost 11,161,058 tokens before.

Those requests reuse the existing SSRF-safe path, `security.WithPinnedIP` plus `security.NewSafeClient(0)`, redirects are not followed, and no URL is ever handed to an external binary.

## Two pre-existing gaps this also closes

While tracing the reported bug I found large media could already reach a provider almost unpriced, independently of the reported symptom:

1. `ExecuteWithChain` treated every `callProvider` error as a provider failure and advanced to the next chain entry. A budget refusal is an error, so refusing merely routed the call to the next provider.
2. The non-Gemini branches of `read_video` and `read_document` reserved without pricing their payload at all.

Measured under a 20,000-token window before this change, a 40 MB video and a 1000-page PDF each reached a provider charged about 1,600 tokens. Budget refusals are now terminal in the chain, and every media branch prices its payload, so both reach no provider at all. Genuine provider failures still fail over; there is a test for each.

## Known limits, stated rather than left to be found

- The `/Type /Page` scan that guards against a forged `/Count` is a floor, not a bound. Pages inside a compressed object stream are invisible to it, and a 9,484-byte PDF built that way is charged 258 tokens for 1000 pages. A real `pdfinfo` reads such files correctly; the scan only ever raises a probed count, never lowers one.
- A video URL whose origin does not serve byte ranges is now refused on the non-Gemini path too, and that refusal is terminal. Upstream forwarded such URLs unpriced. A `HEAD` giving only a size is not enough to price one.
- `read_video` and `read_audio` require `ffprobe`. Docker images install it by default except the `base` variant; bare binaries and the desktop build do not ship it, and fall back to refusing.
- A hostile origin can craft a container `ffprobe` reads as about one second. `Reservation.Reconcile` overwrites the estimate with the provider's reported usage, so this weakens the gate rather than defeating it.
- `read_image` still reserves with plain `reserveToolLLMUsage`. Images ride inside `ChatRequest.Images` rather than out-of-band, so it is a separate question and out of scope here.
- Making refusals terminal applies to every media chain, including `create_image` and `create_video`, not only the three tools above. A window or tenant-cap denial is never fixed by trying another provider, so this looks right, but it is wider than the reported bug.
- New error strings are plain `fmt.Errorf` like the rest of `internal/tools`, not routed through `internal/i18n`. Happy to convert if you prefer, but it would be inconsistent with the surrounding package.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` (run on the four affected packages: `internal/mediabudget`, `internal/tools`, `internal/usage/caps`, `internal/tokencount`)
- [ ] Web UI builds: `cd ui/web && pnpm build` (no UI changes)
- [x] SQL queries use parameterized `$1, $2` (no SQL touched)
- [x] New user-facing strings added to all 3 locales (en/vi/zh) (see the i18n note above)
- [x] Migration version bumped in `internal/upgrade/version.go` (no migration)

## Test Plan

Regression tests for the case you asked for, a large payload that cannot proceed when the budget cannot accommodate it:

- `TestReadVideo_GeminiURL_RejectsHugeMediaBeforeStreaming` and `TestReadVideo_GeminiURL_RangeRefusingOriginBlockedBeforeStreaming` drive `callProvider` against a live origin and assert the streaming GET never happens.
- `TestReadAudio_RefusedAudioNeverReachesItsProvider` and `TestReadDocument_RefusedPDFNeverReachesItsProvider` do the same through their own tools, asserting zero provider hits.
- `TestReserveToolLLMUsageWithMedia_MediaChargeTripsTenantTokenCap` uses a real `usagecaps.Service` and asserts `ErrCapExceeded` plus a recorded block event, with an A/B control that passes without the media charge.
- `TestExecuteWithChain_UnmeasurableMediaIsTerminal` pins that a budget refusal stops the chain, and a separate case pins that a 500 still fails over.
- `internal/mediabudget` covers both probes, the provider ceilings, the forged `/Count`, a mislabelled PDF, overflow and infinite durations, and the sparse head+tail assembly.

Every test passes with `ffprobe` and `pdfinfo` absent from `PATH`.

Manual: ran the gateway against PostgreSQL 18 with pgvector and confirmed `read_video` with a `url` now completes.

## Surface parity
Gateway server, plus the Dockerfile and the three release workflows for the new `ENABLE_MEDIA_PROBES` build arg, plus README and `docs/14-skills-runtime.md`. API contract N/A: nothing in `pkg/protocol` changes. Web UI N/A. CLI N/A.

### DIFF
--- a/.github/workflows/dev-beta-release.yaml
+++ b/.github/workflows/dev-beta-release.yaml
@@ -289,12 +289,14 @@ jobs:
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "false"
+            enable_media_probes: "true"
           - variant: full
             suffix: "-full"
             enable_otel: "false"
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "true"
+            enable_media_probes: "true"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -348,6 +350,7 @@ jobs:
             ENABLE_EMBEDUI=${{ matrix.enable_embedui }}
             ENABLE_PYTHON=${{ matrix.enable_python }}
             ENABLE_FULL_SKILLS=${{ matrix.enable_full_skills }}
+            ENABLE_MEDIA_PROBES=${{ matrix.enable_media_probes }}
             VERSION=${{ needs.beta_version.outputs.tag }}
           cache-from: type=gha,scope=dev-beta-${{ matrix.variant }}
           cache-to: type=gha,mode=max,scope=dev-beta-${{ matrix.variant }}
--- a/.github/workflows/release-beta.yaml
+++ b/.github/workflows/release-beta.yaml
@@ -98,12 +98,14 @@ jobs:
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "false"
+            enable_media_probes: "true"
           - variant: full
             suffix: "-full"
             enable_otel: "false"
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "true"
+            enable_media_probes: "true"
     steps:
       - uses: actions/checkout@v4
 
@@ -149,6 +151,7 @@ jobs:
             ENABLE_EMBEDUI=${{ matrix.enable_embedui }}
             ENABLE_PYTHON=${{ matrix.enable_python }}
             ENABLE_FULL_SKILLS=${{ matrix.enable_full_skills }}
+            ENABLE_MEDIA_PROBES=${{ matrix.enable_media_probes }}
             VERSION=${{ github.ref_name }}
           cache-from: type=gha,scope=beta-${{ matrix.variant }}
           cache-to: type=gha,mode=max,scope=beta-${{ matrix.variant }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -167,16 +167,19 @@ jobs:
             enable_embedui: "false"
             enable_python: "false"
             enable_full_skills: "false"
+            enable_media_probes: "false"
           - variant: latest
             suffix: ""
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "false"
+            enable_media_probes: "true"
           - variant: full
             suffix: "-full"
             enable_embedui: "true"
             enable_python: "true"
             enable_full_skills: "true"
+            enable_media_probes: "true"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -223,6 +226,7 @@ jobs:
             ENABLE_EMBEDUI=${{ matrix.enable_embedui }}
             ENABLE_PYTHON=${{ matrix.enable_python }}
             ENABLE_FULL_SKILLS=${{ matrix.enable_full_skills }}
+            ENABLE_MEDIA_PROBES=${{ matrix.enable_media_probes }}
             VERSION=v${{ needs.release.outputs.version }}
           cache-from: type=gha,scope=release-${{ matrix.variant }}
           cache-to: type=gha,mode=max,scope=release-${{ matrix.variant }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -74,6 +74,7 @@ ARG ENABLE_PYTHON=false
 ARG ENABLE_NODE=false
 ARG ENABLE_FULL_SKILLS=false
 ARG ENABLE_CLAUDE_CLI=false
+ARG ENABLE_MEDIA_PROBES=true
 
 # Copy pinned Python deps (cleaned up after install).
 # requirements-base.txt: shared deps for ENABLE_PYTHON and ENABLE_FULL_SKILLS.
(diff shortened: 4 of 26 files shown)


clark-cant (changes requested):
## Review — Request changesnn**Important: this replaces a conservative accounting failure with systematic under-counting and breaks the stated complete-input budget invariant.** `read_video` may stream a file up to 2 GB, but this change reserves only `tokencount.InlineMediaUnit` (1,600 tokens) for every out-of-band audio, document, or video payload. The PR description explicitly notes a one-hour video can consume roughly 946k tokens while the guard and tenant reservation both charge 1,600.nnThat means an agent with a small remaining context/usage budget can now send arbitrarily large media to a provider, bypass the preflight cap, and only discover the real charge during reconciliation. This is not merely an estimate-quality trade-off: it removes the pre-send enforcement guarantee that the prior code deliberately maintained. The URL transport regression needs fixing, but the fix must retain a conservative, bounded preflight estimate (for example, media-type/duration/page-aware estimation where metadata is available, plus a safe upper-bound/fail-closed policy where it is not) and must prevent caps from being exceeded before provider transport.nnCI is green and the transport tests are useful, but they do not cover the core cap-bypass case. Please add regression tests proving a large out-of-band video/audio/document cannot proceed when the agent context or tenant usage cap cannot accommodate a conservative estimate.nn*Posted by github-maintain cron-safe automation*

thotam:
@clark-cant thanks for the review, you were right to reject the flat unit. Pushed a rewrite and rebased onto `39f66b3a`, single commit. Summary of what changed and what I could not solve, so you can decide quickly whether it is worth a second look.

**The flat `InlineMediaUnit` is gone.** So is every byte-derived estimate. I tried a bitrate floor first and it does not survive contact with real media: a 120-second, 20,627-byte clip priced at 526 tokens against a true 31,560, and a one-hour static-image H.264 file fits in 58,999 bytes. There is no floor to find, so the branch stopped looking for one.

**Every number now comes from a published rate or a published provider limit.** `ffprobe` measures video and audio duration at 263 and 32 tokens/second; `pdfinfo` counts PDF pages at 258 each. When a PDF's page count cannot be read, the charge is the 1000-page provider ceiling, which is a real bound because the provider refuses anything larger. When video or audio duration cannot be measured, the call is refused: `reserveToolLLMUsageUnverifiableMedia` already refused every URL on the Gemini streamed path, so this narrows that refusal to what genuinely cannot be measured rather than removing it.

**A remote video is measured without downloading it:** two ranged GETs, 512 KB head and 512 KB tail, assembled into a sparse temp file sized to the declared total. The tail is load-bearing, head-only probing under-reports `mpeg` by 98% and `ogg` by 79%. On the reported 11,673,105-byte file this yields `duration=40.000000`, identical to `ffprobe` reading the whole URL, for 10,520 tokens against 11,161,058 before. Same pinned-IP SSRF path as the streaming GET; no URL reaches an external binary.

**Two things I found while tracing this that are not from the reported bug**, and I would rather flag than bury:

1. `ExecuteWithChain` treats any `callProvider` error as a provider failure and advances to the next entry, so a budget refusal simply routed the call to the next provider.
2. The non-Gemini branches of `read_video` and `read_document` reserved without pricing their payload at all.

Together, under a 20,000-token window, a 40 MB video and a 1000-page PDF each reached a provider charged about 1,600 tokens. That is your sentence, reachable on `dev` today and independent of this PR. Refusals are now terminal in the chain and every media branch prices its payload; both now reach no provider. Genuine provider failures still fail over, with a test for each. If you would rather have that as a separate PR, say so and I will split it out.

**What I did not solve.** The `/Type /Page` scan that guards against a forged `/Count` is a floor, not a bound: pages inside a compressed object stream are invisible to it, and I can build a 9,484-byte PDF that is charged 258 tokens for 1000 pages. A real `pdfinfo` reads those files correctly and the scan only ever raises a probed count, but closing it properly means decompressing object streams and walking the page tree, which is a PDF parser for untrusted input and felt like the wrong thing to add here. Also: an origin that does not serve byte ranges is now refused on the non-Gemini path too, which upstream did not refuse; `read_video` and `read_audio` need `ffprobe`, installed by default in every Docker variant except `base`; and `read_image` still reserves unpriced, which I left alone since its payload is not out-of-band. The full list is in the PR description.

Happy to split, trim, or change the refusal policy if you would rather have a different trade-off.

thotam:
Closing in favour of a fresh pull request for the same branch, so the change gets picked up by the usual new-PR review flow rather than sitting as an update to a thread that already has a stale changes-requested state.

Nothing is lost: the replacement carries the same branch and the same commit, and its description restates @clark-cant's objection and answers it point by point, so it can be read without this thread. Link to follow in a moment.

The review here was correct and it is what drove the rewrite, so thank you for it.

thotam:
Replacement is #1564, same branch and same commit.

PR #1496, opened 2026-08-02
feat(tools): add optional Parallel web search provider

georgeatparallel:
## Summary

Adds Parallel as an optional, keyless provider in GoClaw's existing web search chain. It stays disabled unless a tenant enables it, while the current provider defaults and DuckDuckGo fallback remain unchanged.

## Type

- [x] Feature
- [ ] Bug fix
- [ ] Hotfix (targeting `main`)
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev`

## Checklist

- [x] `go build ./...` passes
- [x] `go build -tags sqliteonly ./...` passes (if Go changes)
- [x] `go vet ./...` passes
- [ ] Tests pass: `go test -race ./...`
- [x] Web UI builds: `cd ui/web && pnpm build` (if UI changes)
- [x] No hardcoded secrets or credentials
- [x] SQL queries use parameterized `$1, $2` (no string concat)
- [x] New user-facing strings added to all 3 locales (en/vi/zh)
- [x] Migration version bumped in `internal/upgrade/version.go` (if new migration)

## Test Plan

- `go test ./internal/tools -count=1`
- `go test -race ./internal/tools -count=1`
- `go test ./...`
- `go build ./...`
- `go build -tags sqliteonly ./...`
- `go vet ./...`
- `cd ui/web && pnpm build`
- `cd ui/web && pnpm lint` (passes with four pre-existing warning-only unused directives outside this change)
- `git diff --check`
- Credential-free live JSON-RPC smoke request to `https://search.parallel.ai/mcp` (HTTP 200 with the expected nested results payload)

The full-repository race command was not run; race detection was focused on the changed `internal/tools` package, and the full repository test suite passed without the race detector.

Disclosure: I work at Parallel.ai, which operates this MCP endpoint.

PR #1567, opened 2026-09-17
feat(tools): add Serply as a web_search provider

googio:
## Summary

Adds Serply as a `web_search` provider. Serply is a SERP API that returns Google web results as JSON on a tenant-supplied key, so it slots into the existing chain exactly like Exa, Tavily and Brave: read from `config_secrets`, skipped silently when no key is stored, and selectable through the `provider` parameter for cross-engine corroboration.

Disclosure: I work with Serply. The provider is inert unless a tenant stores their own key, the same as the other keyed entries.

## Type
- [x] Feature
- [ ] Bug fix
- [ ] Refactor
- [ ] Docs
- [ ] CI/CD

## Target Branch

`dev`.

## Checklist
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Tests pass: `go test -race ./...` (88 packages, no failures)
- [x] SQL queries use parameterized `$1, $2` (no string concat) (no SQL in this change)
- [x] New user-facing strings added to all 3 locales (en/vi/zh), plus ru. See the i18n note below.
- [ ] Migration version bumped in `internal/upgrade/version.go` (if new migration) (no schema change)

## What changed

| Area | Change |
|---|---|
| `internal/tools/web_search_serply.go` | New provider. GET, key in `X-Api-Key`, maps `results[]` to title/link/description. |
| `internal/tools/web_search.go` | Endpoint and name constants, entry in `defaultSearchProviderOrder`, `provider` param description. |
| `internal/tools/web_search_config.go` | One `buildProviderByName` case. |
| `internal/http/builtin_tools.go` | One `toolSecretKeys` entry so a key typed in the settings form is extracted into `config_secrets` and stripped from the persisted settings blob, rather than stored inline. |
| `ui/web/src/pages/builtin-tools/web-search-chain-form.tsx` | Fifth sortable provider plus the two touch-ups noted below. |
| `ui/web/src/i18n/locales/*/tools.json` | Provider label. |

Serply is key-gated, so unlike Parallel it belongs in the default order: a tenant with no key is skipped by `BuildChainFromStorage` rather than by being absent from the order. No existing chain changes for anyone who has not stored a Serply key.

## Three judgement calls worth your review

**1. Freshness only maps the shortcuts.** Serply forwards Google's `tbs` recency filter upstream, so `pd/pw/pm/py` become `qdr:d/w/m/y`. `normalizeFreshness` also accepts a `YYYY-MM-DDtoYYYY-MM-DD` range, and I deliberately do not send that. I tested the equivalent `cdr:1,cd_min:...,cd_max:...` form against the live API: a Q1 2023 range came back with results published an hour earlier, so the range is silently ignored upstream. Sending it would name a date filter the results do not honour. A range therefore leaves `tbs` unset, which is still ahead of Exa, Tavily and Parallel, all of which ignore freshness entirely. There is a table test pinning this.

**2. Two small UI touch-ups that the fifth provider forced.** The locked DuckDuckGo card hardcoded `#5`, which becomes wrong the moment a provider is added, so it now derives its position from `entries.length`. And `parseInitialEntries` special-cased reconciling a stored `provider_order` against Parallel alone; it now reconciles against `SORTABLE_PROVIDERS`, so a tenant with a saved order sees any newly added provider instead of only Parallel. Both are the same size as the hardcoded alternative and stop the next provider from hitting the bug. Happy to shrink either to a literal if you would rather keep the diff mechanical.

**3. User-Agent.** The provider sends `User-Agent: goclaw` rather than the browser `webSearchUserAgent` constant, which exists for the scraping path and which Brave and Exa also skip. Serply attributes usage by client, so identifying the gateway is more useful than a spoofed Chrome string.

## i18n

The new string went into `en`, `ru`, `vi` and `zh`. I left `ko` alone on purpose: `ko/tools.json` has no `builtin.searchChain` block at all, so adding a lone `providers.serply` key there would be the only entry in an otherwise absent section, and i18next already falls back to `en`. That matches the Parallel provider PR (#1496), which also touched these four. Say the word if you would rather I seed the whole Korean `searchChain` block as a separate change.

## Surface parity

- Gateway server: covered (provider, chain registration, secret extraction).
- API contract: N/A because no wire type, request or response struct changed. The only contract-visible edit is the `provider` parameter description, which is a free-form string, not an enum.
- Web UI: covered (settings form plus locales).
- CLI/runtime package: N/A because the provider is configured through the settings API, not through a CLI command.
- Migrations: N/A because nothing touches the schema. The `hook_web_search_migrate` path deliberately still lists only exa/tavily/brave, since it migrates keys that once lived inline in `config.json5` and Serply never had one there.

## Size

11 files and +359/-10, which is a little under the Parallel provider PR (#1496, 11 files, +395). Roughly half of it is the test file. I mirrored that PR's shape deliberately rather than inventing a different one.

## Test Plan

Unit tests in `web_search_serply_test.go` run against `httptest` fixtures, so they need no key and no network: request shape (method, `X-Api-Key`, and the `q`/`num`/`gl`/`hl`/`tbs` mapping), response mapping, the freshness table including both no-op cases, the result limit, skipping a result with an empty link, and the 401/503/malformed-body error paths plus context cancellation. Two scenarios were added to `TestBuildChainFromStorage` covering a stored Serply key and its absence.

I also ran the provider against the live API before opening this, in a throwaway test that is not part of the diff:

- plain query returned the expected Go 1.26 release notes as the first hit
- `Country: "DE", SearchLang: "de"` returned German results, confirming the `gl`/`hl` mapping
- `Freshness: "pd"` returned a story published the same day, confirming the `tbs` mapping
- a date range returned unfiltered results, which is the behaviour documented above

Full gate locally: `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./...`, `go test -race ./...` (88 packages ok), `cd ui/web && pnpm build` (includes `tsc -b`), `pnpm test` (355 tests), and `eslint` on the changed component.

One note on `go fix ./...`: running it as CONTRIBUTING asks rewrote 25 files across the repo that have nothing to do with this change, so I reverted all of them and kept the diff to the 11 files above. Worth knowing if you expect that command to be a no-op on a clean tree.


clark-cant (commented):
## Maintainer review — github-maintain automation

**Summary**: Adds Serply as a keyed `web_search` provider, wired into the existing first-success-wins chain (constants, default order, provider factory, admin key allowlist, i18n, UI chain form). Inert unless a tenant stores its own Serply key.

**Risk level**: Low — additive; one new provider file plus registration points. No schema, API-contract, or migration changes. 11 files, +359/-10.

**Mandatory gates**
- Duplicate / prior implementation: clear — repo search for "serply" returns only this PR; no existing provider covers it.
- Project standards: docs found (CLAUDE.md / AGENTS.md); change follows the established per-provider pattern (one file per provider, registered in `web_search.go`, `web_search_config.go`, `builtin_tools.go`, 4 locales, UI `ProviderKey` union).
- Strategic necessity: clear value — adds a corroboration engine to a chain that only grows when a tenant opts in; no maintenance burden for tenants without a key.

**Verification performed**
- Key wiring complete: the chain resolves `tools.web.<name>.api_key` generically, so `tools.web.serply.api_key` works; the `builtin_tools.go` allowlist entry makes the admin save path store/strip and report the key correctly.
- Endpoint/response contract cross-checked against Serply public docs: `GET https://api.serply.io/v1/search` with `X-Api-Key` is correct (the `?q=` spelling is explicitly accepted); response `results[].title/link/description` matches the parser.
- `maxSearchCount = 10` matches the provider's documented 10-result cap, so no extra clamping is needed (the comment is accurate).
- Tests: new table-driven coverage for request params, response mapping, limit + missing-link handling, error statuses, and cancellation. Reused helpers (`parallelRoundTripFunc`, `containsString`) exist in-package, so the tests compile.

**Findings**
- *Suggestion* — Geo param: `Country` is mapped to the `gl` query param, but Serply documents geo-targeting via the `X-Proxy-Location` header (US/DE/GB/…). Google-format query params are forwarded, so `gl` likely works — worth confirming, and consider `X-Proxy-Location` if the query param is ignored upstream.
- *Suggestion* — Result URLs: Serply notes `link` can carry Google tracking params (`client=`, `ved=`, `usg=`) on some results. Other providers return bare URLs; consider stripping known tracking params for consistency.
- *Suggestion* (test hygiene) — a Serply test reuses `parallelRoundTripFunc` from `web_search_parallel_test.go`; harmless, but a locally named stub would read better.

**Verdict**: Comment — no critical or important issues; minor suggestions only, safe to merge as-is.

*Posted by github-maintain automation (cron-safe run).*

googio:
Thanks for the review. I tested the first two against the live API rather than guessing, results below, and I have pushed a fix for the third.

**Geo param.** `gl` works. Same query, `num=4`, nothing else changed:

- `gl=gb` returns `bbc.co.uk/sport/football`, `skysports.com`, `thesun.co.uk`
- `gl=us` returns `nfl.com/news`, `espn.com/nfl`, `cbssports.com/nfl`

I also tested `X-Proxy-Location: GB` and `X-Proxy-Location: US` with no `gl` at all, and it returns the same two sets. So both paths work and the query param is not being ignored upstream. I left `gl` in place because it keeps the whole request in one place and matches how the other providers in this file pass options, but if you would rather the header be the documented path I am happy to switch it.

**Result URLs.** I could not reproduce this in the shape described. Across 8 queries and 73 result links, exactly one carried any of `ved` / `usg` / `client` / `sa`, and it was not a tracked web result. It was a local business row:

```
title: a Korean store name, metadata carries rating 4.4 and reviews 135
link:  https://www.google.com/searchviewer/10?client=ms-opera-mini-android&...&sa=X&ved=...
```

That is a local pack entry, and it has no destination page behind it. Stripping the known tracking params would turn it into `https://www.google.com/searchviewer/10`, which is worse than leaving it, since a caller can at least see what the current URL is. The other 72 links were bare, including legitimate Google properties like `developers.google.com/edu/python` and `google.com/finance/quote/TSLA:NASDAQ`, which a param stripper should not touch either.

I also checked the siblings: no provider in `internal/tools` normalizes result URLs today, so adding it only for Serply would make it the odd one out. My read is that the real item here is local pack rows rather than tracking params, it is about 1 in 70, and the fix would be dropping rows whose link is a `google.com` search viewer, not rewriting query strings. Happy to add that filter if you want it, but I would rather not put a param stripper in that does not address the case.

**Test stub.** Fair, and fixed in `5fe92954`. `parallelRoundTripFunc` was the only type of its kind in the package and my test was its only outside user, so the serply test now defines its own `serplyRoundTripFunc`. `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./internal/tools/` and `go test -race ./internal/tools/` are all green on the new head.

clark-cant (approved):
## Maintainer review (delta re-review) — github-maintain automation

**Delta under review**: commit `5fe929545295aa3ff0efdd603fb09174e198d831` (author response to the 2026-09-17 review, comment 2026-09-20T00:07Z). Base `dev`, head `feat/serply-search`.

**Summary**: Adds Serply as a keyed `web_search` provider, wired into the existing first-success-wins chain (constants, default order, provider factory, admin key allowlist, i18n, UI chain form). Inert unless a tenant stores its own Serply key.

**Risk level**: Low — additive; one new provider file plus registration points. No schema, API-contract, or migration changes. 11 files, +365/-10.

**Findings resolved since the previous review**
- *Suggestion (test hygiene)* — resolved in `5fe92954`: the serply cancellation test now defines its own `serplyRoundTripFunc` instead of borrowing `parallelRoundTripFunc` from another test file.
- *Suggestion (geo param)* — addressed: the author reproduced `gl=gb` / `gl=us` against the live API and confirmed `X-Proxy-Location` yields the same result sets, so the query param is not ignored upstream and keeps the request consistent with the other providers in this file. No change required.
- *Suggestion (result-URL tracking params)* — addressed: across 8 queries / 73 links the only param-carrying URL was a local-pack `google.com/searchviewer` row that has no destination page, so a query-string stripper would be the wrong fix. The author offers an optional local-pack filter instead — non-blocking and out of scope.

**Verification performed**
- Delta diff is confined to `internal/tools/web_search_serply_test.go` (+7/-1); no functional changes to the provider.
- Author reports `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./internal/tools/` and `go test -race ./internal/tools/` green on the new head.

**Mandatory gates (unchanged from the prior review)**
- Duplicate / prior implementation: clear — no existing provider covers Serply; no overlapping open PR.
- Project standards: docs found (`CLAUDE.md` / `AGENTS.md`); change follows the established one-file-per-provider pattern.
- Strategic necessity: clear value — adds an opt-in corroboration engine; no maintenance burden for tenants without a key.

**Verdict: Approve** — no critical or important issues; all prior findings resolved or justified with live evidence.

**Merge state**: `MERGEABLE` vs `dev`, `mergeStateStatus=UNSTABLE` — the commit status API reports `pending` with no checks posted (no CI attached to the branch). Not merge-ready under the automation contract (requires CLEAN + CI green), so no merge was attempted. A human maintainer can merge once CI/policy allows.

*Posted by /ck:review-pr at 2026-09-20T00:29:14Z*

PR #1565, opened 2026-09-14
fix(vault): deduplicate wikilinks before batch INSERT

hieunm-aisa:
## Summary
- Deduplicate `(FromDocID, ToDocID, LinkType)` in `SyncDocLinks` before calling `CreateLinks`, preventing PostgreSQL SQLSTATE 21000 error when a document contains multiple `[[target]]` wikilinks pointing to the same destination
- Concatenate contexts with ` | ` separator for duplicate targets so no surrounding text is lost
- Add regression tests verifying deduplication and preservation of distinct targets

Fixes #1473

## Surface parity
- Gateway server: fix in `internal/vault/links.go` + regression test in `internal/vault/links_test.go`
- API contract: N/A — internal store behavior, no wire changes
- Web UI: N/A — backend-only fix
- CLI/runtime package: N/A — no CLI changes

## Test plan
- [x] `TestSyncDocLinks_DeduplicatesSameTarget` — verifies duplicate `[[products]] [[products]]` produces exactly 1 link
- [x] `TestSyncDocLinks_PreservesDistinctTargets` — verifies `[[alpha]] [[beta]] [[alpha]]` produces 2 links (alpha + beta)
- [x] `go build ./internal/vault/ ./internal/store/pg/` — compile check
- [x] `go vet ./internal/vault/ ./internal/store/pg/` — static analysis
- [x] Full `go test ./internal/vault/` suite passes

Issue #494, opened 2026-03-26
feat: re-update agent system prompt


tiendn:
Hiện tại, hệ thống hỗ trợ tạo character files từ `pre-defined` system prompt thông qua Goclaw khi khởi tạo agent. 
Tuy nhiên, việc cập nhật lại logic này sau khi agent đã tồn tại còn gặp khó khăn, do phải chỉnh sửa thủ công từng file character, dẫn đến thiếu đồng nhất.

Thay đổi chính:
- Bổ sung function hỗ trợ cập nhật system prompt.
- Cơ chế: Chạy loop để đồng bộ lại các file character tương tự như quy trình khởi tạo agent mới.

clark-cant:
## Maintainer Triage — Issue #494

**Type:** Enhancement  
**Verdict:** `accept but narrow scope`

### Summary

Requesting the ability to re-sync/update agent system prompt context files (character files) after agent creation. Currently, context files are seeded once at agent creation via `SeedToStore()` in `bootstrap/seed_store.go`, and updating them requires manual editing — leading to inconsistency across agents.

### Codebase Evidence

- **Seed mechanism**: `bootstrap/seed_store.go:70` — `SeedToStore()` seeds embedded templates into `agent_context_files` table at agent creation
- **Agent creation flow**: `internal/http/agents.go:342` — Seeds context files during agent creation (skipped for open agents)
- **Regeneration exists**: `internal/http/summoner_regenerate.go:161` — `storeFiles()` saves LLM-generated files to `agent_context_files`, but this is for summoner-generated content, not predefined template re-sync
- **Storage**: `agent_context_files` table holds SOUL.md, IDENTITY.md, AGENTS.md, etc. per agent
- **No re-seed API**: No existing endpoint or CLI command to re-run `SeedToStore()` for an existing agent

### Assessment

✅ **Valid enhancement** — fits project direction. Agents should be able to update their system prompts without full recreation.

✅ **Building blocks exist** — `SeedToStore`, template system, `agent_context_files` CRUD are all in place.

⚠️ **Scope considerations**:
1. **Conflict handling**: What if user has customized files (e.g., SOUL.md edited via chat with self-evolve enabled)? Re-seed should either skip modified files or offer a `--force` flag
2. **Predefined only**: Open agents use per-user files; this should target predefined agents
3. **Selective re-seed**: Should support re-seeding specific files (e.g., just AGENTS.md) vs all files

### Recommended Implementation

[...]

PR #1570, opened 2026-09-21
feat(agents): let create opt out of background summoning with "summon": false

### LINKED ISSUE #1569: Agent create: background summoning silently overwrites context files written right after create
## Summary

Creating a predefined agent with an `agent_description` starts summoning in a background goroutine. It finishes ~10-20 seconds **after** the create response and writes `SOUL.md`, `IDENTITY.md`, `CAPABILITIES.md` and the agent `frontmatter`.

A client that manages agents as code writes its own context files through `agents.files.set` as soon as create returns. Summoning lands afterwards and replaces them. Nothing fails: no error from the API, nothing in the gateway logs, and the agent quietly runs on generated text instead of the committed role.

## Steps to reproduce

1. `POST /v1/agents` with `agent_type: predefined` and a non-empty `agent_description`.
2. As soon as it returns `201`, write your own `CAPABILITIES.md` via `agents.files.set`.
3. Wait ~20 seconds, then read it back (or `GET /v1/agents/{key}/system-prompt-preview`).

## Expected

Either the write wins, or the API gives the client a way to say "I bring my own files, do not summon".

## Actual

The file contains the summoner's generated template, and `frontmatter` is the generated one too. Real timeline from our gateway (`v3.15.0-beta.212`), where an agent was created and its role written in the same second:

```
08:02:56  POST /v1/agents            -> 201, status=summoning
08:02:56  agents.files.set CAPABILITIES.md   (our role, 5262 chars)
08:02:56  summoning: calling LLM     provider=litellm prompt_len=9077
08:03:09  summoning: raw LLM response length=14421
08:03:09  summoning: completed
```

The role was gone, replaced by the generated `# CAPABILITIES.md - What You Can Do` template. We only noticed because the agent ignored its instructions on the first run; a re-run of the installer silently fixed it, which makes this easy to misdiagnose as a flaky model.

## Why a flag rather than "just poll"

The only workaround today is to poll `GET /v1/agents/{key}` until `status` leaves `summoning`. That requires every automation to know that summoning exists, and it burns an LLM call whose entire output is thrown away. `POST /v1/agents/{id}/cancel-summon` does not help either: by the time the client could call it, the files may already be written.

## Proposal

Accept an optional `"summon": false` on `POST /v1/agents`:

- absent or `true` — current behaviour, existing clients unaffected;
- `false` — the agent is created `active` with the seeded template files and no LLM call is made; `POST /v1/agents/{id}/resummon` is still available afterwards.

Related but distinct: #570 fixes summoning overwriting a user-set `display_name` (same family, different field, fix inside `finishSummon`); #494 asks for re-running summoning on an existing agent.

I have a patch with tests and docs ready and will open a PR against this issue.

yatul:
Fixes #1569.

## Problem

Creating a predefined agent with a description kicks off summoning in a background goroutine. It finishes 10-20 seconds after the create response and writes `SOUL.md`, `IDENTITY.md`, `CAPABILITIES.md` and the agent `frontmatter`.

For a client that manages agents as code that is data loss. Such a client brings its own context files and writes them through `agents.files.set` as soon as create returns; summoning lands afterwards and replaces them. Nothing fails — no error from the API, nothing in the logs — and the agent quietly runs on generated text instead of the committed role. The only workaround today is to poll the agent status until it leaves `summoning`, which means every automation has to know that summoning exists at all, and it pays for an LLM call whose output is discarded.

## Change

`POST /v1/agents` accepts an optional `summon` field:

- absent or `true` — today's behaviour, so existing clients are unaffected;
- `false` — the agent is created `active` with the seeded template files and no LLM call is made. Summoning stays available via `POST /v1/agents/{id}/resummon`.

The decision moved into `shouldSummonOnCreate`, a small pure function, so it can be tested without starting a real background summon (which immediately reaches for the store and the message bus).

The status fallback also now resets an explicitly requested `summoning` status when nothing will summon the agent. That status is only ever cleared by the summoner, so such an agent would otherwise stay stuck in it forever.

## Tests

- `TestShouldSummonOnCreate` — five cases over the decision: default, opted out, no description, open agent, no summoner configured.
- `TestAgentsCreateSummonFalseGoesStraightToActive` — full handler run with `"summon": false`, asserting the persisted status. This is deterministic without any waiting: `handleCreate` starts the goroutine if and only if the status it decided synchronously is `summoning`, so a persisted `active` proves summoning never started. The summoner is still built with nil dependencies, so a regression that starts it anyway fails loudly.
- `TestAgentsCreateWithoutSummonFieldKeepsSummoning` — guards the default.

Checks from `CONTRIBUTING.md`, on `dev`: `go build ./...`, `go build -tags sqliteonly ./...`, `go vet ./internal/http/` clean; `go test -race ./internal/http/` passes for the whole package.

## Surface parity

- **CLI** — N/A: `goclaw agent add` (`cmd/setup_agent.go`) never sends `agent_description`, so it never summoned.
- **WS RPC** — N/A: `agents.create` (`internal/gateway/methods/agents_create.go`) creates agents `active` and never summoned.
- **Web UI** — N/A: the create flow wants summoning and keeps the default.
- **Docs** — updated: `docs/07-bootstrap-skills-memory.md` §7 and the endpoint table in `docs/18-http-api.md`.

## Context

Found while deploying a news-digest agent whose role is kept in a git repository and installed by script. The agent ran once without its role and nobody noticed until its output was wrong; a re-run of the installer silently fixed it, which makes this easy to misdiagnose as a flaky model.


clark-cant (commented):
**Summary**: Adds an optional `"summon": false` to `POST /v1/agents` so a client that manages agents as code can create an agent without the background LLM summoning that otherwise overwrites its context files ~10-20s after create. The decision is extracted into `shouldSummonOnCreate`, a stranded `summoning` status is now reset when nothing will summon, and the change is covered by tests plus docs.

**Risk level**: Low — opt-in field, default behaviour preserved, 4 files, +144/-3.

**Mandatory gates**:
- Duplicate / prior implementation: **clear**. No merged or open PR implements an opt-out. Related-but-distinct: #570 (preserve user-set `display_name` inside `finishSummon`, open) and #494 (re-run summoning on an existing agent, open). Neither covers this outcome.
- Project standards: **docs found** (`CONTRIBUTING.md`, `AGENTS.md`) — one process mismatch, see below.
- Strategic necessity: **clear value**. Removes a silent-data-loss path and a wasted LLM call for agents-as-code, while keeping the prose-described default.

**Findings**:

- **Important** — none found in the code.

- **Suggestion** — `TestAgentsCreateSummonFalseGoesStraightToActive` guards the regression with `time.Sleep(200 * time.Millisecond)`. A deterministic signal (a summoner stub that records invocation, or asserting the goroutine was never started) would be immune to slow CI and to scheduling luck. The nil-dependency summoner is a nice touch regardless.

- **Process (project standards)** — `CONTRIBUTING.md` → *Branch Strategy*: "All PRs target `dev` — `main` is frozen for stable releases"; only hotfixes target `main`. This PR targets `main`. Recent history agrees: 24 of the last 25 merged PRs went to `dev`, and the single `main`-targeted merge (#1565) is now stuck pending a forward-port to `dev`. Please retarget the base to `dev` (one click in the PR header).

- **Note (not caused by this diff)** — the `go` check is red because `main`'s `internal/http` test package does not compile:

```
internal/http/tenant_scope_hotfix_test.go:319:35: cannot use rec (variable of type *recordingBuiltinToolStore)
  as store.BuiltinToolStore value in struct literal: *recordingBuiltinToolStore does not implement
  store.BuiltinToolStore (missing method Upsert)
```

`recordingBuiltinToolStore` never implements `Upsert`, and on `main` the `store.BuiltinToolStore` interface declares it (the "additive fork-specific tools" comment). This reproduces on a clean `main` checkout, exactly as flagged in the PR description. On `dev` the interface has no `Upsert`, so this check is green there — retargeting to `dev` clears the red check without any change to this PR.

**Verified good**:
- Default preserved: `Summon == nil || *Summon` keeps today's behaviour, so existing clients are unaffected.
- Resetting an explicitly requested `summoning` status when nothing will summon is correct — that status is otherwise only cleared by the summoner, so such an agent would stay stuck forever.
- `shouldSummonOnCreate` matches the original predicate exactly, plus the opt-out.
- Surface parity is addressed with reasons (CLI/WS/Web N/A), and `docs/07` + `docs/18` are updated.
- Tests exercise the real handler; the nil-dependency summoner makes a regression fail loudly instead of passing quietly.

**Verdict**: **Comment** — no critical or important code issues; safe to merge as-is once the base branch matches `CONTRIBUTING` (`main` is release-only) and the required check is unblocked.

**Unresolved question**: Is `main` intentional here (i.e. treated as a hotfix), or should the base be `dev`? If `dev`, the pre-existing `internal/http` compile break on `main` also becomes moot for this PR.

*Posted by /ck:review-pr at 2026-09-21T08:58:05Z*

yatul (commented):
Thanks for the review — all three points addressed in `4e8ec437`.

**Base branch.** Not intentional, and not a hotfix — I missed the branch strategy in `CONTRIBUTING.md`. Retargeted to `dev` and rebased onto it; the change site in `internal/http/agents.go` is identical on `dev`, so the rebase was clean and the PR is still the same 4 files.

**Red `go` check.** Moot on `dev` as you said: the `internal/http` test package compiles there. I dropped the reviewer note about it from the description, and now run the whole package under the race detector instead of just the new tests: `go test -race ./internal/http/` passes.

**`time.Sleep` in the test.** Removed. I went with an argument rather than a stub, because the status assertion already is the deterministic signal: `handleCreate` starts the goroutine if and only if the status it decided synchronously is `summoning`:

```go
if req.Status == store.AgentStatusSummoning {
    go h.summoner.SummonAgent(...)
}
```

So a persisted `active` proves the goroutine was never spawned, with nothing to wait for. A recording stub would have needed a seam in production code — the summoner is a concrete `*AgentSummoner` — just to observe something the status already tells us. The comment in the test now spells this out. If you would still prefer an explicit seam, happy to add one.

clark-cant (approved):
**Summary**: Re-review of `4e8ec437` — the author addressed all three points from the 08:58Z review. The PR retargets to `dev`, drops the `time.Sleep` regression guard, and the previously red `go` check is now green on this head. Opt-in `"summon": false` on `POST /v1/agents` with default behaviour preserved and tests covering the decision.

**Risk level**: Low — opt-in field, default unchanged, 4 files, +144/-3, `mergeStateStatus=CLEAN`, checks `go`/`web`/`release-versioning` all SUCCESS.

**Mandatory gates**:
- Duplicate / prior implementation: **clear** — unchanged from the prior review; no merged or open PR implements an opt-out (`#570`, `#494` are related but distinct).
- Project standards: **docs found** (`CONTRIBUTING.md`, `AGENTS.md`) — the `main`-freeze mismatch is resolved: base is now `dev`.
- Strategic necessity: **clear value** — removes a silent data-loss path and a wasted LLM call for agents-as-code while keeping the described-agent default.

**Findings**:
- **Critical**: none.
- **Important**: none.
- **Suggestion**: none outstanding. The prior `time.Sleep(200ms)` suggestion is resolved — `TestAgentsCreateSummonFalseGoesStraightToActive` now asserts the persisted status, and the comment correctly explains why a persisted `active` proves the goroutine was never spawned. The nil-dependency summoner keeps a regression loud.

**Verified in the delta**:
- Base is now `dev` (was `main`), matching `CONTRIBUTING.md` "all PRs target `dev`".
- `shouldSummonOnCreate(requested, agentType, description, haveSummoner)` matches the original predicate plus the opt-out; `Summon == nil || *Summon` preserves the default for existing clients.
- The `summoning`-status reset when nothing will summon is correct (that status is only cleared by the summoner).
- Docs (`docs/07`, `docs/18`) updated; surface parity addressed with reasons.
- CI: `go`, `web`, `release-versioning` all SUCCESS on `4e8ec437`.

**Verdict**: **Approve** — no critical or important issues; merge-ready (`CLEAN` + green checks).

*Posted by /ck:review-pr at 2026-09-21T10:27:20Z*

